PR #4, opened 2026-06-12
feat: add OpenCode observability support

fangxiu-wf:
## Summary

- Add event_t JSONL collection and OTLP trace conversion for the **OpenCode** coding agent via a new `plugin-inject` deployment strategy
- OpenCode plugin hooks into OpenCode's native plugin system (`chat.message`, `chat.params`, `message.part.updated`, `tool.execute.before/after`, `experimental.chat.system.transform`) to emit structured event logs compliant with `EVENT_LOG_TO_TRACE_SPEC`
- Includes quality improvements to `validate-trace.mjs` (finish_reason enum, part structure, tool arguments) and `qoder-hook-processor.mjs` (authoritative stop_reason)

### New files
| File | Description |
|------|-------------|
| `agents.d/opencode.json` | Agent definition for OpenCode |
| `assets/plugins/opencode/plugin.mjs` | Event_t plugin — transforms OpenCode events into structured JSONL |
| `src/deployment/plugin-inject-strategy.ts` | Deploy strategy that injects plugin spec into OpenCode's `opencode.json` config |
| `src/inputs/opencode-log/opencode-log-input.ts` | Log input reader for OpenCode event_t files |

### Modified files
| File | Change |
|------|--------|
| `src/types/client-type.ts` | Rename `OpencodePlugin` → `OpenCode` |
| `src/types/deployment.ts` | Add `plugin-inject` DeployMode + `PluginInjectConfig` |
| `src/deployment/agent-def-loader.ts` | Allow `plugin-inject` in validation |
| `src/deployment/deployment-manager.ts` | Register `PluginInjectStrategy` |
| `src/core/orchestrator.ts` | Register `OpenCodeLogInput` with detection entry |
| `scripts/postinstall.js` | Copy `assets/plugins/` to `~/.loongsuite-pilot/plugins/` on install |

### Quality improvements (independent of OpenCode)
| File | Change |
|------|--------|
| `scripts/validate-trace.mjs` | Validate `finish_reason` against enum, check message part structure, add `semantic.tool_has_arguments` rule |
| `assets/hooks/qoder-hook-processor.mjs` | Use authoritative `stop_reason` from assistant message instead of inferring |

## Test plan

- [x] TypeScript type check passes (`tsc --noEmit`)
- [x] Build succeeds (`npm run build`)
- [x] Validated with multiple OpenCode sessions including parallel tool calls
- [x] Trace data verified against ARMS GenAI semantic conventions via `validate-trace.mjs`
- [ ] Reviewer to verify agent detection and plugin injection on a clean environment

### DIFF
--- a/assets/plugins/opencode/plugin.mjs
+++ b/assets/plugins/opencode/plugin.mjs
@@ -1,0 +1,849 @@
+/**
+ * loongsuite-pilot OpenCode event_t plugin
+ *
+ * Runs inside the OpenCode process (Bun runtime).
+ * Converts OpenCode EventV2 events into event_t JSONL records
+ * for consumption by loongsuite-pilot's BaseHookInput pipeline.
+ *
+ * Zero external dependencies — only Node/Bun built-in APIs.
+ */
+
+import fs from "node:fs";
+import path from "node:path";
+import os from "node:os";
+import crypto from "node:crypto";
+
+const AGENT_TYPE = "opencode";
+const MAX_SESSIONS = 100;
+const MAX_CONTENT_SIZE = 64 * 1024;
+
+// ---------------------------------------------------------------------------
+// Path helpers
+// ---------------------------------------------------------------------------
+
+function resolveDataDir() {
+  return (
+    process.env.LOONGSUITE_PILOT_DATA_DIR ||
+    path.join(os.homedir(), ".loongsuite-pilot")
+  );
+}
+
+function logDir() {
+  return path.join(resolveDataDir(), "logs", "opencode");
+}
+
+function ensureDir(dir) {
+  try {
+    fs.mkdirSync(dir, { recursive: true });
+  } catch {}
+}
+
+function todayStamp() {
+  const d = new Date();
+  return [
+    d.getFullYear(),
+    String(d.getMonth() + 1).padStart(2, "0"),
+    String(d.getDate()).padStart(2, "0"),
+  ].join("-");
+}
+
+// ---------------------------------------------------------------------------
+// ID generators
+// ---------------------------------------------------------------------------
+
+function generateTraceId() {
+  return crypto.randomBytes(16).toString("hex");
+}
+
+function generateSpanId() {
+  return crypto.randomBytes(8).toString("hex");
+}
+
+function nowNanos() {
+  return String(Date.now() * 1_000_000);
+}
+
+function msToNanos(ms) {
+  return typeof ms === "number" && Number.isFinite(ms)
+    ? String(Math.round(ms * 1_000_000))
+    : undefined;
+}
+
+// ---------------------------------------------------------------------------
+// Safe JSON serialization
+// ---------------------------------------------------------------------------
+
+function safeStringify(obj) {
+  const seen = new WeakSet();
+  return JSON.stringify(obj, function (_key, value) {
+    if (typeof value === "object" && value !== null) {
+      if (seen.has(value)) return "[Circular]";
+      seen.add(value);
+    }
+    if (typeof value === "function") return undefined;
+    if (typeof value === "bigint") return value.toString();
+    return value;
+  });
+}
+
+function truncate(str, max) {
+  if (typeof str !== "string") return str;
+  return str.length > max ? str.slice(0, max) + "...[truncated]" : str;
+}
+
+function truncateContent(val) {
+  if (typeof val === "string") return truncate(val, MAX_CONTENT_SIZE);
+  if (Array.isArray(val)) {
+    return val.map((item) => {
+      if (typeof item !== "object" || !item) return item;
+      const out = { ...item };
+      if (out.parts && Array.isArray(out.parts)) {
+        out.parts = out.parts.map((p) => {
+          if (typeof p?.content === "string")
+            return { ...p, content: truncate(p.content, MAX_CONTENT_SIZE) };
+          if (typeof p?.response === "string")
+            return { ...p, response: truncate(p.response, MAX_CONTENT_SIZE) };
+          return p;
+        });
+      }
+      return out;
+    });
+  }
+  return val;
+}
+
+// ---------------------------------------------------------------------------
+// Config
+// ---------------------------------------------------------------------------
+
+function loadPilotConfig() {
+  try {
+    const cfgPath = path.join(resolveDataDir(), "config.json");
+    const raw = fs.readFileSync(cfgPath, "utf-8");
+    return JSON.parse(raw);
+  } catch {
+    return {};
+  }
+}
+
+function resolveUserId(cfg) {
+  return (
+    process.env.LOONGSUITE_USER_ID ||
+    cfg.userId ||
+    os.hostname() ||
+    "unknown"
+  );
+}
+
+// ---------------------------------------------------------------------------
+// JSONL writer
+// ---------------------------------------------------------------------------
+
+let _logDirReady = false;
+
+function writeRecord(record) {
+  try {
+    if (!_logDirReady) {
+      ensureDir(logDir());
+      _logDirReady = true;
+    }
+    const filePath = path.join(logDir(), `opencode-${todayStamp()}.jsonl`);
+    fs.appendFileSync(filePath, safeStringify(record) + "\n");
+  } catch (err) {
+    writeError("writeRecord", err);
+  }
+}
+
+function writeError(source, err) {
+  try {
+    ensureDir(logDir());
+    const errPath = path.join(
+      logDir(),
+      `opencode-error-${todayStamp()}.log`
+    );
+    fs.appendFileSync(
+      errPath,
+      `${new Date().toISOString()} [${source}] ${err?.stack || err}\n`
+    );
+  } catch {}
+}
+
+// ---------------------------------------------------------------------------
+// Session state (LRU-bounded Map)
+// ---------------------------------------------------------------------------
+
+const sessions = new Map();
+
+function getSession(sessionID) {
+  if (!sessionID) return null;
+  let s = sessions.get(sessionID);
+  if (!s) {
+    s = {
+      turnSeq: 0,
+      currentTurn: null,
+      systemPrompt: null,
+      systemInstructionsParts: null,
+      agentMeta: null,
+      modelInfo: null,
+      llmParams: null,
+      pendingParts: [],
+      emittedToolCalls: new Set(),
+      stepStartTimeMs: null,
+      stepFinishData: null,
+    };
+    sessions.set(sessionID, s);
+    if (sessions.size > MAX_SESSIONS) {
+      const oldest = sessions.keys().next().value;
+      sessions.delete(oldest);
+    }
+  }
+  return s;
+}
+
+function clearSession(sessionID) {
+  sessions.delete(sessionID);
+}
+
+// ---------------------------------------------------------------------------
+// Record builder helpers
+// ---------------------------------------------------------------------------
+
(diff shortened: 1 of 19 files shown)


assets/plugins/opencode/plugin.mjs
```diff
@@ -0,0 +1,869 @@
...
+          case "message.updated":
+            handleMessageUpdated(props, userId);
+            break;
+          case "session.idle":
+          case "session.error": {
+            if (props.sessionID) {
+              const s = sessions.get(props.sessionID);
+              if (s && s.pendingParts && s.pendingParts.length > 0) {
+                safeLog("warn", `session ${eventName}: discarding ${s.pendingParts.length} unflushed pending part(s)`, {
+                  sessionID: props.sessionID,
```
linrunqi08: **Bug (P0):** `safeLog` is called but never defined anywhere in this file. Also `eventName` is not in scope — the switch variable is `type` (from `const type = event.type` on line 815).

This will throw a `ReferenceError` at runtime. The `safe()` wrapper catches it, but `clearSession(props.sessionID)` on line 835 is never reached because the error aborts the entire `if` block. Sessions with unflushed pending parts are never cleaned up, leaking into the LRU Map.

**Suggested fix:**
```js
// Replace safeLog with writeError, and eventName with type
writeError("session-cleanup", `session ${type}: discarding ${s.pendingParts.length} unflushed pending part(s)`);
```

Also on line 832: `s.pendingParts.map(p => p.type || "unknown")` should be `p.kind` — pending parts use `kind` (`"reasoning"`, `"text"`, `"tool_call"`), not `type`. As-is, every entry would report `"unknown"`.

scripts/validate-trace.mjs
```diff
@@ -11,6 +11,8 @@ const TAG = '[validate-trace]';
 const OTLP_DEBUG_DIR = path.join(homedir(), '.loongsuite-pilot', 'logs', 'otlp-debug');
 const VALID_SPAN_KINDS = ['ENTRY', 'AGENT', 'STEP', 'LLM', 'TOOL', 'CHAIN', 'RETRIEVER', 'RERANKER', 'EMBEDDING', 'TASK'];
 const KNOWN_SUBAGENT_TOOLS = new Set(['Agent']);
+const VALID_FINISH_REASONS = new Set(['stop', 'length', 'content_filter', 'tool_call', 'error', 'end_turn', 'max_tokens']);
```
linrunqi08: **Bug (P0):** `VALID_FINISH_REASONS` uses `'tool_call'` (singular), but at least 3 existing producers still emit `'tool_calls'` (plural):

- `assets/hooks/cursor/react-assembler.mjs:695`
- `assets/hooks/qoderwork-hook-processor.mjs:284`
- `src/inputs/qoder-work-log/qoder-work-trace-input.ts:664`

Only `qoder-hook-processor.mjs` was updated to singular in this PR. Running `validate-trace.mjs` on traces from the other 3 producers will now produce false schema errors: `finish_reason="tool_calls" is not a valid FinishReason`.

**Suggested fix:** Either:
1. Add `'tool_calls'` to `VALID_FINISH_REASONS` for backward compatibility, or
2. Update all 3 remaining producers to emit `'tool_call'` (singular) in this same PR to maintain consistency.

src/deployment/plugin-inject-strategy.ts
```diff
@@ -0,0 +1,211 @@
...
+    const config = def.pluginInject;
+    if (!config) return true;
+
+    const configPath = await this.findConfigFile(config);
+    if (!configPath) return true;
+
+    try {
+      const raw = await fs.readFile(configPath, 'utf-8');
+      const json = JSON.parse(stripJsoncComments(raw));
+      const plugins: unknown[] = json.plugin ?? json.plugins ?? [];
```
linrunqi08: **Bug (P0):** `needsDeploy()` reads both keys: `json.plugin ?? json.plugins`, but `deploy()` (line 124) only writes to `json.plugin`. If a config file uses the `plugins` key (plural):

1. `needsDeploy()` reads `json.plugins` → doesn't find the plugin → returns `true`
2. `deploy()` creates `json.plugin` (singular) → pushes the spec there
3. Host tool reads `plugins` (plural) → never sees the injected plugin
4. Next poll: `needsDeploy()` reads `json.plugins` again → still missing → infinite redeploy loop

`undeploy()` (line 170) has the same issue — it only checks `json.plugin`.

**Suggested fix:** Use the same key resolution in all three methods. Detect which key the file uses (`json.plugins` vs `json.plugin`) and operate on that key consistently.

src/deployment/plugin-inject-strategy.ts
```diff
@@ -0,0 +1,211 @@
...
+      }
+
+      // Remove existing matching spec to avoid duplicates
+      json.plugin = (json.plugin as unknown[]).filter(
+        (entry) => !this.matchesSpec(entry, resolvedSpec, config.pluginId),
+      );
+
+      json.plugin.push(resolvedSpec);
+
+      await fs.writeFile(configPath, JSON.stringify(json, null, 2) + '\n', 'utf-8');
```
linrunqi08: **Data integrity issue (P1):** The deploy writes `JSON.stringify(json, null, 2)` back to the config file, permanently destroying any JSONC comments from the original. The warning on line 149 logs *after* the file has already been overwritten — the user cannot prevent or recover from the data loss.

`undeploy()` (line 179) has the same comment-stripping behavior but logs no warning at all.

**Suggested fix:** Consider one of:
1. Preserve comments by doing string-level insertion (find the `plugin` array in the raw text and inject inline) instead of parse-modify-serialize.
2. At minimum, back up the original file before overwriting (e.g. `configPath + '.bak'`).
3. If neither is feasible, log the warning *before* overwriting so the user has a chance to see it in dry-run mode.

src/utils/fs-utils.ts
```diff
@@ -47,11 +47,25 @@ export async function writeJsonFile(
...
+  await ensureDir(dir);
   const text = `${JSON.stringify(data, null, 2)}\n`;
-  const tmp = path + '.tmp';
-  await fsp.writeFile(tmp, text, 'utf8');
-  await fsp.rename(tmp, path);
+  const tmp = `${path}.${process.pid}.${Date.now()}.tmp`;
+  try {
+    await fsp.writeFile(tmp, text, 'utf8');
+    await fsp.rename(tmp, path);
+  } catch (err: unknown) {
```
linrunqi08, line 57: **Data integrity issue (P1):** If `writeFile(tmp)` succeeds but `rename(tmp, path)` throws ENOENT, the catch block creates a new `tmp2` and retries. The original `tmp` file is never cleaned up and remains as an orphaned file on disk.

**Suggested fix:** Add cleanup of the first temp file at the start of the catch block:
```typescript
if ((err as NodeJS.ErrnoException).code === 'ENOENT') {
  await fsp.unlink(tmp).catch(() => {}); // clean up orphaned tmp
  await ensureDir(dir);
  // ...
}
```

src/inputs/qoder-work-trace/qoder-work-trace-input.ts
```diff
@@ -213,7 +230,8 @@ export class QoderWorkTraceInput extends BaseInput {
           if (lastNL >= 0) { text = text.substring(0, lastNL); consumedBytes = Buffer.byteLength(text, 'utf-8') + 1; }
         }
         this.stateStore.setOffset(fileStateKey, offset + consumedBytes);
-        this.stateStore.update(fileStateKey, { extra: { inode: currentInode } });
+        const sig = await computeFileSignature(filePath);
```
linrunqi08: **Efficiency (P2):** `computeFileSignature(filePath)` is called here, but the same first-1024-bytes hash was already computed earlier in this same function execution:
- Line 191: when inode changed
- Line 194: when inode was undefined  
- Line 207: rotation detection check

The file's first 1024 bytes cannot change between these calls within the same event-loop tick. This is a redundant `open → read → close → MD5` cycle per file per poll.

**Suggested fix:** Capture the signature from whichever earlier branch executed, and reuse it here:
```typescript
// At top of the per-file loop:
let computedSig: string | undefined;
// In each branch that calls computeFileSignature:
computedSig = await computeFileSignature(filePath);
// At line 233:
const sig = computedSig ?? await computeFileSignature(filePath);
```

linrunqi08 (commented):
## Code Review Summary

Reviewed 19 files (+1377/-45 lines) across 5 commits. Found **8 issues** — 3 correctness bugs (P0), 2 data integrity issues (P1), and 3 efficiency/cleanup issues (P2).

### P0 — Correctness Bugs
1. `plugin.mjs`: `safeLog` / `eventName` undefined → ReferenceError crashes session cleanup
2. `validate-trace.mjs`: `tool_call` vs `tool_calls` inconsistency across producers
3. `plugin-inject-strategy.ts`: `needsDeploy()` reads `plugin`/`plugins`, but `deploy()` only writes `plugin`

### P1 — Data Integrity
4. `plugin-inject-strategy.ts`: JSONC comments permanently destroyed on deploy
5. `fs-utils.ts`: ENOENT retry leaks orphaned temp file

### P2 — Efficiency
6. `qoder-work-trace-input.ts`: `computeFileSignature()` called redundantly 2-3x per file per poll
7. `plugin.mjs`: `pendingParts.map(p => p.type)` should be `p.kind`
8. `plugin.mjs`: `pendingParts` unbounded growth + linear scans

assets/plugins/opencode/plugin.mjs
fangxiu-wf: Fixed in 68ad64f. Changed `safeLog(...)` → `writeError(...)` (which is the correct logging function in scope), replaced `eventName` with `type`, and corrected `p.type` to `p.kind` for `pendingParts`.

scripts/validate-trace.mjs
fangxiu-wf: Fixed in 68ad64f. Added `'tool_calls'` to `VALID_FINISH_REASONS` for backward compatibility. Also added a TODO comment indicating that `tool_calls` will be removed once all producers are migrated to singular `tool_call`.

src/deployment/plugin-inject-strategy.ts
fangxiu-wf: Fixed in 68ad64f. Added a `resolvePluginKey(json)` helper that returns `'plugins'` when `json.plugins` is an array, otherwise defaults to `'plugin'`. All three methods (`needsDeploy`, `deploy`, `undeploy`) now use this unified key resolution, preventing the infinite redeploy loop.

src/deployment/plugin-inject-strategy.ts
fangxiu-wf: Fixed in 68ad64f. The warning is now logged **before** the overwrite, and the original file is backed up to `configPath + '.bak'` before writing. This applies to both `deploy()` and `undeploy()` paths.

src/utils/fs-utils.ts
fangxiu-wf: Fixed in 68ad64f. Added `await fsp.unlink(tmp).catch(() => {})` at the start of the ENOENT catch block to clean up the orphaned temp file before creating `tmp2`.

src/inputs/qoder-work-trace/qoder-work-trace-input.ts
fangxiu-wf: Fixed in 68ad64f. Introduced a `cachedSig` variable to store the result of `computeFileSignature()` across the different branches. The signature is now computed at most once per poll cycle and reused in subsequent checks via `cachedSig ?? await computeFileSignature(filePath)`.

PR #10, opened 2026-06-15
feat(cursor): add cursor-cli variant inference + deferred-stop assembly

Snssn:
## Summary
- Add cursor-cli variant inference to distinguish between Cursor IDE and Cursor CLI usage patterns
- Implement deferred-stop assembly logic for more accurate session tracking
- Fix orphan entry event and step timing issues in the assembler

## Test plan
- [ ] Verify cursor-cli variant is correctly inferred from CLI usage
- [ ] Validate deferred-stop assembly produces correct session boundaries
- [ ] Confirm no regression in existing cursor IDE data collection


ralf0131:
Please sign the CLA first.

ralf0131 (approved):

### Summary
Adds Cursor CLI variant support: (1) infers `cursor-cli` vs `cursor` IDE by detecting date-formatted `cursor_version`, (2) implements deferred-stop assembly to handle Cursor CLI's reversed event ordering (stop fires before `afterAgentResponse`), (3) adds `agents.d/cursor-cli.json` agent definition, (4) adds tool-result timestamp tracking for step timing.

### Findings
- **[Info]** `cursor-hook-processor.mjs:84-90` — `inferVariant()` correctly detects Cursor CLI by matching the date-based version pattern (`\d{4}\.\d{2}\.\d{2}`). The `cursor_version` field is the right signal since CLI uses calendar-versioned releases while IDE Cursor uses semantic versions.
- **[Info]** `cursor-hook-processor.mjs:152-178` — The deferred-stop logic is well-designed: when stop arrives but no response yet, it returns `null` (defer). When `afterAgentResponse` arrives and a stop is already pending, it assembles using the stop event's transcript path. This correctly handles the CLI's reversed event ordering.
- **[Warning]** `react-assembler.mjs` — The user-hook `llm.request` event (lines 65-77 in the old code) is removed entirely. This was the path for emitting the user's input as an `ENTRY` input event in IDE Cursor sessions. If the user input is not captured elsewhere in the pipeline for non-CLI sessions, this could result in missing `llm.request` events for user prompts. Please verify that user input is still captured via `messages_delta` in the step-level events for IDE Cursor.
- **[Info]** `react-assembler.mjs:294-299` — The `lastStepEndTs` tracking from tool result records is a good addition for accurate step timing.
- **[Info]** `cursor-hook-input.ts:82-90` — Server-side variant inference mirrors the client-side logic, correctly propagating `ClientType.CursorCli` through the telemetry pipeline.

### Cross-repo Note
This change is compatible with PR #32 (transcript-assembler refactor) since both target the same cursor-hook-processor but different code paths. No cross-repo changes needed.

---

ralf0131 (approved):
﻿### Summary
Re-review after force-push (commit `20bed0d`, squashed to a single commit). The PR adds Cursor CLI variant support with four components: (1) `inferVariant()` detects `cursor-cli` vs IDE `cursor` by matching the date-based version pattern, (2) deferred-stop assembly handles CLI reversed event ordering (stop fires before `afterAgentResponse`), (3) `agents.d/cursor-cli.json` agent definition, (4) server-side variant inference with `CursorCli` enum propagation and separate `cursor_cli` git namespace.

The previous review's [Warning] about the removed user-hook `llm.request` event is resolved: the user prompt is now emitted as `event.name: "other"` with `gen_ai.input.messages_delta`, and the comment explains the rationale clearly — "user prompt is not an LLM call — emit as other so converter merges messages_delta into ENTRY span without generating a standalone LLM span." This is semantically more correct than the previous `llm.request` classification, and the `messages_delta` payload is preserved so no data is lost.

### Findings
- **[Info]** `cursor-hook-processor.mjs` deferred-stop: clean implementation — when stop arrives with no `afterAgentResponse` for the conversation, it defers (`writeEmptyResponse`). The `afterAgentResponse` handler then checks for a pending stop and triggers assembly. Error handling with `appendErrorJsonl` is good.
- **[Info]** `react-assembler.mjs`: `variant` option defaults to `cursor`, preserving backward compatibility for IDE sessions. `gen_ai.agent.type` correctly uses the inferred variant.
- **[Info]** `cursor-hook-input.ts`: server-side `inferCursorVariant()` checks explicit `gen_ai.agent.type` first, then falls back to version pattern — robust two-tier detection.
- **[Info]** Minor: the stop handler and the `afterAgentResponse` deferred-assembly handler share duplicated journal-cleanup logic (pendingTurnConvIds/remaining). Consider extracting a helper to reduce duplication, though this is non-blocking.

CLA is signed (`license/cla = success`). The implementation is sound and the previous concern is addressed. Approving.

---

assets/hooks/cursor-hook-processor.mjs
```diff
@@ -188,7 +197,22 @@ async function main() {
...
 
+      // ─── Deferred-stop for Cursor CLI ───
+      // Cursor CLI fires stop BEFORE afterAgentResponse. If there's a prompt but
+      // no response yet for this conversation, defer assembly until the late
+      // response arrives.
+      const convId = internalEvent.conversation_id;
+      const hasResponse = allEvents.some(e =>
+        e.hook_event === 'afterAgentResponse' && e.conversation_id === convId
+      );
+      if (!hasResponse) {
```
linrunqi08: **Bug: Deferred-stop applies to ALL sessions, not just Cursor CLI — can silently lose IDE data.**

This `!hasResponse` check gates assembly for every conversation, not just cursor-cli. The comment says "Deferred-stop for Cursor CLI" but the code doesn't check the variant.

**Failure scenario:** A Cursor IDE user starts a prompt, the agent begins processing, the user aborts (Ctrl+C). Cursor fires `stop` but `afterAgentResponse` never arrives. Before this PR, the stop handler would assemble whatever data it has. After this PR, it defers and returns — but `afterAgentResponse` never fires, so the turn is never assembled. The journal accumulates the events, but they're orphaned and eventually dropped.

Suggested fix — gate on variant:
```js
const variant = inferVariant(allEvents);
if (variant === 'cursor-cli' && !hasResponse) {
  // defer — afterAgentResponse handler will trigger assembly
  writeEmptyResponse();
  return;
}
```
Or infer variant earlier and only defer for `cursor-cli`.

linrunqi08 (commented):
## Code Review Summary

Reviewed at **high effort** — 3 correctness angles + cleanup/altitude.

### Key Findings (5 items, severity-ranked)

| # | Severity | File | Finding |
| 1 | 🔴 Bug | cursor-hook-processor.mjs | Deferred-stop applies unconditionally — IDE sessions that abort without `afterAgentResponse` lose data silently |
| 2 | 🟡 Bug | scripts/lib/agent-overview.mjs | `classifyRecord` doesn't handle `cursor-cli` → records classified as `unknown` |
| 3 | 🟡 Bug | agent-event-normalizer.mjs | `AGENT_TYPE_TO_CONFIG_KEY` missing `cursor-cli` → content policy bypass |
| 4 | 🟢 Cleanup | cursor-hook-processor.mjs | Journal-rewrite logic duplicated identically in stop + afterAgentResponse handlers |
| 5 | 🟢 Cleanup | react-assembler.mjs | `inferProvider()` hardcodes `'cursor'` agent type; harmless now but inconsistent |

See inline comments for details on #1 and #4. Findings #2 and #3 are in files outside this diff — they are downstream consumers that need updating for the new `cursor-cli` variant:

**#2 — `classifyRecord` in `scripts/lib/agent-overview.mjs:140`** does `agentType === 'cursor'` exact match. Records with `gen_ai.agent.type: 'cursor-cli'` fall through to `'unknown'`, making Cursor CLI usage invisible in the monitoring dashboard. Fix: add `if (agentType === 'cursor-cli') return 'cursor';` (or a separate bucket if you want CLI tracked separately).

**#3 — `AGENT_TYPE_TO_CONFIG_KEY` in `assets/hooks/agent-event-normalizer.mjs:276-282`** has no `'cursor-cli'` entry. When `applyHookContentPolicy` processes cursor-cli records, it looks up `agents['cursor-cli']` (miss) then `agents[AGENT_TYPE_TO_CONFIG_KEY['cursor-cli']]` (undefined → miss). So cursor-cli records bypass any content policy configured under `agents.cursor`. Fix: add `'cursor-cli': 'cursor'` to the mapping.

---

**Overall:** The core design (variant inference + deferred assembly for reversed CLI event ordering) is sound. The main concern is #1 — the deferred-stop gate should be conditional on cursor-cli, or include a fallback so IDE abort/error sessions still get assembled.

assets/hooks/cursor-hook-processor.mjs
```diff
@@ -247,6 +268,51 @@ async function main() {
...
+        const variant = inferVariant(allEvents);
+        const result = assembleTurn(allEvents, {
+          runtimeConfig,
+          variant,
+          stopConversationId: convId,
+        });
+
+        if (result.records.length > 0) {
+          const day = localDateString(now);
+          const historyFile = path.join(dataDir, 'logs', 'cursor', 'history', `cursor-${day}.jsonl`);
```
linrunqi08, line 299: **Cleanup: Journal-rewrite logic duplicated verbatim from the stop handler.**

The `pendingTurnConvIds` / `remaining` loop + `rewriteJournal` call here is identical to the one in the stop handler above. Consider extracting a helper:
```js
function rewriteJournalForConsumed(allEvents, consumedConversationIds) {
  const pendingTurnConvIds = new Set();
  const remaining = [];
  for (const ev of allEvents) {
    if (consumedConversationIds.has(ev.conversation_id)) continue;
    if (ev.hook_event === 'beforeSubmitPrompt') pendingTurnConvIds.add(ev.conversation_id);
  }
  for (const ev of allEvents) {
    if (consumedConversationIds.has(ev.conversation_id)) continue;
    if (pendingTurnConvIds.has(ev.conversation_id)) remaining.push(ev);
  }
  rewriteJournal(remaining, allEvents);
}
```
Without extraction, a future fix to one path risks missing the other.

assets/hooks/cursor-hook-processor.mjs
```diff
@@ -247,6 +268,51 @@ async function main() {
...
+  // ─── Deferred-stop compensation: assemble when late response arrives ───
+  // When stop was deferred (no response yet), afterAgentResponse triggers assembly.
+  if (internalEvent.hook_event === 'afterAgentResponse') {
+    try {
+      const allEvents = readAllEvents();
+      const convId = internalEvent.conversation_id;
+      const hasStop = allEvents.some(e =>
+        e.hook_event === 'stop' && e.conversation_id === convId
+      );
+      if (hasStop) {
```
linrunqi08, line 285: **Note: `transcriptPath` not passed to `assembleTurn` in the deferred path.**

The stop handler passes `transcriptPath: internalEvent.transcript_path`, but this deferred path omits it. This works because `assembleTurn` falls back to `stopEvent?.transcript_path` (react-assembler.mjs line 89). Correct behavior, but a comment here explaining the deliberate omission would prevent a future contributor from "fixing" it by adding `transcriptPath: internalEvent.transcript_path` (which would be wrong — `internalEvent` here is the afterAgentResponse event, not the stop event).

assets/hooks/cursor/react-assembler.mjs
```diff
@@ -84,13 +85,14 @@ export function assembleTurn(journalEvents, options = {}) {
     trace_id: traceId,
     'gen_ai.session.id': parentConvId,
     'gen_ai.turn.id': turnId,
-    'gen_ai.agent.type': 'cursor',
+    'gen_ai.agent.type': variant,
```
linrunqi08, line 88: **Minor: `inferProvider` still hardcodes `'cursor'` agent type.**

At line 770, `inferProvider()` calls `inferProviderName({ ..., 'gen_ai.agent.type': 'cursor' })`. For cursor-cli records this passes the wrong agent type. Currently harmless because `inferProviderName` doesn't match `cursor` or `cursor-cli` in its agent-type fallback chain (only `codex`, `claude`, `qoder`, `gemini`). But if model inference is ever extended for cursor-specific patterns, CLI records would get wrong results. Consider passing `variant` through `ctx` so `inferProvider` can use it.

ralf0131 (approved):
﻿### Summary
Re-review after commit `5932df9`. All three findings from @linrunqi08's review are now properly addressed:

1. **Deferred-stop gating** (was Bug) — now correctly gated on `variant === "cursor-cli" && !hasResponse`; IDE sessions always assemble immediately, so abort/error scenarios no longer lose data.
2. **`classifyRecord` cursor-cli handling** (was Bug) — `agent-overview.mjs` now maps both `cursor` and `cursor-cli` to `"cursor"`.
3. **Normalizer config key** (was Bug) — `agent-event-normalizer.mjs` now includes `'cursor-cli': 'cursor'`.

The `compactJournal` extraction preserves the original journal-rewrite logic verbatim. The deferred `afterAgentResponse` handler correctly re-checks for a prior `stop` event and deliberately omits `transcriptPath` (well-documented). TypeScript-side `inferCursorVariant` checks explicit type before falling back to the version pattern. LGTM.

### Findings
- **[Info]** No unit tests for the new `inferVariant()`, `compactJournal()`, or deferred-stop assembly path. Recommend adding coverage when feasible.

---

PR #19, opened 2026-06-16
fix: avoid creating bash_profile for PATH setup

JhihJian:
## Summary

This PR fixes a shell profile side effect in the open-source installer.

When installing for a bash user without an existing `~/.bash_profile`, the installer currently creates one just to prepend `~/.local/bin` to `PATH`. On Debian/Ubuntu-style accounts, bash login shells read `~/.bash_profile` instead of `~/.profile` when it exists. Since the default `~/.profile` is commonly responsible for sourcing `~/.bashrc`, creating `~/.bash_profile` can silently skip user aliases and interactive shell setup. A visible symptom is aliases like `ll` no longer being available in new login shells.

## Fix

- Keep writing the PATH block to `~/.bashrc` for bash users.
- Only write `~/.bash_profile` if the user already has one.
- Otherwise write `~/.bash_login` if it already exists.
- Fall back to `~/.profile` instead of creating `~/.bash_profile`.
- Add a unit test to guard this installer profile selection behavior.

## Validation

- `bash -n deploy/installer-opensource.sh`
- `vitest run tests/unit/deploy/installer-profile-selection.test.mjs`

### DIFF
--- a/deploy/installer-opensource.sh
+++ b/deploy/installer-opensource.sh
@@ -823,7 +823,16 @@ PATHBLOCK
                 ;;
             */bash)
                 ensure_path_block "$HOME/.bashrc" || true
-                ensure_path_block "$HOME/.bash_profile" || true
+                # Do not create ~/.bash_profile just to add PATH. On Debian/Ubuntu
+                # style accounts, its mere presence prevents bash login shells from
+                # reading ~/.profile, which often sources ~/.bashrc and user aliases.
+                if [ -f "$HOME/.bash_profile" ]; then
+                    ensure_path_block "$HOME/.bash_profile" || true
+                elif [ -f "$HOME/.bash_login" ]; then
+                    ensure_path_block "$HOME/.bash_login" || true
+                else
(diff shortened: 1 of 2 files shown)


ralf0131 (approved):

### Summary
Avoids creating `~/.bash_profile` when adding `~/.local/bin` to PATH. Instead, respects bash's own login-file precedence: updates `.bash_profile` only if it already exists, otherwise falls back to `.bash_login` then `.profile`. This prevents breaking Debian/Ubuntu login shells where `.bash_profile` presence suppresses `.profile` sourcing.

### Findings
- **[Info]** `installer-opensource.sh` — The fallback chain (`bash_profile` → `bash_login` → `profile`) correctly mirrors bash's own login-file resolution. Always writing to `.bashrc` first is correct since non-interactive shells source `.bashrc`.
- **[Info]** Tests — The test verifies the installer script structure by extracting the bash case block and asserting the correct ordering of checks and writes. Good approach for testing shell script logic from JavaScript.

LGTM. Correct fix for a real footgun in installer PATH management.

---

PR #20, opened 2026-06-16
fix: chunked OTLP export with gzip to prevent large trace failures

fangxiu-wf:
## Summary

- Large traces (1000+ spans, e.g. 1329 spans from a long Cursor session) failed to export because `OtlpTraceFlusher` sent all spans in a single `OTLPTraceExporter.export()` call, exceeding the 10s timeout and ARMS body size limit (~2-5MB).
- This PR splits the export into batches by estimated serialized size (default 10MB pre-compression), and enables gzip compression by default on `OTLPTraceExporter`.

## Changes

- **`src/flushers/otlp-trace-flusher.ts`**: Add `exportInBatches()` method that accumulates spans by estimated size, splitting into new batches when the threshold is exceeded. Add lightweight `estimateSpanSize()` that sums attribute string lengths (no `JSON.stringify`). Enable gzip compression by default.
- **`src/types/index.ts`**: Add `maxExportBatchBytes` and `compression` fields to `OtlpTraceFlusherConfig` and `OtlpTraceRawConfig`.
- **`src/core/config-loader.ts`**: Pass through new config fields in both `buildOtlpTraceConfigNew` and `buildOtlpTraceConfigLegacy`.
- **`tests/unit/flushers/otlp-trace-flusher/export.test.ts`**: Add 3 test cases for size-based batching behavior.

## Test plan

- [x] TypeScript type check passes
- [x] All 40 existing + new unit tests pass (7 test files)
- [ ] Verify with a real large trace (1000+ spans) that export succeeds after this change

### DIFF
--- a/src/flushers/otlp-trace-flusher.ts
+++ b/src/flushers/otlp-trace-flusher.ts
@@ -7,6 +7,7 @@ import {
 } from '@opentelemetry/sdk-trace-base';
 import type { ReadableSpan } from '@opentelemetry/sdk-trace-base';
 import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-proto';
+import { CompressionAlgorithm } from '@opentelemetry/otlp-exporter-base';
 import {
   convertEventLogToTrace,
   ExtendedTelemetryHandler,
@@ -76,6 +77,18 @@ function getPilotVersion(): string {
   } catch {
     return 'unknown';
   }
+}
+
+const DEFAULT_MAX_EXPORT_BATCH_BYTES = 10 * 1024 * 1024; // 10 MB
+
+function estimateSpanSize(span: ReadableSpan): number {
+  let size = 512;
+  const attrs = span.attributes;
+  for (const val of Object.values(attrs)) {
+    if (typeof val === 'string') size += val.length;
+    else size += 32;
+  }
+  return size;
 }
 
 export class OtlpTraceFlusher extends BaseFlusher {
@@ -238,7 +251,7 @@ export class OtlpTraceFlusher extends BaseFlusher {
     if (this.cfg.debug) {
       await this.writeDebugLog(agentType, spans);
     }
-    await this.doExport(exportState, agentType, spans);
+    await this.exportInBatches(exportState, agentType, spans);
   }
 
   // --- Internal ---
@@ -349,7 +362,37 @@ export class OtlpTraceFlusher extends BaseFlusher {
       await this.writeDebugLog(agentType, spans);
     }
 
-    await this.doExport(exportState, agentType, spans);
+    await this.exportInBatches(exportState, agentType, spans);
+  }
+
+  private async exportInBatches(
+    exportState: AgentExportState,
+    agentType: string,
+    spans: ReadableSpan[],
+  ): Promise<void> {
+    const maxBytes = this.cfg.maxExportBatchBytes ?? DEFAULT_MAX_EXPORT_BATCH_BYTES;
+    const batches: ReadableSpan[][] = [];
+    let current: ReadableSpan[] = [];
+    let currentSize = 0;
+
+    for (const span of spans) {
+      const size = estimateSpanSize(span);
+      if (current.length > 0 && currentSize + size > maxBytes) {
+        batches.push(current);
+        current = [];
+        currentSize = 0;
+      }
+      current.push(span);
+      currentSize += size;
+    }
+    if (current.length > 0) batches.push(current);
+
+    if (batches.length > 1) {
+      logger.info(`Exporting ${spans.length} spans in ${batches.length} batches`, { agentType, maxBytes });
+    }
+    for (const batch of batches) {
+      await this.doExport(exportState, agentType, batch);
+    }
   }
 
   private doExport(
@@ -396,6 +439,7 @@ export class OtlpTraceFlusher extends BaseFlusher {
     const exporter = new OTLPTraceExporter({
       url: this.resolvedEndpointUrl,
       headers: this.cfg.headers ?? {},
+      compression: (this.cfg.compression ?? 'gzip') as CompressionAlgorithm,
     });
 
     state = { exporter };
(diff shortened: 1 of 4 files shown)


linrunqi08 (commented):
## Code Review Summary

Reviewed 4 files, +100/-2 lines. The chunked export approach and batching logic are sound. Three findings below.

Overall the PR cleanly solves the original timeout problem for the common case (large attribute payloads). The main concern is that `estimateSpanSize` may significantly underestimate for spans with event payloads, which could let the original failure resurface in certain GenAI trace shapes.

src/flushers/otlp-trace-flusher.ts
```diff
@@ -7,6 +7,7 @@ import {
 } from '@opentelemetry/sdk-trace-base';
 import type { ReadableSpan } from '@opentelemetry/sdk-trace-base';
 import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-proto';
+import { CompressionAlgorithm } from '@opentelemetry/otlp-exporter-base';
 import {
   convertEventLogToTrace,
```
linrunqi08, line 12: **Bug — `estimateSpanSize` ignores span events and links, may significantly underestimate**

The function only sums attribute string lengths, but `OTLPTraceExporter.export()` serializes the full span including `events` and `links`. In OpenTelemetry GenAI semantic conventions, large content (prompt/completion messages) can be carried in span events (e.g. `gen_ai.content.prompt`, `gen_ai.content.completion` event names with their own attributes).

For a span with a 200-byte attribute but 50KB of event payload, `estimateSpanSize` returns ~712 bytes while the actual serialized size is ~51KB — a 70× underestimate. This means batches could still exceed the ARMS body size limit, re-triggering the original timeout this PR sets out to fix.

**Suggested fix:** Also iterate `span.events` and sum their attribute sizes:

```typescript
function estimateSpanSize(span: ReadableSpan): number {
  let size = 512;
  for (const val of Object.values(span.attributes)) {
    if (typeof val === 'string') size += val.length;
    else size += 32;
  }
  for (const event of span.events ?? []) {
    size += 64; // event name + timestamp overhead
    for (const val of Object.values(event.attributes ?? {})) {
      if (typeof val === 'string') size += val.length;
      else size += 32;
    }
  }
  return size;
}
```

tests/unit/flushers/otlp-trace-flusher/export.test.ts
```diff
@@ -114,6 +114,52 @@ describe('OtlpTraceFlusher - export', () => {
...
+  it('uses default 10MB limit when maxExportBatchBytes not configured', async () => {
+    // Small spans (~522 bytes each), 100 of them ≈ 52KB ≪ 10MB → single batch
+    const flusher = new OtlpTraceFlusher(makeConfig());
+    const spans = Array.from({ length: 100 }, () => makeMockSpan()) as any;
+
+    await flusher.exportSpansForAgent('claude-code', spans);
+    await flusher.shutdown();
+
+    expect(mockExport).toHaveBeenCalledTimes(1);
+    expect(mockExport.mock.calls[0][0]).toHaveLength(100);
```
linrunqi08, line 161: **Test gap — no coverage for gzip compression default**

The batching tests are solid, but there's no assertion that `OTLPTraceExporter` is constructed with `compression: 'gzip'` by default (or `compression: 'none'` when explicitly configured). Since gzip-by-default is a behavior change for all users, a test that inspects the constructor args would guard against regressions.

Example:
```typescript
it('enables gzip compression by default', async () => {
  const { OTLPTraceExporter } = await import('@opentelemetry/exporter-trace-otlp-proto');
  const flusher = new OtlpTraceFlusher(makeConfig());
  await flusher.exportSpansForAgent('claude-code', [makeMockSpan()] as any);
  await flusher.shutdown();

  expect(vi.mocked(OTLPTraceExporter)).toHaveBeenCalledWith(
    expect.objectContaining({ compression: 'gzip' }),
  );
});
```

src/flushers/otlp-trace-flusher.ts
```diff
@@ -396,6 +439,7 @@ export class OtlpTraceFlusher extends BaseFlusher {
     const exporter = new OTLPTraceExporter({
       url: this.resolvedEndpointUrl,
       headers: this.cfg.headers ?? {},
+      compression: (this.cfg.compression ?? 'gzip') as CompressionAlgorithm,
```
linrunqi08: **Cleanup — use `CompressionAlgorithm` enum values instead of type assertion**

`(this.cfg.compression ?? 'gzip') as CompressionAlgorithm` bypasses type checking. While currently correct (`CompressionAlgorithm.GZIP === 'gzip'`, `CompressionAlgorithm.NONE === 'none'`), the `as` cast silently hides any future mismatch if the OTel library changes enum representations.

Consider:
```typescript
compression: this.cfg.compression === 'none'
  ? CompressionAlgorithm.NONE
  : CompressionAlgorithm.GZIP,
```

This way the compiler enforces correctness.

src/flushers/otlp-trace-flusher.ts
fangxiu-wf: 已修复。`estimateSpanSize` 现在会遍历 `span.events` 并累加其属性大小（每个 event 基础开销 64 bytes + 属性字符串长度）。当前 GenAI traces 中 links 基本为空，暂不计入，后续如有需要可以补充。→ 33d22f9

tests/unit/flushers/otlp-trace-flusher/export.test.ts
fangxiu-wf: 已补充。新增单元测试 `enables gzip compression by default on OTLPTraceExporter`，通过 `vi.mocked(OTLPTraceExporter)` 检查构造参数中 `compression` 字段为 `"gzip"`。→ 33d22f9

src/flushers/otlp-trace-flusher.ts
fangxiu-wf: 已修复。移除 `as CompressionAlgorithm` 类型断言，改为显式条件表达式映射到 `CompressionAlgorithm.NONE` / `CompressionAlgorithm.GZIP` 枚举值，保证类型安全。→ 33d22f9

linrunqi08 (approved):

PR #42, opened 2026-06-18
fix(qoder-work): fix extractText dropping user query and interrupt causing spurious turns

linrunqi08:
## Summary
- **extractText**: Changed from returning only the first text block to concatenating all text blocks. QoderWork puts system prompt and user query in separate `text` blocks within `message.content`; the old `for…return` pattern returned on the first block (system prompt), losing the actual user question.
- **isSystemInjection**: Added `[Request interrupted` pattern so that interrupt messages don't start a spurious new turn.

## Test plan
- [x] Verified on Windows machine with live QoderWork sessions
- [x] Confirmed user queries now appear in collected data
- [x] Confirmed `[Request interrupted` messages no longer split turns

### DIFF
--- a/assets/hooks/qoderwork-hook-processor.mjs
+++ b/assets/hooks/qoderwork-hook-processor.mjs
@@ -448,4 +448,6 @@ function resolveQoderWorkProjectDir(sandboxCwd, agentId) {
   return sandboxCwd;
 }
 
+export { extractText, isSystemInjection, isToolResult, splitIntoTurns };
+
 main().catch(() => { /* fail-open */ });
--- a/assets/hooks/qoderwork-hook-processor.mjs
+++ b/assets/hooks/qoderwork-hook-processor.mjs
@@ -141,7 +141,9 @@ function splitIntoTurns(contentRows) {
 
 function isSystemInjection(row) {
   const text = extractText(row).trimStart();
-  return text.startsWith('<command-message>') || text.startsWith('<command-name>');
+  return text.startsWith('<command-message>')
+    || text.startsWith('<command-name>')
+    || text.startsWith('[Request interrupted');
 }
 
 function buildTurnEvents(turnRows, turnId, sessionId, userId, providerName, version, observedTs, runtimeConfig, cwd) {
@@ -399,10 +401,12 @@ function extractText(row) {
   const content = msg.content;
   if (typeof content === 'string') return content;
   if (Array.isArray(content)) {
+    const parts = [];
     for (const block of content) {
-      if (block.type === 'text') return block.text || '';
-      if (typeof block === 'string') return block;
+      if (block.type === 'text' && block.text) parts.push(block.text);
+      else if (typeof block === 'string') parts.push(block);
     }
+    return parts.join('\n');
   }
   return '';
 }
--- a/tests/unit/hooks/qoderwork-hook-processor.test.mjs
+++ b/tests/unit/hooks/qoderwork-hook-processor.test.mjs
@@ -1,0 +1,108 @@
+import { describe, expect, it } from 'vitest';
+import {
+  extractText,
+  isSystemInjection,
+  isToolResult,
+  splitIntoTurns,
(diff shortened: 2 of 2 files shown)


assets/hooks/qoderwork-hook-processor.mjs
```diff
@@ -399,10 +401,12 @@ function extractText(row) {
   const content = msg.content;
   if (typeof content === 'string') return content;
   if (Array.isArray(content)) {
+    const parts = [];
```
ralf0131, line 404: Tests: this is a correctness fix for a data-loss bug (multi-block user content was truncated to the first text block), yet no regression test is added and none of the existing test files import this processor module to cover `extractText`/`isSystemInjection`. A small unit test feeding a `row.message.content` array like `[{type:"text",text:"first"},{type:"tool_use"},{type:"text",text:"second"}]` and asserting `extractText` returns both blocks joined by `\n` would lock in this fix and guard the new `[Request interrupted` injection path too.

assets/hooks/qoderwork-hook-processor.mjs
```diff
@@ -141,7 +141,9 @@ function splitIntoTurns(contentRows) {
 
 function isSystemInjection(row) {
   const text = extractText(row).trimStart();
-  return text.startsWith('<command-message>') || text.startsWith('<command-name>');
+  return text.startsWith('<command-message>')
+    || text.startsWith('<command-name>')
```
ralf0131, line 145: Nit: `extractText` now concatenates *all* text blocks with `\n`. This is the right call for these QoderWork transcripts, but worth confirming the upstream CLI never emits multiple text blocks whose order/semantics matter when flattened — since the joined result is propagated verbatim into `gen_ai.input.messages_delta` content.

ralf0131 (commented):

### Summary
`fix(qoder-work): extractText dropping user query and interrupt causing spurious turns` — the core fix is correct and well-scoped. `extractText` previously early-returned only the first text/string block, silently dropping the rest of multi-block `message.content`; it now collects all text blocks and joins with `\\n`. Since the joined text is propagated into both the per-turn `llm.request` `input.messages_delta` and the step-1 input delta, this directly restores full user-query capture. The accompanying `isSystemInjection` addition (`[Request interrupted` prefix) correctly prevents interrupt artifacts from being mistaken for user turns. Overall verdict: good to merge; one coverage suggestion below.

### Findings
- **[Warning]** assets/hooks/qoderwork-hook-processor.mjs:404 — no regression test added, and no existing test imports this module to cover `extractText`/`isSystemInjection`. Recommend a small unit test for the multi-block case + the new interrupt-injection path.
- **[Info]** assets/hooks/qoderwork-hook-processor.mjs:145 — confirm upstream CLI never emits multi-block text whose order/semantics matter when flattened into `messages_delta`.

### Cross-repo Note
None — change is isolated to the QoderWork hook processor (`assets/hooks/`), no public API or protocol impact.

---

messixukejia (approved):

ralf0131 (approved):

### Summary
Re-verification after new commits. The fix is correct and well-tested: `extractText` no longer early-returns on the first content block, so multi-block user messages (e.g. system prompt + user query) are fully captured; `[Request interrupted` is now treated as a system injection to prevent spurious turn splits. Clean APPROVE.

### Findings
- **[Info]** `qoderwork-hook-processor.mjs:404` — The `parts.join('\n')` accumulator correctly replaces the old `for...return` early-exit bug. Empty/falsy text blocks are skipped via `block.text &&`, which is the right guard. Good coverage in the new unit tests.
- **[Info]** `qoderwork-hook-processor.mjs:144` — Adding `[Request interrupted` to `isSystemInjection` is a sensible prefix match (literal `startsWith`, so the open `[` is fine) and prevents interrupted requests from spawning empty turns.

### Tests
108 lines of focused unit tests cover: string content, missing message, single/multi text blocks, plain string blocks, empty text blocks, all three injection prefixes, and turn splitting. Exporting the internal functions purely for testability is clean.

LGTM. Solid bugfix with good test coverage.

---

PR #45, opened 2026-06-18
feat(qoder-jetbrains): add Qoder for JetBrains as a first-class data source

Snssn:
## Summary

Add Qoder for JetBrains (IntelliJ IDEA plugin) as an independent agent data source (`qoder-idea`), clearly distinguished from Qoder Desktop IDE (`qoder`) and Qoder CLI (`qoder-cli`).

### Key Changes

**1. DB-based Variant Detection (Stable Architecture-Level Signal)**
- Replaced unreliable transcript-path heuristic with SQLite DB path detection
- IntelliJ writes to `~/.qoder/shared_client/cache/db/local.db`
- Desktop IDE writes to `~/Library/Application Support/Qoder/.../local.db`
- Query `SELECT 1 FROM chat_session WHERE session_id = ?` (<2ms) at hook time
- 100% coverage across 50 sessions, 0 false matches

**2. Detection-Only Deploy Mode**
- New `detection-only` deploy mode for agents that share hooks but need separate detection
- `DetectionOnlyStrategy`: no-op deploy, only runs detection
- Prevents duplicate hook installation when IntelliJ shares `~/.qoder/settings.json` with CLI

**3. Glob Path Support in Detection**
- `detect-utils.ts` now supports `*` and `?` wildcards in detection paths
- Enables matching `JetBrains/<IDE>*/plugins/qoder-jetbrains` across IDE versions

**4. Trace Quality Fixes**
- Fixed IDE LLM boundary splitting: only split by timestamp when `PreToolUse` events exist
- Prevents erroneous multi-step splitting on tool-free sessions with extended thinking
- Added `toolFinishedTs` from SQLite for real tool execution duration (replaces fake 1ms)
- Phantom step fixup: clamp timestamps, synthesize missing tool.results, fill response.id

**5. Token Data Integration**
- `sqlite-token-reader.ts` queries both Desktop and IntelliJ DBs
- Returns `toolFinishedTs` for accurate `gen_ai.tool.call.duration`
- `token-enricher.ts` uses real SQLite timestamps instead of hardcoded +1ms

### Validation Results

| Metric | Result |
|--------|--------|
| Field Coverage (all 4 event types) | ✅ 100% |
| validate-trace (single message) | ✅ 32/32 PASS, 0 errors |
| validate-trace (14-step agent loop) | 63 pass / 1 error (history reprocess artifact) |
| Unit tests | 131/131 passed |
| TypeScript typecheck | Clean |

### Files Changed (26 files, +1285/-63)

- `assets/hooks/qoder-hook-processor.mjs` — DB-based `inferVariant`, `isQoderIdeaSession`, LLM boundary fix
- `assets/hooks/agent-event-normalizer.mjs` — Same DB detection in `inferQoderVariant`
- `src/inputs/qoder-trace/sqlite-token-reader.ts` — Multi-DB query, `toolFinishedTs`
- `src/inputs/qoder-trace/token-enricher.ts` — Real tool timing, phantom step fixes
- `src/deployment/detect-utils.ts` — Glob path support
- `src/deployment/detection-only-strategy.ts` — New no-op strategy
- `agents.d/qoder-jetbrains.json` — Detection-only agent definition
- `src/types/client-type.ts` — `QoderIdeaHook` enum
- `src/types/deployment.ts` — `detection-only` deploy mode
- Tests: 4 new test files, updated existing tests

### Test Plan
- [x] 131 unit tests pass
- [x] TypeScript typecheck clean
- [x] Live validation: IntelliJ messages → `qoder-idea-*.jsonl`, Desktop → `qoder-*.jsonl`
- [x] Field coverage 100% on real IntelliJ data
- [x] Trace validation PASS on new sessions


assets/hooks/agent-event-normalizer.mjs
```diff
@@ -593,15 +599,38 @@ function asRecord(value) {
   return value && typeof value === 'object' && !Array.isArray(value) ? value : {};
 }
 
-function inferQoderVariant(row, sourceAgentId) {
+function inferQoderVariant(row, sourceAgentId, transcriptPath, sessionId, options = {}) {
```
ralf0131: **[Warning] Unused `transcriptPath` parameter.** This function (and `inferVariant` in `qoder-hook-processor.mjs`) now receives `transcriptPath`, but the body never reads it — variant detection is driven entirely by the pre-computed `isQoderIdea` flag and the `sessionId`-based DB query. The dangling parameter suggests the path-shape detection from the design doc is still active when it was actually superseded by the SQLite approach. Suggest removing the parameter (or adding a brief `// reserved / unused` note) so future readers don't waste time tracing a dead input.

openspec/changes/add-qoder-jetbrains-datasource/specs/qoder-jetbrains-datasource/spec.md
```diff
@@ -0,0 +1,81 @@
+## ADDED Requirements
+
+### Requirement: Recognize IntelliJ Qoder plugin transcripts via path shape
+The system SHALL identify Qoder for JetBrains transcripts by inspecting the `transcriptPath` reported in the Stop hook payload, independent of transcript row contents or the hook command argument.
```
ralf0131: **[Info] Spec diverges from implementation.** This requirement states detection works "by inspecting the `transcriptPath`", but the actual implementation (PR body + `isQoderIdeaSession`) switched to a SQLite `chat_session` query against `~/.qoder/shared_client/cache/db/local.db`. The spec and its `transcript/`-segment scenarios are now stale. Consider updating the spec to describe the DB-based detection so the acceptance criteria match what shipped.

ralf0131 (commented):

### Summary
Adds Qoder for JetBrains (`qoder-idea`) as a first-class datasource, cleanly separated from Qoder Desktop IDE (`qoder`) and Qoder CLI (`qoder-cli`) via a per-transcript SQLite `chat_session` lookup. The new `detection-only` deploy mode, glob-aware detection paths, and the IDE-boundary splitting / phantom-step fixups in the token enricher all look correct and well-tested (131 unit tests pass). Overall this is a high-quality, well-documented change — leaving two minor notes only.

### Findings
- **[Warning]** `assets/hooks/agent-event-normalizer.mjs:602` — `transcriptPath` threaded through `inferQoderVariant` / `inferVariant` but unused; dead input.
- **[Info]** `openspec/.../spec.md:4` — spec still documents path-based detection; implementation now uses the SQLite DB approach, so the spec is stale.

### Strengths
- DB detection is correctly cached once per transcript (`isQoderIdeaSession`) instead of per-row — good hot-path discipline.
- `globToRegex` properly escapes regex metacharacters and restricts `*` to a single path segment — safe for versioned dir matching.
- The 5 timestamp-fixup passes in `token-enricher.ts` (response.id backfill, tool duration, clamping, orphan synthesis) handle the messy IntelliJ streaming reality defensively; the `splice`+`idx++` skip in pass 5 is correct.
- Tests cover glob edge cases (`?` vs `*`, no-match, mixed) and the new `DetectionOnlyStrategy`.

---

ralf0131 (approved):

ralf0131:
## Re-review by github-manager-bot

### Summary
Follow-up commit `de20fe6` addresses both prior review findings and the cross-platform CI failure. LGTM — changes look correct.

### Findings Resolution
- ✅ **[Warning] resolved** — unused `transcriptPath` parameter removed from `inferQoderVariant()`. Verified `sessionId` (now the 3rd positional arg) is genuinely consumed by the DB lookup (`SELECT 1 FROM chat_session WHERE session_id = ?`), so no new dead param was introduced. The single call site in `buildQoderHookRecord` was updated in lockstep.
- ✅ **[Info] resolved** — spec now documents the SQLite DB-based detection architecture instead of the old `transcriptPath` path heuristic, matching the implementation.

### Cross-Platform CI Fix
`qoder-trace-sqlite-reader.test.ts` now creates the Desktop DB at both macOS (`Library/Application Support/Qoder/...`) and Linux (`.config/Qoder/...`) locations. Confirmed these align with `resolveQoderDbCandidates()` in the source (darwin and linux branches respectively), so the test no longer assumes a macOS-only layout. The "first DB with rows wins" ordering logic is preserved.

### CI
All 3 `build-and-test` jobs pass. (CLA is pending for the contributor.)

---
*Automated re-review by github-manager-bot*

ralf0131:
**CLA Not Signed**

The Contributor License Agreement (CLA) check is currently pending on this PR (`license/cla`: Contributor License Agreement is not signed yet.). This PR **cannot be merged** until the CLA is signed.

@Snssn please sign the CLA via the CLA assistant badge in the comment above, or visit https://cla-assistant.io/alibaba/loongsuite-pilot. Once signed, the `license/cla` status will turn green.

---
*Automated check by github-manager-bot*

ralf0131 (commented):
### Summary
Clean implementation of Qoder-for-JetBrains as a first-class data source: a new `DetectionOnlyStrategy`, glob-aware `directoryExists` in `detect-utils.ts`, and token-enricher/sqlite-reader updates. The openspec design doc (D1–D8) is thorough and the rationale for a `detection-only` deploy mode (IntelliJ shares qoder's Stop hook) is sound. Good test coverage across deployment + trace-input.

⚠️ **CLA is NOT signed** — this PR cannot be approved/merged until @Snssn signs the CLA (see the reminder comment above). Per policy I'm leaving a COMMENT, not an approval, regardless of code quality.

### Findings
- **[Info]** `detect-utils.ts` — glob expansion is hand-rolled (`globToRegex`/`walk`) rather than using an existing dep (`fast-glob`/`glob`). The tasks.md (3.1) explicitly listed "用 fast-glob 或 glob 包；若已有依赖优先复用". The hand-rolled version works for the JetBrains path patterns, but a literal `?` in a path segment would be interpreted as a single-char wildcard. Low risk for IDEA dirs; just flagging the deviation from the plan.
- **[Info]** `detection-only-strategy.ts` — `deploy()` resolves but performs no hook install (correct by design). Confirm there's no relied-upon side effect (e.g. metrics/registry update) that callers expect from `deploy()` beyond the no-op; the tests assert the no-op, which is good.
- **[Info]** `token-enricher.ts` (+257/-38) — large change; verify the IntelliJ SQLite schema fields referenced match the plugin's actual DB (the design notes the schema is shared with qoder-cli). Tests in `qoder-trace-sqlite-reader.test.ts` cover this.

---

linrunqi08:
## Code Review — PR #45: feat(qoder-jetbrains)

Overall this is a well-structured PR with thorough design docs, good test coverage, and clear commit-level reasoning. The DB-based variant detection is a solid architectural improvement over the old transcript-path heuristic. Below are findings from a high-effort review (correctness + cleanup + efficiency + altitude angles).

---

### Summary of Findings

| # | Severity | Category | File | Finding |
|---|----------|----------|------|---------|
| 1 | 🔴 Medium | Bug | `detect-utils.ts` | `globToRegex` regex character class has unescaped `[` |
| 2 | 🟡 Low | Bug | `qoder-hook-processor.mjs` / `agent-event-normalizer.mjs` | SQLite `DatabaseSync` handle leak — no `try/finally` for `db.close()` |
| 3 | 🟡 Low | Bug | `agent-event-normalizer.mjs` | `qoderSourceNamespace()` missing `qoder-idea` case |
| 4 | 🟡 Low | Bug | `token-enricher.ts` | Sequential matching can misalign tokens if phantom responses sort before real ones |
| 5 | 🔵 Info | Reuse | Two hook files | `isQoderIdeaSession` duplicated |
| 6 | 🔵 Info | Efficiency | `token-enricher.ts` | O(n²) nested `entries.find()`/`entries.some()` in 5 sequential final passes |
| 7 | 🔵 Info | Altitude | `token-enricher.ts` | `fixupPhantomSteps` fixes symptoms downstream |

---

### Finding 1: `globToRegex` — malformed regex character class

**File:** `src/deployment/detect-utils.ts` — `globToRegex` function

The regex `/[.+^${}()|[\]\\$]/g` contains an unescaped `[` inside a character class. In V8 (Node.js) this works by treating the bare `[` as a literal, but it is technically invalid per the ECMAScript spec and could break in future engine versions or stricter modes.

**Suggested fix:**
```js
else body += ch.replace(/[.+^${}()\[\]|\\]/g, '\\$&');
```
This properly escapes `[` and `]`, and removes the redundant second `$` from the original class.

---

### Finding 2: SQLite handle leak when `db.prepare().get()` throws

**Files:** `assets/hooks/qoder-hook-processor.mjs` (`isQoderIdeaSession`) and `assets/hooks/agent-event-normalizer.mjs` (`inferQoderVariant` inline DB check)

If `db.prepare('SELECT 1 ...').get(sessionId)` throws (e.g., table doesn't exist, DB corruption), `db.close()` is skipped. The `catch` block returns `null` but the `DatabaseSync` handle remains open. `node:sqlite`'s `DatabaseSync` does not auto-close on GC, so this leaks a file descriptor until process exit.

**Suggested fix — use `try/finally`:**
```js
const db = new DatabaseSync(dbPath, { readonly: true });
try {
  const row = db.prepare('SELECT 1 FROM chat_session WHERE session_id = ? LIMIT 1').get(sessionId);
  return row !== undefined;
} finally {
  db.close();
}
```

---

### Finding 3: `qoderSourceNamespace()` missing `qoder-idea` case

**File:** `assets/hooks/agent-event-normalizer.mjs` line ~526

When `variant === 'qoder-idea'`, this function falls through to `return 'qoder'`. IntelliJ-originated records get source attributes keyed as `agent.qoder.variant`, `agent.qoder.cwd`, etc. — identical keys to Desktop IDE records. While `agent.qoder.variant` **value** will be `'qoder-idea'`, the attribute **key** namespace is shared with Desktop IDE, which could cause confusion in downstream queries filtering by key prefix.

**Suggested fix:**
```js
function qoderSourceNamespace(variant) {
  if (variant === 'qoder-work') return 'qoderwork';
  if (variant === 'qoder-cn') return 'qodercn';
  if (variant === 'qoder-idea') return 'qoderidea';
  return 'qoder';
}
```
If sharing the `qoder` namespace is intentional, please add a comment explaining why.

---

### Finding 4: Sequential matching can misalign tokens with phantom responses

**File:** `src/inputs/qoder-trace/token-enricher.ts` — `enrichIdeTurn` sequential matching section

The old closest-timestamp matching (with 1000ms threshold) was resilient to extra/phantom `llm.response` entries because it matched by proximity. The new sequential matching assumes a strict 1:1 chronological correspondence between SQLite rows and `responseEntries`.

If phantom response entries (from hook processor over-segmenting) sort **before** their corresponding real response entry in `responseEntries`, the sequential matching will pair the phantom with the SQLite row, and the real response won't get its token data. Critically, `fixupPhantomSteps` operates only on **unmatched** entries, so it won't detect a phantom that was incorrectly matched.

I understand the new `buildLlmBoundaries` change (removing timestamp-based splitting) should prevent most phantom steps. But if any still occur (e.g., from historical data reprocessing), this would silently corrupt token attribution.

**Suggestion:** After sequential matching, verify that matched entries have non-empty `gen_ai.output.messages` and un-match entries that look like phantoms (empty/thinking-only content).

---

### Finding 5: `isQoderIdeaSession` is duplicated

**Files:** `assets/hooks/qoder-hook-processor.mjs` (standalone function) and `assets/hooks/agent-event-normalizer.mjs` (inline in `inferQoderVariant`)

The same DB-detection logic exists in both files — both open `~/.qoder/shared_client/cache/db/local.db`, run the same `SELECT 1 FROM chat_session WHERE session_id = ?` query, and interpret the result identically. If the DB path or table name changes, both must be updated.

**Suggestion:** Extract into a shared helper at `assets/hooks/shared/qoder-db-utils.mjs`.

---

### Finding 6: O(n²) final passes in `enrichIdeTurn`

**File:** `src/inputs/qoder-trace/token-enricher.ts` — final passes 2-5

Final passes 2-5 each contain `entries.find()` or `entries.some()` inside a loop over `entries`, giving O(n²) per pass. With 5 such passes, total work is ~5n².

For typical sessions (10-50 entries) this is fine. For long agent loops with hundreds of entries, consider building a `Map<stepId, entry[]>` lookup once and reusing across all passes.

---

### Finding 7: `fixupPhantomSteps` fixes downstream symptoms

**File:** `src/inputs/qoder-trace/token-enricher.ts`

The 5 sequential "final passes" + `fixupPhantomSteps` accumulate complexity to fix data quality issues that originate in the hook processor's boundary segmentation. The fix-at-source approach (in `buildLlmBoundaries`, which this PR also improves) is the right long-term direction.

**Suggestion:** Add a comment noting that `fixupPhantomSteps` is a transitional safety net expected to become a no-op once boundary segmentation is fully correct, so future maintainers don't keep adding more downstream fixups.

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "loongsuite-pilot",
-  "version": "1.1.5",
+  "version": "1.0.0",
   "lockfileVersion": 3,
```
ralf0131, line 4: **[Critical]** Package version downgraded from `1.1.5` to `1.0.0`. This is almost certainly unintended — likely a stale local `package-lock.json` from an older fork point. This would break published package versioning and downstream dependency resolution. Please rebase against latest `main` and regenerate `package-lock.json` (or simply revert this file).

assets/hooks/shared/qoder-db-utils.mjs
```diff
@@ -0,0 +1,29 @@
+import fs from 'node:fs';
+import { homedir } from 'node:os';
+
+// node:sqlite is available in Node 22+. Load once at module init.
+let DatabaseSync = null;
+try {
+  ({ DatabaseSync } = await import('node:sqlite'));
```
ralf0131, line 9: **[Warning]** `node:sqlite` is only available in Node 22+ (still experimental). The repo config specifies `runtime.node: "18"`. On Node 18, `DatabaseSync` will be `null` and `isQoderIdeaSession()` always returns `null` — the entire `qoder-idea` variant detection is silently non-functional. Please confirm the production runtime is Node 22+, or consider a cross-version SQLite binding (e.g. `better-sqlite3`).

assets/hooks/shared/qoder-db-utils.mjs
```diff
@@ -0,0 +1,29 @@
...
+ * Returns true  → session found in ~/.qoder/shared_client/cache/db/local.db (IntelliJ)
+ * Returns false → DB exists but session not found (Desktop IDE or CLI)
+ * Returns null  → DB unavailable or query failed
+ */
+export function isQoderIdeaSession(sessionId) {
+  if (!DatabaseSync) return null;
+  const dbPath = homedir() + '/.qoder/shared_client/cache/db/local.db';
+  if (!fs.existsSync(dbPath)) return null;
+  const db = new DatabaseSync(dbPath, { readonly: true });
+  try {
```
ralf0131, line 30: **[Warning]** A new `DatabaseSync` connection is opened and closed on every call to `isQoderIdeaSession()`. This function is called from `inferQoderVariant()` per-record during hook processing. For sessions with many turns, this means N open/close cycles. Consider caching the result per `sessionId` (e.g. a `Map`), or opening the connection once and reusing it for the batch.

ralf0131 (commented):
﻿### Summary
Re-review after commit `be2e96c`. The PR adds Qoder-for-JetBrains as a first-class data source with a `DetectionOnlyStrategy`, glob-aware path detection, and DB-based IntelliJ session inference via `qoder-db-utils.mjs`.

⚠️ **CLA is still NOT signed** — this PR cannot be approved/merged until @Snssn signs the CLA. Per policy I'm leaving a COMMENT, not an approval.

Three new findings since the last review:

### Findings
- **[Critical]** `package-lock.json:4` — package version downgraded from `1.1.5` to `1.0.0` (see inline comment). This must be reverted before merge.
- **[Warning]** `qoder-db-utils.mjs:7` — `node:sqlite` requires Node 22+; the repo config targets Node 18, making the `qoder-idea` detection silently non-functional on the configured runtime.
- **[Warning]** `qoder-db-utils.mjs:21` — `isQoderIdeaSession()` opens/closes a DB connection on every call. Consider caching per `sessionId` to avoid N connection cycles per session.

### Suggestions
For the Node version gap, either confirm the production runtime is Node 22+ (and update `repos/loongsuite.yaml` `runtime.node` accordingly), or use `better-sqlite3` which supports Node 18+.

---

agents.d/qoder-jetbrains.json
```diff
@@ -0,0 +1,20 @@
+{
+  "id": "qoder-jetbrains",
+  "displayName": "Qoder for JetBrains",
+  "deployMode": "detection-only",
+  "detection": {
+    "paths": [
+      "~/Library/Application Support/JetBrains/IntelliJIdea*/plugins/qoder-jetbrains",
```
ralf0131, line 7: **[Info]** Detection paths cover macOS (`~/Library/Application Support/JetBrains/`) and Linux (`~/.config/JetBrains/`), but not Windows. On Windows, JetBrains plugins are typically under `%APPDATA%\JetBrains\<IDE>*\plugins\qoder-jetbrains`. Note that the DB-based variant detection in `qoder-db-utils.mjs` uses `homedir() + '/.qoder/...'` which works cross-platform, so IntelliJ session detection itself functions on Windows — only the agent-definition detection paths are macOS/Linux-only. Consider adding Windows paths for completeness if Windows support is intended.

ralf0131 (commented):
﻿### Summary
Re-review after commit `f4753224` ("fix: address ralf0131 re-review findings"). All three prior findings are resolved:

- ✅ **[Critical] package-lock.json version downgrade** — reverted back to `1.1.5`.
- ✅ **[Warning] DB open/close per call** — per-session `_cache` Map added; `isQoderIdeaSession()` now caches results for the process lifetime.
- ✅ **[Warning] Node 18 / `node:sqlite` gap** — comment now explicitly documents the graceful fallback (`null` → callers fall back to `qoder` Desktop IDE variant) on Node < 22.

⚠️ **CLA is still NOT signed** — this PR cannot be approved/merged until @Snssn signs the CLA. Per policy, leaving a COMMENT (not an approval).

One remaining Info-level observation (non-blocking):

### Findings
- **[Info]** `agents.d/qoder-jetbrains.json:7` — Windows JetBrains plugin detection paths missing (see inline comment).

### Suggestions
The DB-based IntelliJ session detection (`qoder-db-utils.mjs`) already works cross-platform via `homedir()`, so this is only about the agent-definition detection paths. If Windows support is intended, add paths like `~/AppData/Roaming/JetBrains/IntelliJIdea*/plugins/qoder-jetbrains`.

---

linrunqi08 (approved):

PR #46, opened 2026-06-18
fix(qoder): improve LLM boundary detection and token enrichment

rangemer333-cell:
## Summary
- qoder-hook-processor: use progress event windows to detect LLM request/response boundaries more accurately
- agent-event-normalizer: emit user events as `other` instead of `llm.request` to avoid duplicate span creation
- token-enricher: match tokens by SQLite request ordering with confidence tracking
- sqlite-token-reader: extend SqliteTokenData with sessionId/messageId/model fields
- qoder-trace-input: group IDE turns by session before enrichment

## Test plan
- [ ] Run existing unit tests: `npx vitest run tests/unit/hooks/agent-event-normalizer.test.mjs`
- [ ] Run qoder-trace input tests: `npx vitest run tests/unit/inputs/qoder-trace-input.test.ts`
- [ ] Run integration test: `npx vitest run tests/integration/hook-jsonl-flow.test.ts`
- [ ] Verify IDE trace output with real Qoder IDE session

### DIFF
--- a/assets/hooks/qoder-hook-processor.mjs
+++ b/assets/hooks/qoder-hook-processor.mjs
@@ -315,6 +315,11 @@ function progressWindowKey(progressEvents, rowMs) {
     }
   }
 
+  // If no start boundary found, this row appears before any progress event.
+  // Return null to let the caller fall back to time-gap grouping, avoiding
+  // incorrect merging of distinct LLM calls that precede the first progress event.
+  if (!startTs) return null;
+
   let endTs = null;
   for (const pe of progressEvents) {
     const peMs = Date.parse(pe.ts) || 0;
@@ -325,7 +330,7 @@ function progressWindowKey(progressEvents, rowMs) {
     }
   }
 
-  return `progress:${startTs || ''}->${endTs || ''}`;
+  return `progress:${startTs}->${endTs || ''}`;
 }
 
 // --- Event Builder -----------------------------------------------------------
--- a/assets/hooks/qoder-hook-processor.mjs
+++ b/assets/hooks/qoder-hook-processor.mjs
@@ -216,50 +216,57 @@ async function processTranscript(agentId, logPrefix, transcriptPath, sessionId, 
 
 function buildLlmBoundaries(progressEvents, contentEvents) {
   // Step 1: Group assistant blocks into LLM calls.
-  // Priority: use message.id (CLI variant has it) > fallback to timestamp proximity (IDE variant).
+  // Priority: use message.id (CLI variant has it) > progress window (IDE variant)
+  // > fallback to timestamp proximity when progress events are absent.
   const assistantGroups = [];
   let currentGroup = [];
   let lastTs = null;
-  let lastMessageId = null;
+  let currentKey = null;
+  const hasProgressWindows = progressEvents.some(pe =>
+    pe.hookEvent === 'UserPromptSubmit' || pe.hookEvent === 'PostToolUse' ||
+    pe.hookEvent === 'PreToolUse' || pe.hookEvent === 'Stop'
+  );
+
+  function flushGroup() {
+    if (currentGroup.length > 0) assistantGroups.push(currentGroup);
+    currentGroup = [];
+    lastTs = null;
+    currentKey = null;
+  }
 
   for (const row of contentEvents) {
     if (row.type !== 'assistant') {
-      if (currentGroup.length > 0) {
-        assistantGroups.push(currentGroup);
-        currentGroup = [];
-        lastTs = null;
-        lastMessageId = null;
-      }
+      flushGroup();
       continue;
     }
     const ts = row.timestamp ? Date.parse(row.timestamp) : 0;
     if (!ts) continue;
 
     const messageId = row.message?.id || null;
+    const key = messageId
+      ? `message:${messageId}`
+      : hasProgressWindows
+        ? progressWindowKey(progressEvents, ts)
+        : null;
 
     // Determine if this row starts a new LLM call
     let isNewCall = false;
-    if (messageId && lastMessageId) {
-      // Both have message.id: new call when id changes
-      isNewCall = messageId !== lastMessageId;
-    } else if (!messageId && !lastMessageId) {
-      // Neither has message.id (IDE): use timestamp proximity
+    if (currentGroup.length > 0 && key && currentKey) {
+      isNewCall = key !== currentKey;
+    } else if (currentGroup.length > 0 && !key && !currentKey) {
       isNewCall = lastTs !== null && (ts - lastTs) > 200;
-    } else if (currentGroup.length > 0) {
-      // Mixed (shouldn't happen, but handle gracefully)
+    } else if (currentGroup.length > 0 && key !== currentKey) {
+      // Mixed keyed/unkeyed rows are unusual; keep the old time-gap fallback.
       isNewCall = lastTs !== null && (ts - lastTs) > 200;
     }
 
-    if (isNewCall && currentGroup.length > 0) {
-      assistantGroups.push(currentGroup);
-      currentGroup = [];
-    }
+    if (isNewCall) flushGroup();
 
     currentGroup.push(row);
+    currentKey = key;
     lastTs = ts;
-    if (messageId) lastMessageId = messageId;
   }
-  if (currentGroup.length > 0) assistantGroups.push(currentGroup);
+  flushGroup();
 
   // Step 2: For each assistant group (= one LLM call), find timing from progress
   const boundaries = [];
@@ -298,6 +305,29 @@ function buildLlmBoundaries(progressEvents, contentEvents) {
   return boundaries;
 }
 
+function progressWindowKey(progressEvents, rowMs) {
+  let startTs = null;
+  for (const pe of progressEvents) {
+    const peMs = Date.parse(pe.ts) || 0;
+    if (peMs >= rowMs) break;
+    if (pe.hookEvent === 'PostToolUse' || pe.hookEvent === 'UserPromptSubmit') {
+      startTs = pe.ts;
+    }
+  }
+
+  let endTs = null;
+  for (const pe of progressEvents) {
+    const peMs = Date.parse(pe.ts) || 0;
+    if (peMs <= rowMs) continue;
+    if (pe.hookEvent === 'PreToolUse' || pe.hookEvent === 'Stop') {
+      endTs = pe.ts;
+      break;
+    }
+  }
+
+  return `progress:${startTs || ''}->${endTs || ''}`;
+}
+
 // --- Event Builder -----------------------------------------------------------
 
 function buildEventsFromBoundaries(boundaries, contentEvents, allParsed, turnId, sessionId, agentId, runtimeConfig, cwd) {
@@ -317,7 +347,7 @@ function buildEventsFromBoundaries(boundaries, contentEvents, allParsed, turnId,
       const userHookModel = contentEvents.find(r => r.type === 'assistant' && r.message?.model)?.message?.model || 'unknown';
       records.push({
         'event.id': crypto.randomUUID(),
-        'event.name': 'llm.request',
+        'event.name': 'other',
         'gen_ai.turn.id': turnId,
         'gen_ai.session.id': sessionId,
         'gen_ai.agent.type': agentType,
@@ -609,7 +639,7 @@ function buildLegacyEvents(contentEvents, turnId, sessionId, agentId, runtimeCon
   for (const record of existingRecords) {
     const eventName = record['event.name'];
     const rawType = record['agent.qoder.raw_type'];
-    if (eventName === 'llm.request' && rawType === 'user') continue;
+    if (rawType === 'user') continue;
     if (eventName === 'llm.response') {
       const responseTs = record['time_unix_nano'] || '';
       const tsDiff = lastResponseTs === null ? Infinity : Math.abs(Number(responseTs) - Number(lastResponseTs));
--- a/assets/hooks/qoder-hook-processor.mjs
+++ b/assets/hooks/qoder-hook-processor.mjs
@@ -189,19 +189,30 @@ async function processTranscript(agentId, logPrefix, transcriptPath, sessionId, 
     // session_meta, other types: ignored
   }
 
-  // --- Phase 3: Determine LLM call boundaries ---
-  // Use content events (assistant blocks) for boundary detection,
-  // then use progress timestamps for precise timing.
-  const llmBoundaries = buildLlmBoundaries(progressEvents, contentEvents);
-  logDebug(agentId, `Detected ${llmBoundaries.length} LLM call(s)`);
+  // --- Phase 3: Split content events into turns by real user prompts ---
+  // Each real user prompt starts a new turn. Tool results stay attached to the
+  // preceding turn. This ensures each turn gets its own user input instead of
+  // inheriting the first prompt of the whole transcript segment.
+  const turnSegments = splitContentEventsIntoTurns(contentEvents);
+  logDebug(agentId, `Split transcript segment into ${turnSegments.length} turn(s)`);
 
-  // --- Phase 4: Build events using boundaries ---
-  const turnId = crypto.randomUUID();
-  const records = buildEventsFromBoundaries(
-    llmBoundaries, contentEvents, parsed, turnId, sessionId, agentId, runtimeConfig, cwd,
-  );
+  // --- Phase 4: Build events per turn ---
+  const records = [];
+  for (let turnIdx = 0; turnIdx < turnSegments.length; turnIdx++) {
+    const turnContentEvents = turnSegments[turnIdx];
+    const turnId = crypto.randomUUID();
 
-  logDebug(agentId, `Produced ${records.length} events, turn_id=${turnId}`);
+    // Determine LLM call boundaries within this turn.
+    const llmBoundaries = buildLlmBoundaries(progressEvents, turnContentEvents);
+    logDebug(agentId, `Turn ${turnIdx + 1}: detected ${llmBoundaries.length} LLM call(s)`);
+
+    const turnRecords = buildEventsFromBoundaries(
+      llmBoundaries, turnContentEvents, parsed, turnId, sessionId, agentId, runtimeConfig, cwd,
+    );
+    records.push(...turnRecords);
+
+    logDebug(agentId, `Turn ${turnIdx + 1}: produced ${turnRecords.length} events, turn_id=${turnId}`);
+  }
 
   // --- Phase 5: Write to history ---
   const rowsToAppend = records.map(r => JSON.stringify(r));
@@ -586,6 +597,34 @@ function assignContentToBoundaries(boundaries, contentEvents) {
 }
 
 // --- Helpers -----------------------------------------------------------------
+
+/**
+ * Split a list of content events into turns.
+ * Each real user prompt (type === 'user' and not a tool result) starts a new
+ * turn. Tool results and assistant content following a prompt belong to that
+ * turn until the next real user prompt.
+ */
+function splitContentEventsIntoTurns(contentEvents) {
+  const turns = [];
+  let currentTurn = [];
+
+  for (const row of contentEvents) {
+    if (row.type === 'user' && !isToolResult(row)) {
+      if (currentTurn.length > 0) {
+        turns.push(currentTurn);
+      }
+      currentTurn = [row];
+    } else {
+      currentTurn.push(row);
+    }
+  }
+
+  if (currentTurn.length > 0) {
+    turns.push(currentTurn);
+  }
+
+  return turns;
+}
 
 function findPostToolUseTs(boundaries, currentIdx) {
   if (currentIdx + 1 < boundaries.length) {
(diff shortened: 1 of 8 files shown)


src/inputs/qoder-trace/sqlite-token-reader.ts
```diff
@@ -8,29 +8,48 @@ import { createLogger } from '../../utils/logger.js';
...
+    SELECT
+      cm.id AS message_id,
+      cm.session_id AS session_id,
+      cm.request_id AS request_id,
+      cm.gmt_create AS gmt_create,
+      cm.token_info AS token_info,
+      cm.model_info AS model_info,
+      cr.extra AS record_extra
+    FROM chat_message cm
+    LEFT JOIN chat_record cr ON cr.request_id = cm.request_id
```
ralf0131, line 35: The `LEFT JOIN chat_record` on `request_id` can produce duplicate SQLite rows if `chat_record` ever contains multiple rows for the same `request_id`. Duplicates may inflate token counts or cause assistant-count mismatches in `matchIdeTurnsBySqliteOrder`. Add `DISTINCT` or aggregate `record_extra` per request to guarantee one row per message.

assets/hooks/qoder-hook-processor.mjs
```diff
@@ -298,6 +305,29 @@ function buildLlmBoundaries(progressEvents, contentEvents) {
...
+  for (const pe of progressEvents) {
+    const peMs = Date.parse(pe.ts) || 0;
+    if (peMs <= rowMs) continue;
+    if (pe.hookEvent === 'PreToolUse' || pe.hookEvent === 'Stop') {
+      endTs = pe.ts;
+      break;
+    }
+  }
+
+  return `progress:${startTs || ''}->${endTs || ''}`;
```
ralf0131: `progressWindowKey` returns `progress:->${endTs}` for any assistant row that appears before the first `UserPromptSubmit`/`PostToolUse` event. Rows sharing that null start and the same end boundary will be grouped into the same LLM call, which can merge distinct calls when the transcript begins with assistant content before a start progress event.

src/inputs/qoder-trace/token-enricher.ts
```diff
@@ -216,6 +223,134 @@ export function enrichIdeTurn(
...
+}
+
+function matchIdeTurnsBySqliteOrder(
+  entries: AgentActivityEntry[],
+  requestGroups: Array<[string, SqliteTokenData[]]>,
+  used: Set<AgentActivityEntry>,
+  tokenWritten: Set<string>,
+): void {
+  if (requestGroups.length === 0) return;
+  if (!requestGroups.every(([, rows]) => rows.every(row => row.messageId && row.sessionId))) return;
```
ralf0131, line 252: `matchIdeTurnsBySqliteOrder` bails out entirely if any SQLite row is missing `messageId` or `sessionId`, falling back to timestamp matching for the whole session. A single incomplete row thus disables the more reliable order-based matching for all rows; consider handling partial metadata more leniently.

src/inputs/qoder-trace/token-enricher.ts
```diff
@@ -216,6 +223,134 @@ export function enrichIdeTurn(
...
+    return;
+  }
+
+  const candidateGroups = sessionGroups.slice(sessionGroups.length - turnGroups.length);
+  for (let i = 0; i < turnGroups.length; i++) {
+    const [, turnEntries] = turnGroups[i];
+    const [requestId, sqliteRows] = candidateGroups[i];
+    const responses = turnEntries.filter(e => e['event.name'] === 'llm.response');
+
+    if (responses.length !== sqliteRows.length) {
```
ralf0131, line 276: When `responses.length !== sqliteRows.length` for a turn, the turn is marked low-confidence and skipped entirely, so those responses receive no tokens (they are later zeroed out). Consider a per-row fallback to timestamp matching within that turn to retain whatever token data can be confidently attributed.

ralf0131 (commented):

### Summary
This is a meaningful improvement to Qoder trace ingestion: LLM request/response boundary detection in `qoder-hook-processor.mjs` now uses progress-event windows, user prompts are normalized to `other` (to avoid duplicate span creation), and token enrichment matches SQLite request ordering with confidence tracking. Tests were updated and 31/31 unit tests pass. The changes are largely sound, but there is one cross-cutting correctness gap worth addressing before merge, plus a few robustness suggestions below.

### Findings
- **[Warning]** `src/inputs/qoder-trace/token-enricher.ts:167` — the `userBoundary` lookup still only matches `event.name === 'llm.request'`, but user prompts are now emitted as `other` by `agent-event-normalizer`. As a result the user-boundary entry is `undefined` for IDE turns, so the first `llm.request` timestamp is no longer aligned with the actual user prompt time. Update the predicate to also match `other` events that carry `gen_ai.input.messages_delta`. (Primary blocker.)
- **[Warning]** `src/inputs/qoder-trace/sqlite-token-reader.ts:35` — the `LEFT JOIN` on `request_id` can fan out duplicate rows when `chat_record` has multiple rows per `request_id`, inflating token counts. Consider `DISTINCT` or aggregating per request.
- **[Info]** `assets/hooks/qoder-hook-processor.mjs:328` — a null `progressWindowKey` start can merge distinct LLM calls when a transcript begins with assistant content before the first start-progress event.
- **[Info]** `src/inputs/qoder-trace/token-enricher.ts:250` — order-based matching bails out for the whole session on a single incomplete SQLite row; a partial-metadata fallback would be more resilient.
- **[Info]** `src/inputs/qoder-trace/token-enricher.ts:274` — when response/SQLite row counts differ for a turn, the whole turn is zeroed; a per-row timestamp fallback would preserve attributable token data.

### Suggestions
1. Fix the `userBoundary` predicate (finding #1) so the new `other` event path keeps timestamp alignment intact — this is the main consistency issue introduced by the normalizer change.
2. Make the SQLite join row-count-safe (finding #2) since token accounting feeds cost/usage dashboards downstream.

### Cross-repo Note
The Qoder IDE hook (`assets/hooks/*.mjs`) and the Node trace input (`src/inputs/qoder-trace/*.ts`) are coupled by the `event.name` contract; a change in the event type emitted by the hook must be reflected in every consumer in `token-enricher.ts`. Worth a follow-up pass over all `event.name === 'llm.request'` predicates to ensure they align with the new `other` mapping.

---

ralf0131 (commented):
### Summary
Re-review after two new commits (`2e27e541`, `51d552ca`) since the previous review at `328116e0`. The new commits address issues raised in the first review and add a cold-start fix. Boundary detection logic and token enrichment are now more robust.

### Follow-up on Previous Findings
- ✅ **[Resolved]** `token-enricher.ts` — `userBoundary` predicate now correctly matches `other` events carrying `gen_ai.input.messages_delta` in addition to `llm.request`. This was the primary blocker from the previous review.
- ✅ **[Resolved]** `qoder-hook-processor.mjs` — `progressWindowKey` now returns `null` when no start boundary is found, letting the caller fall back to time-gap grouping instead of incorrectly merging distinct LLM calls.
- ✅ **[Resolved]** Cold start replay — new commit `51d552ca` prevents full history replay when state is empty, with integration test coverage.

### Remaining Concern
- ⚠️ **[Still Open]** `sqlite-token-reader.ts:35` — The `LEFT JOIN` on `request_id` can still fan out duplicate rows when `chat_record` has multiple rows per `request_id`, inflating token counts. This file was not modified in the new commits. If this is intentional (e.g., `chat_record` has a unique constraint on `request_id`), consider adding a clarifying comment. Otherwise, `DISTINCT` or per-request aggregation would prevent inflation.

### Suggestions
The fixes are solid and the new test coverage is thorough. The only remaining concern is the `LEFT JOIN` fan-out in `sqlite-token-reader.ts`. If that is confirmed safe, the PR looks good to merge.

---

src/inputs/qoder-trace/token-enricher.ts
```diff
@@ -216,6 +225,134 @@ export function enrichIdeTurn(
...
+  (response as Record<string, unknown>).__matched_gmt_create = row.gmtCreate;
+  response['gen_ai.request.id'] = requestId;
+  (response as Record<string, unknown>)['agent.request_id'] = requestId;
+  response['gen_ai.response.id'] = row.messageId || requestId;
+  (response as Record<string, unknown>)['qoder.match_confidence'] = 'high';
+  (response as Record<string, unknown>)['qoder.match.strategy'] = 'sqlite_request_order';
+
+  if (row.model && row.model !== 'unknown') {
+    response['gen_ai.request.model'] = row.model;
+    response['gen_ai.response.model'] = row.model;
```
linrunqi08, line 318: **Bug (PLAUSIBLE): Request fallback in `applySqliteRowToIdeResponse` can corrupt multi-step turns**

```ts
const request = findStepRequest(allEntries, response) ?? turnEntries.find(e => e['event.name'] === 'llm.request');
```

When `findStepRequest` fails to find a request with a matching `step_id`, the fallback always picks the **first** `llm.request` in the turn. In a multi-step turn with N responses where step_id matching fails for more than one, each response overwrites the same first request's `gen_ai.request.id`, `agent.request_id`, `gen_ai.request.model`, and confidence metadata with different values.

**Failure scenario**: IDE turn with 3 LLM steps where step_id is not set on entries (e.g., legacy data). Response 1 sets `request.gen_ai.request.id = "req-A"`, response 2 overwrites it with `"req-B"`, response 3 overwrites with `"req-C"`. The request ends up attributed to `req-C` only, and req-A/req-B lose their request association.

**Suggested fix**: Track which requests have already been matched (similar to the `used` set for responses) to prevent the same request from being reassigned.

src/inputs/qoder-trace/token-enricher.ts
```diff
@@ -216,6 +225,134 @@ export function enrichIdeTurn(
...
+      [...rows].sort((a, b) => a.gmtCreate - b.gmtCreate),
+    ] as [string, SqliteTokenData[]])
+    .sort((a, b) => a[1][0].gmtCreate - b[1][0].gmtCreate);
+}
+
+function matchIdeTurnsBySqliteOrder(
+  entries: AgentActivityEntry[],
+  requestGroups: Array<[string, SqliteTokenData[]]>,
+  used: Set<AgentActivityEntry>,
+  tokenWritten: Set<string>,
```
linrunqi08, line 249: **Design concern: All-or-nothing guard on metadata completeness is overly conservative**

```ts
if (!requestGroups.every(([, rows]) => rows.every(row => row.messageId && row.sessionId))) return;
```

If **any single** SQLite row across **any** request group has an empty `messageId` or `sessionId` (which defaults to `''` from `sqlite-token-reader.ts` when the DB value is null), the **entire** structural matching path is abandoned — even for turns where the data is complete.

**Failure scenario**: 10 request groups, 9 have full metadata, 1 historical row has null `message_id` in the DB. All 10 groups fall back to timestamp-only matching, losing the high-confidence structural matches for the 9 good groups.

**Suggestion**: Filter out incomplete rows/groups before matching rather than bailing out entirely. Or at minimum, make the guard per-group rather than global.

src/inputs/qoder-trace/sqlite-token-reader.ts
```diff
@@ -8,29 +8,48 @@ import { createLogger } from '../../utils/logger.js';
...
+    WHERE cm.session_id = ?
```
linrunqi08, line 36: **Bug (PLAUSIBLE): LEFT JOIN with `chat_record` can multiply result rows**

```sql
LEFT JOIN chat_record cr ON cr.request_id = cm.request_id
```

If `chat_record` has multiple rows per `request_id` (no uniqueness constraint visible), each `chat_message` row gets duplicated for each matching `chat_record` row. The downstream `groupSqliteRowsByRequest` would create multiple entries per request with the same token data but potentially different `model` values from different `chat_record.extra` rows.

**Failure scenario**: A `request_id` has 2 `chat_record` rows (e.g., retry or update). The `chat_message` row is returned twice. `groupSqliteRowsByRequest` puts both in the same group (same `requestId`). The enricher matches 2 SQLite rows against 1 hook response → `assistant_count_mismatch` → low confidence fallback.

**Suggestion**: If `chat_record.request_id` is not unique, add `GROUP BY cm.id` or `LIMIT 1` on the join to pick one `chat_record` row per message. Or use a subquery to select the latest `chat_record` per `request_id`.

assets/hooks/qoder-hook-processor.mjs
```diff
@@ -298,6 +316,34 @@ function buildLlmBoundaries(progressEvents, contentEvents) {
   return boundaries;
 }
 
+function progressWindowKey(progressEvents, rowMs) {
```
linrunqi08, line 319: **Efficiency: `progressWindowKey` has O(N×M) complexity**

This function does two linear scans of `progressEvents` (M elements) and is called once per assistant content row (N rows) inside `buildLlmBoundaries`. Total: O(N × M).

For a long IDE session with hundreds of tool calls (large M) and many assistant blocks (large N), this becomes noticeable.

**Suggestion**: Preprocess `progressEvents` into sorted intervals once before the loop in `buildLlmBoundaries`. Then use binary search to find the containing interval for each row timestamp. This reduces per-row cost from O(M) to O(log M).

assets/hooks/qoder-hook-processor.mjs
```diff
@@ -609,7 +683,7 @@ function buildLegacyEvents(contentEvents, turnId, sessionId, agentId, runtimeCon
   for (const record of existingRecords) {
     const eventName = record['event.name'];
     const rawType = record['agent.qoder.raw_type'];
-    if (eventName === 'llm.request' && rawType === 'user') continue;
+    if (rawType === 'user') continue;
     if (eventName === 'llm.response') {
       const responseTs = record['time_unix_nano'] || '';
       const tsDiff = lastResponseTs === null ? Infinity : Math.abs(Number(responseTs) - Number(lastResponseTs));
```
linrunqi08, line 689: **Bug (PLAUSIBLE): Broadened filter skips `tool.result` step.id assignment in legacy mode**

Old:
```js
if (eventName === 'llm.request' && rawType === 'user') continue;
```
New:
```js
if (rawType === 'user') continue;
```

The old filter only skipped user-hook events (`llm.request` + user) for step.id assignment. The new filter skips **all** user-typed records, including `tool.result` events from user rows (where `inferQoderEventName` returns `'tool.result'` for content with `type === 'tool_result'`).

**Failure scenario**: Legacy mode transcript with tool results. `buildQoderHookRecord` generates `tool.result` events with `rawType === 'user'`. The new filter skips these for step.id assignment → tool.result events have no `gen_ai.step.id` → they become orphaned in trace/span construction.

**Note**: This may be intentional if the old behavior was already wrong (tool.result from user rows getting incorrect step.ids), but worth confirming.

src/inputs/qoder-trace/token-enricher.ts
```diff
@@ -216,6 +225,134 @@ export function enrichIdeTurn(
...
+      continue;
+    }
+
+    for (let j = 0; j < responses.length; j++) {
+      applySqliteRowToIdeResponse(entries, turnEntries, responses[j], sqliteRows[j], requestId, used, tokenWritten);
+    }
+  }
+}
+
+function groupEntriesByTurn(entries: AgentActivityEntry[]): Array<[string, AgentActivityEntry[]]> {
```
linrunqi08, line 287: **Bug (PLAUSIBLE): Positional matching assumes responses are in chronological order, but they are not explicitly sorted**

```ts
const responses = turnEntries.filter(e => e['event.name'] === 'llm.response');
// ...
for (let j = 0; j < responses.length; j++) {
  applySqliteRowToIdeResponse(..., responses[j], sqliteRows[j], ...);
}
```

`responses` preserves the insertion order from `entries`, while `sqliteRows` are explicitly sorted by `gmtCreate ASC` (via `groupSqliteRowsByRequest`). The positional matching `responses[j] ↔ sqliteRows[j]` assumes both are in the same chronological order.

**Failure scenario**: If entries within a turn are not in chronological order (e.g., due to buffered/concurrent writes to JSONL), the j-th response by file order may not correspond to the j-th SQLite row by time order. Token data and request IDs would be assigned to the wrong responses.

**Suggestion**: Sort `responses` by `time_unix_nano` before positional matching to guarantee alignment with the time-sorted SQLite rows.

linrunqi08 (commented):
## Code Review Summary

Reviewed with high effort across correctness, cross-file tracing, efficiency, and altitude angles.

**Overall**: The PR significantly improves LLM boundary detection and token enrichment. The turn-splitting, progress-window keying, and SQLite structural matching are well-designed. However, I found several issues ranging from potential data corruption in edge cases to an efficiency concern.

6 findings below as inline comments, ranked by severity.

ralf0131 (commented):
### Summary
Re-review after new commit `cb52d9ad` (\u201cfix(qoder-hook): split multi-turn transcripts into separate turns\u201d). The multi-turn splitting fix is correct and well-tested.

### Incremental Assessment
- \u2705 **Multi-turn splitting**: `splitContentEventsIntoTurns` correctly splits content events by real user prompts, ensuring each turn gets its own `turnId` and correct user input. The `isToolResult` guard properly keeps tool results attached to the preceding turn.
- \u2705 **Progress event scoping**: Passing all `progressEvents` to `buildLlmBoundaries` per turn is safe \u2014 the function uses timestamps from turn-scoped `contentEvents` to match progress events, so cross-turn contamination cannot occur.
- \u2705 **Test coverage**: New integration test verifies 2 distinct turns with correct user inputs (\u201chello\u201d vs \u201c\u5b8c\u6210\u529b\u625a\u7b2c143\u9898\u201d) and distinct turn IDs.

### Still Open (from previous review)
- \u26a0\ufe0f **`sqlite-token-reader.ts` LEFT JOIN fan-out**: The `LEFT JOIN` on `request_id` can still produce duplicate rows when `chat_record` has multiple rows per request, inflating token counts. This was flagged in the first review and has not been addressed yet.

No blocking issues in the incremental changes. The multi-turn fix is ready to merge once the remaining SQLite JOIN concern is resolved.

---

src/inputs/qoder-trace/token-enricher.ts
rangemer333-cell: 不需要修复。 这个 fallback 是针对历史数据的防御性兜底：

当前代码保证所有新产出的 JSONL 记录都有 gen_ai.step.id
fallback 只会在 step_id 缺失的极端情况下触发，而这种情况下 turn 内通常只有一个 request，取第一个是正确的
多 step turn 不会触发 fallback（step_id 一定存在）
如果要加固，可以在 fallback 触发时打 low confidence 标记（表明是兜底匹配），但这属于防御性增强而非 bug 修复，优先级低。

src/inputs/qoder-trace/token-enricher.ts
rangemer333-cell: 不需要修复。 理由：

触发条件在现实中极低概率（SQLite 主键不会为空）
即使触发，时间戳 fallback 路径保证数据不丢失
影响范围仅是 match_confidence 标记从 high 降级为无标记，不影响 token 数据正确性
可以作为低优先级的防御性增强（per-group filter + low confidence 标记），但不是 bug fix。

src/inputs/qoder-trace/sqlite-token-reader.ts
rangemer333-cell: chat_record 表的 request_id 是 primary key。
主键保证了唯一性。一个 request_id 在 chat_record 中最多只有一行。LEFT JOIN 不会产生行膨胀。

这个 concern 的前提假设（"request_id 无唯一约束，可能有多行"）不成立。实际 schema 中 request_id 就是主键，JOIN 结果是严格 1:1（或 1:0 当 chat_record 无对应行时）。

assets/hooks/qoder-hook-processor.mjs
rangemer333-cell: 不需要修复。 N 和 M 的实际规模决定了 O(N×M) 在这里是 O(千) 级别，耗时 <1ms，不构成性能问题。hook 处理器的真正瓶颈在文件 IO，而非内存中的数组遍历。

assets/hooks/qoder-hook-processor.mjs
rangemer333-cell: 不需要修复。 新的 if (rawType === 'user') continue; 比旧代码更安全：

旧代码会给 tool.result 分配可能错误的 step.id（因为 stepCounter 基于 response 时间推进，tool.result 出现在 response 之后会被错误归入下个 step）
新代码让 tool.result 在 legacy 模式下没有 step.id，避免了错误关联
Legacy 路径本身就是降级模式（注释写了 "Limitations: no llm.request synthesis"），不追求完美的 span 关联

src/inputs/qoder-trace/token-enricher.ts
rangemer333-cell: 链路中的顺序保证
hook-processor 写入顺序：buildEventsFromBoundaries 按 boundaries 顺序（chronological）循环，每个 boundary 先输出 llm.request，再输出 llm.response，再输出 tool.call/tool.result。records 数组的 response 顺序一定是时间递增的。

JSONL 文件顺序 = records 数组顺序：line 218 records.map(r => JSON.stringify(r)) 直接按数组顺序写入文件。

读取顺序 = 文件顺序：qoder-trace-input.ts 逐行 parse，entries 数组保持文件行顺序。

groupEntriesByTurn 保持插入顺序：Map 的 set 保持键的插入顺序，数组 push 保持元素顺序。filter 也保持原数组顺序。

结论：turnEntries.filter(e => e['event.name'] === 'llm.response') 的结果顺序和 sqliteRows（按 gmtCreate ASC 排序）是一致的——都是时间递增。

linrunqi08 (approved):

PR #47, opened 2026-06-18
fix(windows): fix .sh→.ps1 replacement when hookCommand has trailing args

linrunqi08:
## Summary
- Fix regex in `agent-def-loader.ts` from `/\.sh$/` to `/\.sh(?=\s|$)/` so that `.sh→.ps1` replacement works when `hookCommand` has trailing arguments (e.g. `hook.sh qoder`)
- Without this fix, on Windows the `.sh` path is written to settings.json as-is, causing the hook to never fire

## Root cause
`qoder.json` defines `hookCommand` as `$PILOT_DATA/hooks/qoder-loongsuite-pilot-hook.sh qoder`. The `$` anchor in the old regex requires `.sh` at the very end of the string, but here `.sh` is followed by ` qoder`, so the replacement silently fails.

## Test plan
- [x] Verified regex handles both cases: `hook.sh` (no args) and `hook.sh qoder` (with args)
- [ ] Deploy to a Windows machine and confirm `~/.qoder/settings.json` gets the `.ps1` command

### DIFF
--- a/src/deployment/agent-def-loader.ts
+++ b/src/deployment/agent-def-loader.ts
@@ -126,7 +126,7 @@ export class AgentDefLoader {
     result = resolveHome(result);
 
     if (process.platform === 'win32') {
-      result = result.replace(/\.sh$/, '.ps1');
+      result = result.replace(/\.sh(?=\s|$)/, '.ps1');
     }
     return result;
   }


src/deployment/agent-def-loader.ts
```diff
@@ -126,7 +126,7 @@ export class AgentDefLoader {
     result = resolveHome(result);
 
     if (process.platform === 'win32') {
-      result = result.replace(/\.sh$/, '.ps1');
+      result = result.replace(/\.sh(?=\s|$)/, '.ps1');
```
ralf0131, line 129: Nice fix — the lookahead `(?=\s|$)` correctly handles both `hook.sh` (no args) and `hook.sh qoder` (with args), where the old `$` anchor silently missed the latter. Small follow-up: the existing `tests/unit/deployment/agent-def-loader.test.ts` covers `$PILOT_DIR`/`$PILOT_DATA` variable replacement but has no test for this Windows `.sh→.ps1` branch. Consider adding a case (e.g. a mocked `process.platform === 'win32'` assertion) so this regression stays covered.

ralf0131 (approved):

### Summary
Correct, minimal fix: the Windows `.sh→.ps1` replacement regex is widened from `/\.sh$/` to `/\.sh(?=\s|$)/` so a `hookCommand` like `hook.sh qoder` (trailing args) now gets the `.ps1` rewrite instead of silently failing. The root-cause analysis is clear and accurate, and the change is correctly scoped to the `process.platform === 'win32'` branch. Approving.

### Findings
- **[Info]** `src/deployment/agent-def-loader.ts:129` — lookahead fix is correct; consider adding a unit test for this Windows branch (see inline comment).

---

messixukejia (approved):

PR #48, opened 2026-06-22
fix(qoder-work): strip version field from output entries

Snssn:
## Summary
- The raw `version` field from qoderwork hook records was leaking into output entries via the PostToolUse legacy path (`extra` → `agent.version`), appearing in JSONL output and SLS reporting.
- Delete `record['version']` after extracting agent version info, while preserving `agent.qoderwork.version` for internal tracking.
- Added 2 test cases to verify the fix.

## Test plan
- [x] Unit tests pass (19/19 in `qoder-work-input.test.ts`)
- [x] New test: `agent.qoderwork.version` is preserved in canonical entries
- [x] New test: plain `version` field is stripped from PostToolUse output entries


linrunqi08 (approved):

ralf0131 (approved):
LGTM. Simple and correct fix — strips the stray `version` field from the raw record before building the canonical entry, preventing it from leaking into the output. Test properly verifies both `agent.qoderwork.version` preservation and `version` removal.

---

PR #53, opened 2026-06-22
fix(deploy): avoid SIGPIPE when previewing package contents

YaoYingLong:
The script deploy/package-opensource.sh enables strict mode via set -euo pipefail.

At the end of the script, the command tar -tzf "$OUTPUT_PATH" | head -20 is executed.

In Linux environments, head -20 closes the pipe immediately after printing the first 20 lines. This sends a SIGPIPE signal to the upstream tar process, making tar exit with code 141.

Due to set -o pipefail, the entire pipeline returns a non-zero exit code, which triggers set -e to terminate the whole script and results in the build failure.

### DIFF
--- a/deploy/package-opensource.sh
+++ b/deploy/package-opensource.sh
@@ -138,7 +138,7 @@ echo "    ✅ $ZIP_OUTPUT_PATH ($ZIP_SIZE)"
 # ── Summary ──
 echo ""
 echo "==> Contents:"
-tar -tzf "$OUTPUT_PATH" | head -20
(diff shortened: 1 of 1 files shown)


ralf0131 (approved):
Good fix. Replacing `head -20` with `sed -n '1,20p'` avoids the SIGPIPE that `head` triggers by closing its stdin early — under `set -o pipefail` `tar` would exit non-zero (141) and abort the script. `sed -n` drains the full stream so `tar` completes cleanly. LGTM.

---

PR #56, opened 2026-06-23
fix(opencode): preserve turnSeq + pre-create plugin-inject log dirs

fangxiu-wf:
## Summary

- Fix OpenCode plugin `clearSession()` wiping `turnSeq` on every `session.idle`, causing all subsequent conversations to reuse `gen_ai.turn.id = ses_xxx:t1`
- This led to the OTLP trace flusher dropping 2nd+ conversations as "late arrivals" for an already-flushed turn key
- Persist `turnSeq` in a separate LRU-bounded map (`sessionTurnSeqs`) so the counter monotonically increases within a process lifetime
- Pre-create log directories for plugin-inject agents (opencode, qwen-code-cli) to avoid 5-minute discovery delay after fresh install

## Root Cause

1. `session.idle` fires between conversations → `clearSession()` deletes the entire session object
2. Next `getSession()` creates a fresh session with `turnSeq: 0` → `handleChatMessage` increments to 1 → always generates `ses_xxx:t1`
3. Flusher's `flushedTurnKeys` set already contains `turn:ses_xxx:t1` from the first conversation → subsequent events are dropped
4. (Secondary) After `--purge` reinstall, plugin-inject agent log dirs don't exist → `fs.watch` fails → 5min polling fallback

## Test plan

- [x] Trigger 3+ conversations in a single OpenCode session
- [x] Verify `~/.loongsuite-pilot/logs/opencode/` shows distinct `gen_ai.turn.id` values (`:t1`, `:t2`, `:t3`)
- [x] Verify `~/.loongsuite-pilot/logs/otlp-debug/` produces 3 separate traces with unique trace IDs
- [x] Unit tests pass: `vitest run tests/unit/hooks/opencode-plugin-session.test.mjs` (5 tests)

### DIFF
--- a/assets/plugins/opencode/plugin.mjs
+++ b/assets/plugins/opencode/plugin.mjs
@@ -183,13 +183,14 @@ function writeError(source, err) {
 // ---------------------------------------------------------------------------
 
 const sessions = new Map();
+const sessionTurnSeqs = new Map();
 
 function getSession(sessionID) {
   if (!sessionID) return null;
   let s = sessions.get(sessionID);
   if (!s) {
     s = {
-      turnSeq: 0,
+      turnSeq: sessionTurnSeqs.get(sessionID) || 0,
       currentTurn: null,
       systemPrompt: null,
       systemInstructionsParts: null,
@@ -211,6 +212,14 @@ function getSession(sessionID) {
 }
 
 function clearSession(sessionID) {
+  const s = sessions.get(sessionID);
+  if (s) {
+    sessionTurnSeqs.set(sessionID, s.turnSeq);
+    if (sessionTurnSeqs.size > MAX_SESSIONS) {
+      const oldest = sessionTurnSeqs.keys().next().value;
+      sessionTurnSeqs.delete(oldest);
+    }
+  }
   sessions.delete(sessionID);
 }
 
--- a/assets/plugins/opencode/plugin.mjs
+++ b/assets/plugins/opencode/plugin.mjs
@@ -214,6 +214,7 @@ function getSession(sessionID) {
 function clearSession(sessionID) {
   const s = sessions.get(sessionID);
   if (s) {
+    sessionTurnSeqs.delete(sessionID);
     sessionTurnSeqs.set(sessionID, s.turnSeq);
     if (sessionTurnSeqs.size > MAX_SESSIONS) {
       const oldest = sessionTurnSeqs.keys().next().value;
--- a/assets/plugins/opencode/plugin.mjs
+++ b/assets/plugins/opencode/plugin.mjs
@@ -190,7 +190,7 @@ function getSession(sessionID) {
   let s = sessions.get(sessionID);
   if (!s) {
     s = {
-      turnSeq: sessionTurnSeqs.get(sessionID) || 0,
+      turnSeq: sessionTurnSeqs.get(sessionID) ?? 0,
       currentTurn: null,
       systemPrompt: null,
       systemInstructionsParts: null,
--- a/assets/plugins/opencode/plugin.mjs
+++ b/assets/plugins/opencode/plugin.mjs
@@ -205,7 +205,7 @@ function getSession(sessionID) {
     sessions.set(sessionID, s);
     if (sessions.size > MAX_SESSIONS) {
       const oldest = sessions.keys().next().value;
-      sessions.delete(oldest);
+      clearSession(oldest);
     }
   }
   return s;
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -857,9 +857,11 @@ export class Orchestrator extends EventEmitter {
     );
 
     // --- OpenCode Log (event_t plugin JSONL) ---
-    const opencodeLogDir = path.join(this.dataDir, 'logs', 'opencode');
-    // Pre-create log dir so fs.watch in AgentDiscoveryService succeeds immediately,
+    // Plugin-inject agents (opencode, qwen-code-cli) don't create their log dirs
+    // during hook deployment (unlike cursor/claude/codex whose shell hooks mkdir -p).
+    // Pre-create here so fs.watch in AgentDiscoveryService succeeds immediately,
     // avoiding a 5-minute polling fallback delay after fresh install with --purge.
+    const opencodeLogDir = path.join(this.dataDir, 'logs', 'opencode');
     await ensureDir(opencodeLogDir);
     const opencodeLogInput = new OpenCodeLogInput({
       stateStore: this.stateStore,
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -858,6 +858,9 @@ export class Orchestrator extends EventEmitter {
 
     // --- OpenCode Log (event_t plugin JSONL) ---
     const opencodeLogDir = path.join(this.dataDir, 'logs', 'opencode');
+    // Pre-create log dir so fs.watch in AgentDiscoveryService succeeds immediately,
+    // avoiding a 5-minute polling fallback delay after fresh install with --purge.
+    await ensureDir(opencodeLogDir);
     const opencodeLogInput = new OpenCodeLogInput({
       stateStore: this.stateStore,
       logDir: opencodeLogDir,
@@ -878,6 +881,8 @@ export class Orchestrator extends EventEmitter {
 
     // --- Qwen Code CLI Log (transcript-driven hook JSONL) ---
     const qwenCodeCliLogDir = path.join(this.dataDir, 'logs', 'qwen-code-cli');
+    // Pre-create log dir so fs.watch in AgentDiscoveryService succeeds immediately.
+    await ensureDir(qwenCodeCliLogDir);
     const qwenCodeCliLogInput = new QwenCodeCliLogInput({
       stateStore: this.stateStore,
       logDir: qwenCodeCliLogDir,
--- a/tests/unit/hooks/opencode-plugin-session.test.mjs
+++ b/tests/unit/hooks/opencode-plugin-session.test.mjs
@@ -1,0 +1,142 @@
+import { describe, expect, it, beforeEach } from 'vitest';
+import { createRequire } from 'node:module';
+import { fileURLToPath } from 'node:url';
+import path from 'node:path';
+
+const PLUGIN_PATH = path.resolve(
+  fileURLToPath(import.meta.url),
+  '../../../../assets/plugins/opencode/plugin.mjs',
+);
+
+/**
+ * Extract session management functions from plugin.mjs for unit testing.
+ * Since the plugin exports only the default hook object, we re-implement
+ * the session logic here mirroring the source to test the invariants.
+ */
+
+const MAX_SESSIONS = 100;
+
+function createSessionManager() {
+  const sessions = new Map();
+  const sessionTurnSeqs = new Map();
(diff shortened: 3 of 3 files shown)


assets/plugins/opencode/plugin.mjs
```diff
@@ -183,13 +183,14 @@ function writeError(source, err) {
...
 const sessions = new Map();
+const sessionTurnSeqs = new Map();
 
 function getSession(sessionID) {
   if (!sessionID) return null;
   let s = sessions.get(sessionID);
   if (!s) {
     s = {
-      turnSeq: 0,
+      turnSeq: sessionTurnSeqs.get(sessionID) || 0,
```
linrunqi08: **Bug: LRU eviction in `getSession` bypasses turnSeq persistence**

The fix correctly saves `turnSeq` in `clearSession()`, but the pre-existing LRU eviction in `getSession()` (lines 206-209) still calls `sessions.delete(oldest)` directly — never calling `clearSession()`. When `sessions` reaches `MAX_SESSIONS` and a new session arrives, the oldest session is silently evicted and its `turnSeq` is lost.

This means the PR's invariant ("turnSeq monotonically increases within a process lifetime") can be violated:
1. 100+ concurrent sessions → `getSession` LRU-evicts session A (turnSeq=5)
2. Session A's turnSeq is NOT saved to `sessionTurnSeqs`
3. Later `session.idle` fires for A → `clearSession(A)` → `sessions.get(A)` is undefined → nothing saved
4. A reconnects → `getSession(A)` → `sessionTurnSeqs.get(A)` returns undefined → turnSeq resets to 0 → same trace-loss bug

**Suggested fix**: replace the direct `sessions.delete(oldest)` with `clearSession(oldest)` in the LRU eviction block of `getSession()`, so turnSeq is always preserved before eviction:
```js
if (sessions.size > MAX_SESSIONS) {
  const oldest = sessions.keys().next().value;
  clearSession(oldest);  // saves turnSeq before deleting
}
```

src/core/orchestrator.ts
```diff
@@ -858,6 +858,9 @@ export class Orchestrator extends EventEmitter {
 
     // --- OpenCode Log (event_t plugin JSONL) ---
     const opencodeLogDir = path.join(this.dataDir, 'logs', 'opencode');
+    // Pre-create log dir so fs.watch in AgentDiscoveryService succeeds immediately,
+    // avoiding a 5-minute polling fallback delay after fresh install with --purge.
+    await ensureDir(opencodeLogDir);
```
linrunqi08, line 865: **Incomplete: only 2 of 9 log dirs get `ensureDir`**

The `ensureDir` fix is applied to `opencodeLogDir` and `qwenCodeCliLogDir`, but 7 other log dirs use the same `watchPaths: [xxxLogDir]` + `isAvailable: () => directoryExists(xxxLogDir)` detection pattern:
- `qoderCnLogDir` (line 586)
- `qoderWorkLogDir` (line 602)
- `qoderWorkCNLogDir` (line 658)
- `qoderCliLogDir` (line 747)
- `cursorHookLogDir` (line 800)
- `claudeCodeLogDir` (line 820)
- `codexLogDir` (line 840)

After `--purge`, these dirs would also be missing and `fs.watch` would fail, causing the same 5-minute polling fallback delay the comment describes. Consider applying the same `ensureDir` treatment to all log dirs for consistency, or adding a note explaining why only these two need it.

linrunqi08 (commented):
## Code Review — 2 findings

Overall the fix is well-scoped and the root cause analysis is spot-on. Two issues worth addressing before merge:

1. **Bug**: LRU eviction in `getSession` bypasses turnSeq persistence — see inline comment on `plugin.mjs`
2. **Incomplete**: `ensureDir` only covers 2 of 9 log dirs with the same `watchPaths` detection pattern — see inline comment on `orchestrator.ts`

ralf0131 (commented):
### Summary
PR #56 fixes a real trace-loss bug: `clearSession()` (on `session.idle`) wiped `turnSeq`, so every subsequent conversation reused `gen_ai.turn.id = ses_xxx:t1` and the OTLP flusher dropped them as late arrivals. The fix persists `turnSeq` in a separate LRU-bounded `sessionTurnSeqs` map — the core approach is correct and well-scoped.

@linrunqi08 already raised two solid points I agree with: (1) the pre-existing `getSession()` LRU eviction calls `sessions.delete(oldest)` directly, bypassing turnSeq persistence — should call `clearSession(oldest)`; (2) only 2 of 9 log dirs get `ensureDir()` treatment. I will not repeat those.

A few additional observations below.

---

assets/plugins/opencode/plugin.mjs
```diff
@@ -211,6 +212,14 @@ function getSession(sessionID) {
 }
 
 function clearSession(sessionID) {
+  const s = sessions.get(sessionID);
+  if (s) {
+    sessionTurnSeqs.set(sessionID, s.turnSeq);
```
ralf0131, line 218: The new sessionTurnSeqs 'LRU' does not update key recency when a session is reused. Map.set() on an existing key preserves insertion order, so a session that is cleared, reactivated, and cleared again stays at its original position and can be evicted while less-recently-used sessions survive. This can reset turnSeq and cause gen_ai.turn.id reuse. Fix by deleting the existing entry before set() to move it to the most-recent position.

assets/plugins/opencode/plugin.mjs
```diff
@@ -183,13 +183,14 @@ function writeError(source, err) {
 // ---------------------------------------------------------------------------
 
 const sessions = new Map();
+const sessionTurnSeqs = new Map();
```
ralf0131, line 186: No automated tests are added for the new turnSeq persistence behavior. Add unit tests covering: (1) turnSeq is restored after session.idle + reactivation, (2) turnSeq remains monotonic across multiple idle cycles, and (3) LRU boundary behavior when MAX_SESSIONS is exceeded.

src/core/orchestrator.ts
```diff
@@ -878,6 +881,8 @@ export class Orchestrator extends EventEmitter {
 
     // --- Qwen Code CLI Log (transcript-driven hook JSONL) ---
```
ralf0131, line 884: The qwen-code-cli ensureDir() change is outside the stated PR scope of fixing OpenCode turnSeq persistence. Consider splitting unrelated log-dir pre-creation fixes into a separate change, or update the PR title/description to cover both agents.

assets/plugins/opencode/plugin.mjs
fangxiu-wf: 已修复。`getSession` 中 LRU 淘汰路径现在改用 `clearSession(oldest)` 而非直接 `sessions.delete(oldest)`，确保 turnSeq 通过 `sessionTurnSeqs` 持久化后再移除 session。

见 commit `035877f`。

src/core/orchestrator.ts
fangxiu-wf: 已在代码中添加注释说明：只有 `plugin-inject` 模式的 agent（opencode, qwen-code-cli）需要在此处 `ensureDir`，原因是它们不经过 shell hook 部署阶段（不像 cursor/claude/codex 的 hook 脚本会 `mkdir -p` 创建日志目录）。

其余 7 个 input 要么：
1. 使用 `hook` 模式部署，shell hook 中已创建目录
2. 监听的 `watchPath` 与 `logDir` 不同（如 cursor 监听 `~/.cursor/`）

见 commit `3b0bd31` 和 `035877f` 中 `orchestrator.ts` 的注释。

assets/plugins/opencode/plugin.mjs
fangxiu-wf: 已修复。`clearSession()` 中现在先 `sessionTurnSeqs.delete(sessionID)` 再 `sessionTurnSeqs.set(sessionID, ...)`, 确保 key 被移动到 Map 末尾（最新位置），维护正确的 LRU 淘汰顺序。

见 commit `035877f`。

assets/plugins/opencode/plugin.mjs
fangxiu-wf: 已添加。新增 `tests/unit/hooks/opencode-plugin-session.test.mjs`，覆盖以下场景：

1. ✅ turnSeq 在 session.idle 后恢复
2. ✅ turnSeq 跨多次 idle 周期单调递增
3. ✅ LRU 淘汰通过 clearSession 保留 turnSeq
4. ✅ sessionTurnSeqs 的 LRU delete-before-set 维护 key 新鲜度
5. ✅ 不同 session 间 turnSeq 独立

所有 5 个测试已通过。见 commit `035877f`。

src/core/orchestrator.ts
fangxiu-wf: 已更新 PR 标题和描述以反映完整范围。qwen-code-cli 与 opencode 属于同一类 `plugin-inject` agent，共享完全相同的目录发现问题，一并修复更合理。

PR 标题已改为：`fix(opencode): preserve turnSeq + pre-create plugin-inject log dirs`

src/core/orchestrator.ts
```diff
@@ -857,7 +857,12 @@ export class Orchestrator extends EventEmitter {
     );
 
     // --- OpenCode Log (event_t plugin JSONL) ---
+    // Plugin-inject agents (opencode, qwen-code-cli) don't create their log dirs
+    // during hook deployment (unlike cursor/claude/codex whose shell hooks mkdir -p).
+    // Pre-create here so fs.watch in AgentDiscoveryService succeeds immediately,
+    // avoiding a 5-minute polling fallback delay after fresh install with --purge.
     const opencodeLogDir = path.join(this.dataDir, 'logs', 'opencode');
+    await ensureDir(opencodeLogDir);
```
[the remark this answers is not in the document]
fangxiu-wf: 不采纳。项目中 ensureDir 的实际实现是 async 的（fs-utils.ts:89）：使用 await fsp.mkdir(path, { recursive: true })，返回 Promise<void>，await 完全正确。同文件其他位置也均使用 await ensureDir(...)。

src/core/orchestrator.ts
```diff
@@ -878,6 +883,8 @@ export class Orchestrator extends EventEmitter {
 
     // --- Qwen Code CLI Log (transcript-driven hook JSONL) ---
     const qwenCodeCliLogDir = path.join(this.dataDir, 'logs', 'qwen-code-cli');
+    // Pre-create log dir so fs.watch in AgentDiscoveryService succeeds immediately.
+    await ensureDir(qwenCodeCliLogDir);
```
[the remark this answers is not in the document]
fangxiu-wf: 同上，ensureDir 是 async 函数，await 正确。

assets/plugins/opencode/plugin.mjs
```diff
@@ -183,13 +183,14 @@ function writeError(source, err) {
...
 
 function getSession(sessionID) {
   if (!sessionID) return null;
   let s = sessions.get(sessionID);
   if (!s) {
     s = {
-      turnSeq: 0,
+      turnSeq: sessionTurnSeqs.get(sessionID) || 0,
       currentTurn: null,
```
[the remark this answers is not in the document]
fangxiu-wf: 已采纳，改为 `?? 0`。语义更精确，避免对假设性 falsy 值的误判。见 commit 2a2de1e。

linrunqi08 (approved):

PR #57, opened 2026-06-23
fix(codex): improve interrupted turn trace recovery

rangemer333-cell:
## Summary
- preserve interrupted Codex turn ordering and correct assistant-message/token step attribution
- add local diagnostics for completed turns where Codex Hook state is absent, without trace fallback
- serialize tool.result.status in normalized JSONL output

## Test Plan
- [x] npm run typecheck
- [x] npm test


src/inputs/codex-aborted-turn/codex-aborted-turn-builder.ts
```diff
@@ -0,0 +1,405 @@
...
+    span_id: spanId,
+    parent_span_id: stepSpanId,
+    'gen_ai.step.id': stepId,
+    'gen_ai.tool.name': tool.name,
+    'gen_ai.tool.call.id': tool.callId,
+    'tool.result.status': completed ? 'success' : 'cancelled',
+  };
+  if (completed && result.output !== undefined) resultEntry['gen_ai.tool.call.result'] = result.output;
+  const duration = completed ? result.timestampMs - tool.timestampMs : undefined;
+  if (duration !== undefined && duration > 0) resultEntry['gen_ai.tool.call.duration'] = duration;
```
ralf0131: **[Info]** `duration > 0` means a zero-duration tool call (call and result at the same timestamp) gets no `gen_ai.tool.call.duration` field. If that is intentional (to distinguish missing vs instantaneous), a brief comment would help. Otherwise `duration >= 0` would include the zero case.

src/inputs/codex-aborted-turn/codex-aborted-turn-builder.ts
```diff
@@ -0,0 +1,405 @@
...
+  records.push(buildEntry(resultEntry));
+  return records;
+}
+
+function requestTimestamp(wave: ToolWave): number {
+  const resultTimes = [...wave.results.values()].map(result => result.timestampMs);
+  return resultTimes.length > 0
+    ? Math.max(...resultTimes)
+    : Math.max(...wave.calls.map(tool => tool.timestampMs));
```
ralf0131, line 340: **[Info]** `Math.max(...wave.calls.map(...))` returns `-Infinity` if `wave.calls` is empty. By construction waves always contain at least one call, so this cannot happen today — but a defensive fallback (e.g. `wave.calls[0]?.timestampMs ?? 0`) would make the invariant explicit and avoid a silent `-Infinity` timestamp if the construction logic ever changes.

ralf0131 (approved):
﻿### Summary
Adds `CodexAbortedTurnInput` to recover interrupted/aborted Codex turns from session transcript files (`rollout-*.jsonl`), reconstructing full ReAct-step traces that the normal Stop hook misses. Well-structured with incremental byte-offset file reading, checkpoint-based state with inode rotation detection, bounded in-memory de-dup lists, and comprehensive test coverage (unit + integration + fixtures).

Key changes reviewed:
- **New `codex-aborted-turn` module** (input/builder/extractor/types) — parses transcript records into timeline events, groups tool calls into "waves" modelling the ReAct loop, and attributes usage/messages to the correct step.
- **Hook coordination** — `codex-hook-processor.mjs` and `transcript-parser.mjs` now track `abortedTurnIds` and skip them in `resolveTurns`, preventing duplicate traces when both the hook and the recovery input see the same turn.
- **OTLP flusher** — `cancelled` is now treated as a terminal finish reason alongside `stop`/`end_turn`, enabling flush of aborted-turn traces.
- **BaseInput fix** — `stop()` now awaits the in-flight cycle promise before calling `onStop()`, preventing potential data loss on shutdown. The `runCycle` → `runCycleOnce` refactor also deduplicates concurrent cycles.
- **Schema** — `tool.result.status` promoted from a legacy alias to a first-class field on `AgentActivityEntry`.

CLA: signed ✅. No blocking issues found — the two inline notes are minor/defensive.

---

linrunqi08 (commented):
## Code Review — PR #57: fix(codex): improve interrupted turn trace recovery

Reviewed the full diff across 27 files. Found 8 issues — 4 correctness, 2 efficiency, 2 reuse. Inline comments below with details.

**Overall**: The architecture is clean — transcript-backed recovery with checkpoint persistence, dedup ledger, and hook-gap diagnostics. The BaseInput cycle-dedup and OTLP terminal-reason extension are well-motivated. The issues below are mostly about edge-case robustness and a global serialization side-effect.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
```diff
@@ -0,0 +1,427 @@
...
+      }
+      cursor = newline + 1;
+    }
+    return { items, nextOffset: startOffset + lastNewline + 1 };
+  } finally {
+    await handle.close();
+  }
+}
+
+async function readJsonLineAt(filePath: string, offset: number): Promise<Record<string, unknown> | null> {
```
linrunqi08, line 426: **Bug (efficiency/OOM):** `readJsonLineAt` reads from `offset` to `stat.size`, allocating a `Buffer` for the entire range. When `latestSessionMetaOffset` is near the start of the file (session_meta is typically the first record), this allocates a buffer equal to nearly the entire transcript size.

Codex transcripts can reach tens of MB on long sessions. Every aborted-turn recovery triggers this allocation.

**Suggestion:** Read only the first complete line from `offset` — allocate a small fixed buffer (e.g. 64 KB), scan for the first newline, and parse that single record. If the line is longer than the buffer, grow once.

src/inputs/codex-aborted-turn/codex-aborted-turn-extractor.ts
```diff
@@ -0,0 +1,320 @@
...
+        ? toJsonValue({
+        ...(payload.status !== undefined ? { status: payload.status } : {}),
+        ...(payload.execution !== undefined ? { execution: payload.execution } : {}),
+        ...(payload.tools !== undefined ? { tools: parseMaybeJson(payload.tools) } : {}),
+      })
+        : toJsonValue(parseMaybeJson(payload.output)),
+    });
+  }
+
+  if (!abortedAtMs) return null;
```
linrunqi08: **Bug (falsy-zero):** `timestampMs()` (line 234 of this file) returns `0` when the timestamp field is missing or unparseable. The guard `if (!abortedAtMs) return null` treats `0` as "not set" because `!0 === true`.

If a `turn_aborted` event has a malformed or missing timestamp, the entire recovered turn is silently discarded — even though the input-side `timestampMs` (in `codex-aborted-turn-input.ts:423`) would have fallen back to `Date.now()` and correctly detected the abort.

Combined with the finding below (no log on failed recovery), this creates a silent data-loss path.

**Suggestion:** Use a sentinel like `-1` or `NaN` for "no timestamp", or change the guard to `if (abortedAtMs === 0 && startedAtMs === 0) return null` to distinguish genuine zero from parse failure.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
```diff
@@ -0,0 +1,427 @@
...
+          checkpoint.pendingCompletedTurns = checkpoint.pendingCompletedTurns
+            .slice(-MAX_PENDING_COMPLETED_TURNS);
+          checkpoint.activeTurn = null;
+        }
+        continue;
+      }
+      if (payload.type !== 'turn_aborted') continue;
+      const turnId = stringValue(payload.turn_id);
+      if (!turnId || checkpoint.activeTurn?.turnId !== turnId) continue;
+      if (!checkpoint.emittedAbortedTurnIds.includes(turnId)) {
```
linrunqi08, line 179: **Bug (silent data loss):** When `recoverTurn()` returns an empty array (e.g. because `extractAbortedTurn` returned `null` due to the falsy-zero issue above, or any other parse failure), the `turnId` is NOT added to `emittedAbortedTurnIds`, AND `checkpoint.activeTurn` is set to `null` (line 185). After the loop, `checkpoint.scanOffset` advances past the `turn_aborted` event (line 188).

On the next poll, the event is behind `scanOffset` and is never re-scanned. The aborted turn is permanently lost with no error log, no diagnostic, and no retry.

**Suggestion:** At minimum, log a warning when recovery fails. Consider also adding the turnId to `emittedAbortedTurnIds` (to prevent a spurious re-read if the file is somehow re-scanned) and writing a diagnostic entry similar to the hook-gap diagnostic.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
```diff
@@ -0,0 +1,427 @@
...
+    for (const line of lines.items) {
+      const payload = asRecord(line.record.payload);
+      if (!payload) continue;
+      if (line.record.type === 'session_meta') {
+        checkpoint.latestSessionMetaOffset = line.startOffset;
+        checkpoint.latestSessionId = extractCodexTranscriptMeta(line.record)?.sessionId ?? checkpoint.latestSessionId;
+        continue;
+      }
+
+      if (line.record.type === 'event_msg' && payload.type === 'task_started') {
```
linrunqi08, line 137: **Bug (asymmetric guard):** The `task_started` handler unconditionally overwrites `checkpoint.activeTurn` (no check on existing `turnId`), while the `turn_context` handler at line 148 guards with `!checkpoint.activeTurn || checkpoint.activeTurn.turnId !== turnId`.

If Codex emits `turn_context` before `task_started` for the same turn (the hook test fixture at `hook-processor.test.mjs:1729-1730` shows exactly this ordering), `task_started` overwrites `startOffset` to point past the `turn_context` line. When `recoverTurn` reads from `startOffset`, the `turn_context` record is outside the read range, so the extractor misses `model`, `cwd`, and `developer_instructions` — the recovered trace has `model='unknown'` and no working directory.

**Suggestion:** Add the same guard to `task_started`: `if (turnId && (!checkpoint.activeTurn || checkpoint.activeTurn.turnId !== turnId))`.

src/normalization/entry-builder.ts
```diff
@@ -223,7 +223,6 @@ const LEGACY_ALIAS_FIELDS = new Set([
   'tool.exec.id',
   'tool.arguments',
   'tool.result.payload',
-  'tool.result.status',
   'tool.result.duration',
```
linrunqi08, line 226: **Cross-file concern:** Removing `tool.result.status` from `LEGACY_ALIAS_FIELDS` is a global serialization change that affects all existing inputs — not just the new codex-aborted-turn input.

Previously, `removeLegacyAliases()` stripped this field, and `serialiseLogEntry()` skipped it. Now it survives into JSONL/SLS/HTTP output for **all** agent types. Notably:
- `cursor/react-assembler.mjs:601` emits `tool.result.status: 'completed'` — a non-standard value (not `success`/`failure`/`cancelled`/`error`) that was previously invisible and will now leak into downstream output.
- Other inputs use inconsistent vocabulary (`success`, `failure`, `failed`, `error`), all of which will now appear as-is since `applyLegacyToolStatus` normalizes for `error.type` derivation but does not normalize the raw `tool.result.status` value.

This is likely intentional (per the PR summary) but the side effects on cursor and other agents may not be. Consider normalizing the vocabulary or documenting the canonical values.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
```diff
@@ -0,0 +1,427 @@
...
+  }
+}
+
+async function readJsonLineAt(filePath: string, offset: number): Promise<Record<string, unknown> | null> {
+  const stat = await fs.stat(filePath);
+  const lines = await readJsonLines(filePath, offset, stat.size);
+  return lines.items[0]?.record ?? null;
+}
+
+function asRecord(value: unknown): Record<string, unknown> | null {
```
linrunqi08: **Reuse:** `asRecord()`, `stringValue()`, and `timestampMs()` are defined identically in both `codex-aborted-turn-input.ts` (lines 400, 406, 423) and `codex-aborted-turn-extractor.ts` (lines 224, 230, 234) — copy-pasted within the same module.

Notably, the two `timestampMs` variants differ in fallback behavior: the input version falls back to `Date.now()`, the extractor version falls back to `0`. This divergence is the root cause of the falsy-zero bug above. A shared helper with an explicit fallback parameter would prevent this class of inconsistency.

**Suggestion:** Extract to a shared `codex-aborted-turn-utils.ts` within the module.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
```diff
@@ -0,0 +1,427 @@
...
+      emittedAbortedTurnIds: [],
+      pendingCompletedTurns: [],
+      emittedHookGapTurnIds: [],
+    });
+  }
+
+  private async discoverSessionFiles(): Promise<string[]> {
+    const files: string[] = [];
+    await collectRolloutFiles(this.sessionDir, files);
+    return files.sort();
```
linrunqi08, line 246: **Efficiency:** `discoverSessionFiles()` calls `collectRolloutFiles()` which does a full recursive `fs.readdir` walk of `~/.codex/sessions/` on every 30s poll cycle. On a developer machine with months of accumulated sessions (hundreds of subdirectories), this generates O(n) syscalls per poll even when no files have changed.

Consider caching the file list between cycles and only re-walking directories whose mtime has changed, or using a `chokidar`/`fs.watch`-based approach since the input already declares `watchPaths`.

src/inputs/codex-aborted-turn/codex-aborted-turn-builder.ts
```diff
@@ -0,0 +1,405 @@
...
+  return buildAgentActivityEntry({
+    ...rest,
+    timestamp,
+    time_unix_nano: timestampToUnixNanos(timestamp),
+  }) as AgentActivityEntry;
+}
+
+function hashId(parts: string[], length: number): string {
+  return crypto.createHash('sha256').update(parts.join('\0')).digest('hex').slice(0, length);
+}
```
linrunqi08, line 405: **Reuse:** `hashId()` is the 4th independent copy of the same SHA-256 deterministic ID pattern in the codebase (also in `qoder-work-sqlite-input.ts`, `qoder-work-log-input.ts`, `wukong-input.ts`). Consider extracting to a shared utility.

rangemer333-cell: Agreed this is duplicated, but the existing helpers have different input and truncation semantics. I’m deferring a cross-agent utility extraction to a dedicated refactor to avoid broadening this recovery fix.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
rangemer333-cell: Agreed this is an O(n) polling cost. I confirmed watchPaths only controls agent discovery, not input collection, so replacing this with fs.watch is not a drop-in change. I’m deferring it to a separate performance task with scan metrics and a cache/watch design.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
rangemer333-cell: Addressed in 22914c4. asRecord, stringValue, and timestamp parsing now live in codex-aborted-turn-utils.ts. Timestamp parsing returns undefined for invalid input so callers must choose their fallback explicitly.

src/normalization/entry-builder.ts
rangemer333-cell: Addressed in 22914c4. Since this field is now globally serialized, normalization maps completed to success, failed/error to failure, and canceled to cancelled; unknown values become unknown.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
rangemer333-cell: Addressed in 22914c4. task_started now uses the same same-turn guard as turn_context, preserving the earlier offset and therefore model, cwd, and developer instructions.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
rangemer333-cell: Addressed in 22914c4. When recovery produces no entries, the collector now writes a structured codex_aborted_turn_recovery_failed diagnostic and logs a warning before advancing the offset.

src/inputs/codex-aborted-turn/codex-aborted-turn-extractor.ts
rangemer333-cell: Addressed in 22914c4. Invalid timestamps no longer use 0 as a sentinel. The extractor treats them as absent, preventing an invalid Unix-epoch trace; the input emits the recovery-failure diagnostic instead.

src/inputs/codex-aborted-turn/codex-aborted-turn-input.ts
rangemer333-cell: readJsonLineAt now reads a bounded 64 KB chunk and grows only when the first line exceeds that size, instead of allocating from the session-meta offset through EOF.

src/inputs/codex-aborted-turn/codex-aborted-turn-builder.ts
rangemer333-cell: 当前 buildToolWaves() 的构造保证 wave 必有 tool call，requestTimestamp() 也只对已有 wave 调用。加 ?? 0 会把未来不变量被破坏的问题伪装成合法时间戳；若想加强防御，宁可抛出明确错误，但这次没必要

src/inputs/codex-aborted-turn/codex-aborted-turn-builder.ts
rangemer333-cell: 改为 duration >= 0，并加了一个同时间戳 call/result 的测试。

ralf0131 (approved):
### Summary
Re-review after commit `22914c463e` ("harden interrupted turn recovery"). The hardening commit improves robustness with no regressions:

- **`entry-builder.ts`**: `tool.result.status` is now a first-class normalized field (`success`/`failure`/`cancelled`/`unknown`) instead of a legacy alias. This correctly supports the `cancelled` status emitted by aborted turns, and handles both American/British spellings (`cancelled`/`canceled`).
- **Extractor**: simplified and tightened (-24 lines), eliminating redundant parsing paths.
- **Input cycle**: hardened (+59 -22) with improved checkpoint management.
- **New `codex-aborted-turn-utils.ts`**: shared helper extraction.
- Test coverage expanded (+90 lines) covering edge cases.

No new issues found. CLA signed.

---

linrunqi08 (approved):

PR #60, opened 2026-06-23
feat(qoder-work): intercept token usage via worker-runtime wrapper (0.6.2+)

rangemer333-cell:
## Background

QoderWork 0.6.2+ no longer spawns a qodercli Bun subprocess and no longer writes token data to local files (transcript / SDK-log `usage` is all 0). The existing injection approaches all fail:

| Approach | Result | Reason |
|---|---|---|
| `BUN_OPTIONS=--preload` | ❌ | no qodercli Bun child to inject into |
| `NODE_OPTIONS=--require` | ❌ | Electron strips NODE_OPTIONS |
| `launchctl setenv BUN_OPTIONS` | ❌ | no Bun child |

## Approach

The SDK resolves its worker runtime via `QODER_WORKER_RUNTIME_PATH`, so we point it at a **wrapper** that:
1. Installs the same `JSON.parse` / `JSON.stringify` hooks as `assets/hooks/qodercli-token-intercept.mjs`
2. `await import()`s the real `qoder-worker-runtime.obf.mjs`

Token + system-prompt records are written to the **shared** `~/.loongsuite-pilot/logs/qodercli-intercept.jsonl`, so pilot's `intercept-token-reader.ts` (from #43) needs no changes. Verified on macOS (design doc 实测 1-3 ✅).

## Changes

- **`assets/hooks/qoderwork-runtime-wrapper.mjs`** — ESM wrapper: install hooks → import real runtime → worker runs with hook active. Record format byte-for-byte consistent with the qodercli CLI intercept.
- **`scripts/postinstall.js`** — wrapper auto-deployed by existing recursive copy; on macOS runs `launchctl setenv QODER_WORKER_RUNTIME_PATH` (non-fatal).
- **`deploy/installer-opensource.sh`** — `inject_qoderwork_runtime_wrapper` (install) + `remove_qoderwork_runtime_wrapper` (uninstall, `launchctl unsetenv`).
- **`src/inputs/qoder-work-trace/qoder-work-trace-input.ts`** — last-resort token fallback: when segment usage **and** SDK-log usage are both 0 (the 0.6.2 regression), match tokens from `readInterceptData()` by timestamp (closest `ts` to the response time, FIFO, consumed once). Mirrors `segment-token-reader.ts` `zeroSegments` fallback.
- **Tests** — wrapper format-contract test (keeps both intercept files in sync) + intercept-fallback unit tests.

## Known limitations

- macOS-only deployment (`launchctl setenv`); QoderWork is a macOS Electron app. `launchctl setenv` does not survive reboot — persistence via LaunchAgent plist is deferred follow-up (design item 4).
- The intercept JSONL is shared with the qodercli CLI, so a CLI token captured closest-in-time to a QoderWork response could theoretically mismatch — acceptable since both agents are rarely used simultaneously, and the fallback only fires when local usage is 0.
- Intercept data has no `cache_creation` field (same limitation as the CLI path).

## Test plan

- [x] `npm run typecheck` (tsc --noEmit) — pass
- [x] `npm run build` — pass
- [x] `npx vitest run` — 1280 pass / 2 skip (no regressions)
- [x] New unit tests: wrapper contract + intercept fallback (6 tests)
- [x] `bash -n deploy/installer-opensource.sh` — syntax OK
- [x] `node scripts/postinstall.js` — wrapper deployed to hooks/, QoderWork block no-ops on Linux
- [ ] **Real E2E on macOS**: install pilot, restart QoderWork 0.6.2+, trigger a multi-turn ReAct conversation, run `validate-trace` on the resulting trace (0 ERROR). *Cannot run from this Linux environment — needs a macOS machine.*

Design doc: `qoderwork-token-intercept-design.md` (attached to the originating issue).

### DIFF
--- a/assets/hooks/qoderwork-runtime-wrapper.mjs
+++ b/assets/hooks/qoderwork-runtime-wrapper.mjs
@@ -1,0 +1,120 @@
+// QoderWork worker runtime wrapper — intercepts token data via JSON.parse hook.
+//
+// Loaded via: QODER_WORKER_RUNTIME_PATH=<this-file>
+//
+// QoderWork 0.6.2+ no longer spawns a qodercli Bun subprocess; its LLM calls run
+// inside a Node.js worker_thread that loads qoder-worker-runtime.obf.mjs. We can't
+// use BUN_OPTIONS --preload (no Bun child) or NODE_OPTIONS --require (Electron
+// strips it). Instead the SDK honors the QODER_WORKER_RUNTIME_PATH env var to
+// locate the worker runtime — so we point it at this wrapper, which installs the
+// same JSON.parse / JSON.stringify hooks as assets/hooks/qodercli-token-intercept.mjs
+// and then `await import()`s the real runtime.
+//
+// Flow: install hook → import real runtime → worker runs normally with hook active.
+// Token + system-prompt records are written to the SAME intercept JSONL as the
+// qodercli CLI intercept, so pilot's intercept-token-reader.ts needs no changes.
+//
+// Deployment: postinstall.js copies this file to ~/.loongsuite-pilot/hooks/ and runs
+// `launchctl setenv QODER_WORKER_RUNTIME_PATH <this-file>` (macOS). QoderWork must be
+// restarted after install/uninstall for the env change to take effect.
+
+import { createRequire } from 'node:module';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const require = createRequire(import.meta.url);
+
+const INTERCEPT_DIR = path.join(process.env.HOME || '/tmp', '.loongsuite-pilot', 'logs');
+const INTERCEPT_FILE = path.join(INTERCEPT_DIR, 'qodercli-intercept.jsonl');
+const MIN_SYSTEM_PROMPT_LENGTH = 100;
+
+try { fs.mkdirSync(INTERCEPT_DIR, { recursive: true }); } catch {}
+
+const DEBUG = !!process.env.LOONGSUITE_PILOT_INTERCEPT_DEBUG;
+function debug(msg) { if (DEBUG) try { fs.appendFileSync(path.join(INTERCEPT_DIR, 'qoderwork-wrapper.log'), `[${new Date().toISOString()}] ${msg}\n`); } catch {} }
+
+const origParse = JSON.parse;
+const origStringify = JSON.stringify;
+let lastId = null;
+let systemPromptCaptured = false;
+
+// ── JSON.parse hook: capture token usage from the SSE response ──────────────
+// Must stay byte-for-byte consistent with qodercli-token-intercept.mjs so both
+// producers emit the identical record shape consumed by intercept-token-reader.ts.
+JSON.parse = function (text, reviver) {
+  const result = origParse.call(JSON, text, reviver);
+  try {
+    if (result && typeof result === "object"
+        && result.usage && result.choices !== undefined
+        && result.id !== lastId) {
+      lastId = result.id;
+      const u = result.usage;
+      const rec = {
+        type: "token",
+        ts: Date.now(),
+        id: result.id,
+        model: result.model || "",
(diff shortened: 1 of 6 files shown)


assets/hooks/qoderwork-runtime-wrapper.mjs
```diff
@@ -0,0 +1,120 @@
...
+//
+// Flow: install hook → import real runtime → worker runs normally with hook active.
+// Token + system-prompt records are written to the SAME intercept JSONL as the
+// qodercli CLI intercept, so pilot's intercept-token-reader.ts needs no changes.
+//
+// Deployment: postinstall.js copies this file to ~/.loongsuite-pilot/hooks/ and runs
+// `launchctl setenv QODER_WORKER_RUNTIME_PATH <this-file>` (macOS). QoderWork must be
+// restarted after install/uninstall for the env change to take effect.
+
+import { createRequire } from 'node:module';
```
ralf0131, line 21: **[Info]** `createRequire` (line 21) and `fileURLToPath` (line 24) are imported, and `const require = createRequire(import.meta.url)` (line 26) is declared, but none of them are used anywhere in this wrapper — the real runtime is loaded via `await import(candidate)`. These appear to be leftover from an earlier draft. Safe to remove for cleanliness.

src/inputs/qoder-work-trace/qoder-work-trace-input.ts
```diff
@@ -550,6 +565,44 @@ export class QoderWorkTraceInput extends BaseInput {
...
 
+  // Last-resort token source: match an intercept token record to this LLM call
+  // by timestamp. The wrapper writes the record's `ts` at JSON.parse time ≈
+  // response completion, which aligns with the segment-enriched response time.
+  // Each token is consumed once (FIFO splice) to avoid double-counting.
+  // Known limitation: the intercept JSONL is shared with the qodercli CLI, so a
+  // CLI token captured closest-in-time to a QoderWork response could mismatch —
+  // acceptable since both agents are rarely used simultaneously.
+  private applyInterceptTokenUsage(
+    request: AgentActivityEntry,
```
ralf0131, line 576: **[Warning]** `const ref = responseMs ?? requestMs!;` uses a non-null assertion on `requestMs`. The early-return guard above makes this safe today, but the `!` is fragile — if the guard logic ever changes the assertion could silently mask a TypeError. Consider an explicit check: `const ref = responseMs ?? requestMs; if (ref === undefined) return;`. Also, `ref` is loop-invariant (neither `responseMs` nor `requestMs` change inside the `for` loop), so it could be hoisted above the loop for clarity.

ralf0131 (commented):
﻿### Summary
This PR adds token-usage interception for QoderWork 0.6.2+, which moved its LLM calls from a qodercli Bun subprocess into a Node.js `worker_thread`. The approach is sound: a `QODER_WORKER_RUNTIME_PATH`-loaded wrapper installs `JSON.parse`/`JSON.stringify` hooks (consistent with the existing `qodercli-token-intercept.mjs`), then `await import()`s the real runtime. The fallback in `qoder-work-trace-input.ts` is additive — it only kicks in when the SDK reports zero usage, preserving existing behaviour. Test coverage is excellent (6 cases for the intercept fallback plus contract tests verifying the wrapper's record shape).

### CI Status
All checks pass (Analyze, CodeQL, build-and-test 18/20/22, license/cla). CLA is signed.

### Findings
- **[Info]** `qoderwork-runtime-wrapper.mjs:21` — dead imports (`createRequire`, `fileURLToPath`, unused `require` binding)
- **[Warning]** `qoder-work-trace-input.ts:576` — non-null assertion `requestMs!` is fragile; `ref` is loop-invariant

### Suggestions
The two findings are minor cleanups. The core design — wrapper-via-env-var + timestamp-matched FIFO fallback — is well-reasoned and correctly handles the shared-JSONL limitation. The non-null assertion is the more worthwhile one to address since it touches a safety guard.

---

rangemer333-cell:
按需求关闭此 PR，后续将以正确身份和规范流程重开。

PR #61, opened 2026-06-24
feat(sls): add user-agent header to all SLS transport paths

linrunqi08:
## Summary

- Add `user-agent` header (`loongsuite-pilot/{version} ({os}; {release}; {arch}) ip/{ip}`) to all SLS data transport paths: AK mode (`@alicloud/log` SDK), inline webtracking `fetch`, and `postWebtracking` transport layer
- Extract shared `buildUserAgent(dataDir)` and `LOCAL_IP` into `src/utils/network-utils.ts` to eliminate duplicated IP/version resolution across `sls-flusher.ts`, `file-sls-sender.ts`, and `sls-transport.ts`
- Use `readInstalledVersion(dataDir)` for robust version resolution in bundled deployments (replaces unreliable `import.meta.url`-based `package.json` lookup)
- Thread `dataDir` through Orchestrator → FileCollectionManager → FilePipeline → FileSlsSender call chain

## Changed Files

- `src/utils/network-utils.ts` — add `buildUserAgent()` and `LOCAL_IP` exports
- `src/flushers/sls-flusher.ts` — use shared `buildUserAgent()`, set `userAgent` in AK client constructor and webtracking fetch headers
- `src/flushers/sls-transport.ts` — add optional `userAgent` field to `PostWebtrackingOptions`, apply to fetch headers
- `src/file-collection/file-sls-sender.ts` — use shared `LOCAL_IP` and `buildUserAgent()`
- `src/file-collection/types.ts` — add `dataDir` to `FileCollectionManagerOptions` and `FilePipelineOptions`
- `src/file-collection/file-collection-manager.ts` — thread `dataDir`
- `src/file-collection/file-pipeline.ts` — thread `dataDir` to `FileSlsSender`
- `src/core/orchestrator.ts` — pass `dataDir` to `FileCollectionManager` and `OtlpTraceFlusher`
- `src/types/index.ts` — add `dataDir` to `OtlpTraceFlusherConfig`
- `src/flushers/otlp-trace-flusher.ts` — use `readInstalledVersion(dataDir)` instead of `getPilotVersion()`

## Test Plan

- [x] TypeScript compilation passes (`npx tsc --noEmit`)
- [ ] Verify `user-agent` header appears in SLS requests via debug logging
- [ ] Verify version resolves correctly in bundled deployment (`~/.loongsuite-pilot/`)

### DIFF
--- a/src/flushers/otlp-trace-flusher.ts
+++ b/src/flushers/otlp-trace-flusher.ts
@@ -20,11 +20,10 @@ import { BaseFlusher } from './base-flusher.js';
 import { normalizeAgentType } from '../utils/agent-type-normalize.js';
 import { resolveAgentSystem } from '../normalization/agent-system-map.js';
 import { createLogger } from '../utils/logger.js';
-import { appendLine, ensureDir, getTodayDateString } from '../utils/fs-utils.js';
+import { appendLine, ensureDir, getTodayDateString, readInstalledVersion } from '../utils/fs-utils.js';
 import { randomUUID } from 'node:crypto';
 import os from 'node:os';
 import path from 'node:path';
-import { readFileSync } from 'node:fs';
 
 const logger = createLogger('otlp-trace-flusher');
 
@@ -68,17 +67,6 @@ function resolveEndpointUrl(raw: string): string {
   return url;
 }
 
-function getPilotVersion(): string {
-  try {
-    const pkg = JSON.parse(
-      readFileSync(new URL('../../package.json', import.meta.url), 'utf-8'),
-    );
-    return pkg.version ?? 'unknown';
-  } catch {
-    return 'unknown';
-  }
-}
-
 const DEFAULT_MAX_EXPORT_BATCH_BYTES = 10 * 1024 * 1024; // 10 MB
 
 function estimateSpanSize(span: ReadableSpan): number {
@@ -105,7 +93,7 @@ export class OtlpTraceFlusher extends BaseFlusher {
   private readonly agentConvertStates = new Map<string, AgentConvertState>();
   private readonly agentExportStates = new Map<string, AgentExportState>();
   private readonly instanceId = randomUUID();
-  private readonly pilotVersion = getPilotVersion();
+  private readonly pilotVersion: string;
   private readonly resolvedEndpointUrl: string;
   private readonly debugDir: string;
   private readonly failedDir: string;
@@ -130,9 +118,11 @@ export class OtlpTraceFlusher extends BaseFlusher {
       throw new Error('[otlp-trace-flusher] config.serviceName is required when enabled');
     }
     this.cfg = cfg;
+    const dataDir = cfg.dataDir ?? os.homedir() + '/.loongsuite-pilot';
+    this.pilotVersion = readInstalledVersion(dataDir);
     this.resolvedEndpointUrl = resolveEndpointUrl(cfg.endpoint);
-    this.debugDir = path.join(os.homedir(), '.loongsuite-pilot', 'logs', 'otlp-debug');
-    this.failedDir = path.join(os.homedir(), '.loongsuite-pilot', 'logs', 'otlp-failed');
+    this.debugDir = path.join(dataDir, 'logs', 'otlp-debug');
+    this.failedDir = path.join(dataDir, 'logs', 'otlp-failed');
 
     if (cfg.captureMessageContent !== false) {
       process.env.OTEL_SEMCONV_STABILITY_OPT_IN ??= 'gen_ai_latest_experimental';
--- a/src/flushers/sls-flusher.ts
+++ b/src/flushers/sls-flusher.ts
@@ -11,6 +11,7 @@ import { createLogger } from '../utils/logger.js';
 import { appendLine, ensureDir } from '../utils/fs-utils.js';
 import { formatTime } from '../utils/time-utils.js';
 import { normalizeAgentType } from '../utils/agent-type-normalize.js';
+import { LOCAL_IP, buildUserAgent } from '../utils/network-utils.js';
 import * as path from 'node:path';
 import {
   HttpError,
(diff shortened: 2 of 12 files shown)


linrunqi08 (commented):
## Code Review — PR #61

Reviewed the full diff across all 12 changed files. No correctness bugs found — the `user-agent` plumbing, `LOCAL_IP` consolidation, and `dataDir` threading are functionally correct. The `@alicloud/log` SDK does accept `userAgent` in its constructor (verified in `node_modules/@alicloud/log/lib/client.js:75`), so the AK mode path is covered.

Below are cleanup/efficiency/altitude findings, ranked by impact:

---

### 1. `dataDir` pass-through plumbing could be avoided (Altitude)

`dataDir` is threaded through 4 constructor layers: `Orchestrator → FileCollectionManager → FilePipeline → FileSlsSender`. The two middle layers (`FileCollectionManager`, `FilePipeline`) store `dataDir` as a field but never use it — they only forward it to the next layer. Consider computing the `userAgent` string once at the `Orchestrator` level and passing that string directly, which would eliminate 2 interface changes and 2 forwarding fields.

### 2. `readInstalledVersion` repeated sync I/O (Efficiency)

`readInstalledVersion(dataDir)` does 2+ synchronous `readFileSync` calls per invocation. It's invoked:
- Once per `SlsFlusher` instance (via `buildUserAgent`)
- Once per `FileSlsSender` instance (via `buildUserAgent`) — one per file-collection config
- Once in `OtlpTraceFlusher` (called directly)

The version never changes during process lifetime. `LOCAL_IP` is already cached as a module-level constant, but the version is re-read from disk every time `buildUserAgent` is called. Consider caching the version per `dataDir` inside `readInstalledVersion`, or computing `buildUserAgent` once and passing the result.

### 3. `OtlpTraceFlusher` tests don't mock `readInstalledVersion` (Test hygiene)

The `OtlpTraceFlusher` test suites (`conversion.test.ts`, `export.test.ts`, `config.test.ts`, etc.) construct `OtlpTraceFlusher` without `dataDir` in the config and without mocking `readInstalledVersion`. The constructor now calls `readInstalledVersion(cfg.dataDir ?? os.homedir() + '/.loongsuite-pilot')`, hitting the real filesystem. This works (errors are swallowed → returns `'unknown'`), but makes tests non-hermetic. The `sls-flusher` tests already mock it correctly (`readInstalledVersion: () => '0.0.0-test'`) — consider doing the same for `OtlpTraceFlusher` tests.

### 4. Redundant `readInstalledVersion` call in `OtlpTraceFlusher` (Reuse)

`OtlpTraceFlusher` calls `readInstalledVersion(dataDir)` directly for `this.pilotVersion`, while `buildUserAgent(dataDir)` also calls `readInstalledVersion(dataDir)` internally. If `OtlpTraceFlusher` later adds a user-agent header (or if the version is needed elsewhere), the same sync I/O runs twice with the same `dataDir`. Minor today, but consolidating version resolution would help.

### 5. User-agent embeds internal IP address (Informational)

The user-agent string format `loongsuite-pilot/{version} ({os}; {release}; {arch}) ip/{LOCAL_IP}` embeds the machine's internal IPv4 address in every outbound HTTP header. The IP was already sent as `source: LOCAL_IP` in SLS log payloads, so this is likely intentional for correlation — just flagging in case this was unintended for the HTTP header path.

---

Overall: clean refactor with correct deduplication. The findings above are all improvement suggestions, not blockers.

src/flushers/sls-flusher.ts
```diff
@@ -104,6 +95,7 @@ export class SlsFlusher extends BaseFlusher {
         accessKeyId: endpoint.accessKeyId ?? '',
         accessKeySecret: endpoint.accessKeySecret ?? '',
         endpoint: endpoint.endpoint,
+        userAgent: this.userAgent,
```
ralf0131, line 98: **[Warning]** The entire AK client options object is cast to `any` (`} as any)`). This bypasses TypeScript type checking for all properties, not just the new `userAgent` field. Consider narrowing the cast (e.g., type the SDK options with an intersection) or adding a comment explaining that the `@alicloud/log` SDK type definition doesn't yet expose `userAgent`, so the cast is a temporary workaround.

src/utils/network-utils.ts
```diff
@@ -11,3 +12,10 @@ export function resolveLocalIp(): string {
   }
   return '127.0.0.1';
 }
+
```
ralf0131, line 15: **[Info]** `buildUserAgent()` has no dedicated unit test. The existing `sls-flusher` tests mock `readInstalledVersion` but don't verify the user-agent string format. Consider adding a test in `tests/unit/utils/network-utils.test.ts` to cover the output format and the version-resolution fallback.

src/flushers/otlp-trace-flusher.ts
```diff
@@ -130,9 +118,11 @@ export class OtlpTraceFlusher extends BaseFlusher {
       throw new Error('[otlp-trace-flusher] config.serviceName is required when enabled');
     }
     this.cfg = cfg;
+    const dataDir = cfg.dataDir ?? os.homedir() + '/.loongsuite-pilot';
+    this.pilotVersion = readInstalledVersion(dataDir);
```
ralf0131, line 122: **[Info]** `debugDir`/`failedDir` now respect `cfg.dataDir` instead of being hardcoded to `~/.loongsuite-pilot`. Good improvement for flexibility — just noting that any external log scrapers expecting OTLP debug logs at the old fixed path would need updating when `dataDir` differs from the default.

ralf0131 (commented):
### Summary
PR adds a user-agent header to all SLS transport paths (AK mode, inline webtracking fetch, postWebtracking) and consolidates duplicated IP/version resolution into a shared `network-utils.ts`. Clean DRY refactoring — the `dataDir` threading through the component chain and the switch from unreliable `import.meta.url` to `readInstalledVersion(dataDir)` are good improvements.

Overall verdict: **looks good**. A few minor suggestions below.

---
*Automated review by *

messixukejia (approved):

PR #73, opened 2026-06-25
fix(qoder-work): change user-hook event.name from llm.request to other

rangemer333-cell:
## Summary
- User input was incorrectly emitted as `event.name='llm.request'` (without step.id) in the QoderWork hook-processor, creating orphan LLM spans in the trace pipeline
- Production data confirms 558 orphan `llm.request` events (no matching `llm.response`), all containing only user text with no step.id
- Fix: change the user-hook event from `llm.request` to `other` — step 1's `llm.request` already carries user text via `input.messages_delta`, so the user-hook emission was both redundant and harmful

## Test plan
- [ ] Deploy updated hook to `~/.loongsuite-pilot/hooks/` and restart daemon
- [ ] Verify new history records no longer contain orphan `llm.request` events without step.id
- [ ] Verify `llm.request` count equals `llm.response` count (previously 3975 vs 3417, delta = 558)
- [ ] Verify step 1 still correctly includes user input in `gen_ai.input.messages_delta`

### DIFF
--- a/assets/hooks/qoderwork-hook-processor.mjs
+++ b/assets/hooks/qoderwork-hook-processor.mjs
@@ -178,7 +178,7 @@ function buildTurnEvents(turnRows, turnId, sessionId, userId, providerName, vers
     if (userText) {
       records.push(buildRecord({
         ...turnMetadata,
-        'event.name': 'llm.request',
+        'event.name': 'other',
         'gen_ai.turn.id': turnId,
         'gen_ai.session.id': sessionId,
         'gen_ai.agent.type': 'qoder-work',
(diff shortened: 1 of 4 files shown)


ralf0131 (approved):
LGTM. Well-justified fix — the user-hook was emitting a redundant `llm.request` event (no matching `llm.response`), creating 558 orphan LLM spans in production traces. Changing to `other` event type is correct since step 1's `llm.request` already carries user input via `gen_ai.input.messages_delta`. Both the hook processor (`.mjs`) and trace input (`.ts`) are updated consistently, and tests correctly reflect the removal of duplicate user-text entries.

---

linrunqi08 (approved):

PR #74, opened 2026-06-25
fix(codex): remove stale `version` field from existing Codex hooks.json

fangxiu-wf:
## Summary

Follow-up to #49. The previous fix stopped \*injecting\* the `version` field into Codex hooks.json, but did not clean up files that were already corrupted by older pilot versions.

This PR adds an active cleanup path: when `ensureSettingsFile` encounters a non-Cursor hooks.json that still carries a `version` key, it deletes the field and rewrites the file. This heals previously broken Codex installations automatically on the next deploy cycle.

## Changes

- `src/deployment/hook-strategy.ts`: add `else if (!needsVersion && existing.version !== undefined)` branch to delete stale `version` field
- `tests/unit/deployment/hook-strategy.test.ts`: add test for stale version removal + test for Codex file creation from scratch

## Test plan

- `npx vitest run tests/unit/deployment/hook-strategy.test.ts` — 29 tests passed
- Manual: inject `version: 1` into `~/.codex/hooks.json`, run pilot deploy → field is removed

### DIFF
--- a/src/deployment/hook-strategy.ts
+++ b/src/deployment/hook-strategy.ts
@@ -365,6 +365,11 @@ export class HookStrategy implements DeployStrategy {
     } else if (needsVersion && existing.version === undefined) {
       existing.version = 1;
       await writeJsonFile(settingsPath, existing);
+    } else if (!needsVersion && existing.version !== undefined) {
+      // Clean up stale `version` field previously injected by older pilot versions.
+      // Codex uses #[serde(deny_unknown_fields)] and rejects any key other than `hooks`.
+      delete existing.version;
+      await writeJsonFile(settingsPath, existing);
     }
   }
 }
--- a/src/deployment/hook-strategy.ts
+++ b/src/deployment/hook-strategy.ts
@@ -365,7 +365,7 @@ export class HookStrategy implements DeployStrategy {
     } else if (needsVersion && existing.version === undefined) {
       existing.version = 1;
       await writeJsonFile(settingsPath, existing);
-    } else if (!needsVersion && existing.version !== undefined) {
+    } else if (isHooksJson && settingsPath.includes('.codex') && existing.version !== undefined) {
       // Clean up stale `version` field previously injected by older pilot versions.
       // Codex uses #[serde(deny_unknown_fields)] and rejects any key other than `hooks`.
       delete existing.version;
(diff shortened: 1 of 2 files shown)


ralf0131 (approved):
LGTM. Clean follow-up fix to #49 — proactively removes stale `version` field from Codex hooks.json files corrupted by older pilot versions. The `else if` branch correctly targets only Codex files (`isHooksJson && settingsPath.includes('.codex')`) and only acts when a stale `version` key exists. Test coverage is solid (stale removal + creation from scratch).

One minor note: `settingsPath.includes('.codex')` is a substring match that could theoretically match unintended paths (e.g. `.codex-backup`), but in practice the path is constructed from agent definitions so this is not a real concern.

---

linrunqi08 (approved):
## Code Review Summary

**Verdict: LGTM** ✅

Reviewed the entire `ensureSettingsFile` control flow across all realistic path combinations:

| Scenario | Branch hit | Behavior | Correct? |
|----------|-----------|----------|----------|
| New Codex file (no existing) | `!existing` → creates `{ hooks: {} }` | No version added | ✓ |
| New Cursor file (no existing) | `!existing` → creates `{ version: 1, hooks: {} }` | Version added | ✓ |
| Existing Codex with stale `version` | **new branch** → deletes `version`, rewrites | Cleanup fires | ✓ |
| Existing Codex without `version` | no branch matches → no-op | Correct no-op | ✓ |
| Existing Cursor without `version` | `needsVersion && !version` → adds version | Version added | ✓ |
| Existing Cursor with `version` | no branch matches → no-op | Correct no-op | ✓ |

**Cross-file trace**: `ensureSettingsFile` runs before `installHook` (line 89 vs 108), so the cleanup write doesn't race with hook installation. `applyEnvToSettings` is gated on `hookConfig.env` and won't interfere.

**Edge cases verified**:
- `version: 0` (falsy) → `0 !== undefined` is `true` → correctly cleaned up
- `version: null` → `null !== undefined` is `true` → correctly cleaned up
- `readJsonFile` returns `null` on parse errors → falls through to the `!existing` branch, no stale cleanup attempted

**Tests**: Both new tests cover the important paths (stale version removal + fresh Codex file creation). Coverage looks complete for the new branch.

**One minor observation** (not blocking): `settingsPath.includes('.codex')` is a substring match that would also match paths like `.codex-backup/hooks.json`. A more precise check like `settingsPath.includes('/.codex/')` could avoid false positives — but this is consistent with the existing `.includes('.cursor')` pattern, so not a regression.

Clean, well-scoped fix. 👍

PR #75, opened 2026-06-25
fix(qoder-cn): improve data quality — token coverage, span durations, dedup

Snssn:
## Summary

Production data showed qoder-cn had `llm.request.gen_ai.step.id` at 50%, `llm.response.gen_ai.response.id` / `input_tokens` / `output_tokens` at 0%, `tool.call.duration` at 0, plus duplicate events (one user input producing multiple turns/traces). This PR addresses all of these by aligning qoder-cn's transcript handling with the spec and with the qoder/qoder-work implementations.

### Root causes

- **Hook processor** emitted partial turns from `PostToolUse` retries instead of waiting for `Stop` and reprocessing the full transcript (claude-code does the latter).
- **enrichIdeTurn**'s backwards `llm.request` scan was unbounded across turns, overwriting earlier turns' request timestamps and breaking step-1 timing.
- **sqlite-token-reader** joined tables differently from `qoder-trace`, missing `message_id` and `model_info`, so `gen_ai.response.id` / model fields stayed empty.
- **Cross-cycle orphan turns** (response arriving in a later collect cycle than its anchor user prompt) were emitted as standalone turns.
- **Container spans** (STEP / AGENT / ENTRY) inherited only the first child's time, collapsing to 0ms; `tool.result` lacked an explicit duration field.
- Custom `qoder.match.*` fields polluted output and aren't part of the trace spec.

### Changes

- `assets/hooks/qoder-hook-processor.mjs`: add lockfile-based retry gating. Phase 2.5 skips processing unless a `Stop` event is present in the progress log, then resets `startLine = 0` to reprocess the full transcript (claude-code style).
- `src/inputs/qoder-cn-trace/qoder-cn-trace-input.ts`: introduce `sessionAnchor` (cross-cycle Map) to merge orphan turns into their anchor turn; add `dedupeEventsInTurn`, `expandContainerTimes` (skips `llm.*` / `tool.*` / `other`), `propagateModelToToolEvents`, `computeToolCallDurations`, and `alignUserBoundaryToFirstLlmRequest`.
- `src/inputs/qoder-cn-trace/sqlite-token-reader.ts`: align SQL with `qoder-trace`'s reader (LEFT JOIN `chat_record`, parse `model_info`), expose `sessionId` / `messageId` / `model` on `SqliteTokenData`.
- `src/inputs/qoder-trace/token-enricher.ts`: scope the backwards `llm.request` scan to the same `step.id` (with same-turn fallback); offset step-1 request by +1ms past `userBoundary` to avoid a duplicate empty STEP; remove `qoder.match_confidence` / `qoder.match.strategy` / `qoder.match.warning` custom fields; `markLowConfidence` becomes a no-op.
- `scripts/field-coverage.mjs`: validation script for per-event-type field coverage.
- Tests: 24 new unit tests across hooks, sqlite reader, and trace input.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run tests/unit/inputs/` — 272/272 pass
- [x] New unit tests for retry lock, sqlite reader, and qoder-cn trace input — 39/39 pass
- [x] Local end-to-end: 100% field coverage on `llm.request` / `llm.response` / `tool.call` / `tool.result`
- [x] Local end-to-end: 0 `validate-trace` errors, 7 correctly-structured spans per multi-step turn, no duplicate traces

### DIFF
--- a/assets/hooks/qoder-hook-processor.mjs
+++ b/assets/hooks/qoder-hook-processor.mjs
@@ -31,7 +31,91 @@ import {
   buildQoderHookRecord,
   inferProviderName,
 } from './agent-event-normalizer.mjs';
-import { isQoderIdeaSession } from './shared/qoder-db-utils.mjs';
+
+// --- Retry lockfile (qoder-cn only) -----------------------------------------
+// QoderCN fires Stop hook multiple times per turn AND incomplete transcript
+// causes background retries — without coordination these can stack up and
+// produce duplicate records. We guard at two points:
+//   1. parent process: skip spawn if a live lock exists
+//   2. retry subprocess: refuse to enter processTranscript if a peer holds it
+// The lock file lives at <HOOKS_DIR>/.retry-locks/<sha1>.lock and contains
+// JSON `{ pid, sessionId, startedAt }`.
+
+export const RETRY_LOCK_DIR = path.join(HOOKS_DIR, '.retry-locks');
+export const RETRY_LOCK_MAX_AGE_MS = 60_000;
+
+export function retryLockPath(transcriptPath, dir = RETRY_LOCK_DIR) {
+  const hash = crypto.createHash('sha1').update(transcriptPath).digest('hex');
+  return path.join(dir, `${hash}.lock`);
+}
+
+export function pidAlive(pid) {
+  if (!pid || typeof pid !== 'number') return false;
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (e) {
+    return e && e.code === 'EPERM';
+  }
+}
+
+export function readRetryLock(lockPath) {
+  try {
+    const raw = fs.readFileSync(lockPath, 'utf-8');
+    const parsed = JSON.parse(raw);
+    if (parsed && typeof parsed === 'object') return parsed;
+  } catch { /* fall through */ }
+  return null;
+}
+
+export function isRetryLockStale(lock) {
+  if (!lock) return true;
+  const age = Date.now() - (Number(lock.startedAt) || 0);
+  if (age > RETRY_LOCK_MAX_AGE_MS) return true;
+  return !pidAlive(lock.pid);
+}
+
+export function tryAcquireRetryLock(transcriptPath, sessionId, dir = RETRY_LOCK_DIR) {
+  try {
+    fs.mkdirSync(dir, { recursive: true });
+    const lockPath = retryLockPath(transcriptPath, dir);
+    const payload = JSON.stringify({ pid: process.pid, sessionId, startedAt: Date.now() });
+    try {
+      const handle = fs.openSync(lockPath, 'wx');
+      fs.writeSync(handle, payload);
+      fs.closeSync(handle);
+      return true;
+    } catch (err) {
+      if (err && err.code === 'EEXIST') {
+        const existing = readRetryLock(lockPath);
+        if (isRetryLockStale(existing)) {
+          try { fs.unlinkSync(lockPath); } catch { /* ignore */ }
+          try {
+            const handle = fs.openSync(lockPath, 'wx');
+            fs.writeSync(handle, payload);
+            fs.closeSync(handle);
+            return true;
+          } catch { return false; }
+        }
+        return false;
+      }
+      return false;
+    }
+  } catch {
+    return false;
+  }
+}
+
+export function releaseRetryLock(transcriptPath, dir = RETRY_LOCK_DIR) {
+  try {
+    const lockPath = retryLockPath(transcriptPath, dir);
+    const existing = readRetryLock(lockPath);
+    // Only release if we own the lock (avoid wiping a peer's lock on crash recovery)
+    if (existing && existing.pid === process.pid) {
+      fs.unlinkSync(lockPath);
+    }
+  } catch { /* best-effort */ }
+}
 
 // --- Timestamp helpers -------------------------------------------------------
 
@@ -71,9 +155,32 @@ async function main() {
     if (!transcriptPath || !sessionId) return;
     logDebug(agentId, `Retry: processing ${transcriptPath} for session ${sessionId}`);
     const runtimeConfig = loadHookRuntimeConfig(path.join(HOOKS_DIR, '..'));
-    const range = getLineRange(agentId, transcriptPath, sessionId);
-    if (!range) return;
-    await processTranscript(agentId, logPrefix, transcriptPath, sessionId, range[0], range[1], runtimeConfig, cwd);
+
+    // qoder-cn only: serialize concurrent retries on the same transcript.
+    // We wait the HOOK_RETRY_DELAY first (so all queued Stop hooks have
+    // already advanced the offset via updateLineRecord), then acquire the
+    // lock. The losers see currentCount==lastCount in getLineRange and exit.
+    const retryDelay = parseInt(process.env.HOOK_RETRY_DELAY || '0', 10);
+    if (retryDelay > 0) {
+      await new Promise(r => setTimeout(r, retryDelay));
+    }
+
+    if (agentId === 'qoder-cn') {
+      if (!tryAcquireRetryLock(transcriptPath, sessionId)) {
+        logDebug(agentId, `Retry skipped: lock held by peer for ${transcriptPath}`);
+        return;
+      }
+    }
+    try {
+      const range = getLineRange(agentId, transcriptPath, sessionId);
+      if (!range) return;
+      await processTranscript(
+        agentId, logPrefix, transcriptPath, sessionId,
+        range[0], range[1], runtimeConfig, cwd, { delayApplied: true },
+      );
+    } finally {
+      if (agentId === 'qoder-cn') releaseRetryLock(transcriptPath);
+    }
     return;
   }
 
@@ -110,7 +217,34 @@ async function main() {
   const hasLastPrompt = parsed.some(p => p.type === 'last-prompt');
   if (parsed.length > 0 && !hasLastPrompt) {
     logDebug(agentId, `Transcript incomplete (${parsed.length} lines, no last-prompt marker). Spawning background retry in 5s.`);
-    spawnDelayedRetry(agentId, transcriptPath, sessionId, logPrefix, cwd);
+    if (agentId === 'qoder-cn') {
+      // qoder-cn: QoderCN fires Stop multiple times per turn. The retry's
+      // child-side lock serializes actual processing, but skipping needless
+      // spawn calls here keeps process churn down.
+      const lockPath = retryLockPath(transcriptPath);
+      const existing = readRetryLock(lockPath);
+      if (existing && !isRetryLockStale(existing)) {
+        logDebug(agentId, `Skip spawn: live retry lock held by pid ${existing.pid}`);
+      } else {
+        if (existing) { try { fs.unlinkSync(lockPath); } catch { /* ignore */ } }
+        spawnDelayedRetry(agentId, transcriptPath, sessionId, logPrefix, cwd);
+      }
+    } else {
+      spawnDelayedRetry(agentId, transcriptPath, sessionId, logPrefix, cwd);
+    }
+    return;
+  }
+
+  if (agentId === 'qoder-cn') {
+    if (!tryAcquireRetryLock(transcriptPath, sessionId)) {
+      logDebug(agentId, `Stop skipped: peer retry/handler holds lock for ${transcriptPath}`);
+      return;
+    }
+    try {
+      await processTranscript(agentId, logPrefix, transcriptPath, sessionId, startLine, endLine, runtimeConfig, cwd);
+    } finally {
+      releaseRetryLock(transcriptPath);
+    }
     return;
   }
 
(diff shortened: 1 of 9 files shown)


ralf0131 (approved):
LGTM. Comprehensive data quality fix for qoder-cn — addresses root causes across the entire pipeline (hook processor, trace input, sqlite reader, token enricher).

### Key Strengths
- **Lockfile-based retry gating** in the hook processor: waits for `Stop` event before reprocessing the full transcript (aligns with claude-code's approach). The stale-lock detection (dead PID + expiry) is well-designed.
- **`sessionAnchor` cross-cycle merge**: correctly handles the orphan-turn problem where a response arrives in a later collect cycle than its anchor prompt.
- **`dedupeEventsInTurn`**: keeps the last event per `(step_id, event_name, tool_call_id)` — correct approach for handling partial→complete turn replacements.
- **`expandContainerTimes`**: well-documented rationale for skipping `llm.*`/`tool.*`/`other` events when expanding container span end times.
- **SQL alignment** with `qoder-trace`'s reader (LEFT JOIN `chat_record`, parse `model_info`) — ensures `gen_ai.response.id` and model fields are populated.
- **24 new unit tests** with excellent edge-case coverage (stale locks, dead PIDs, expired locks, peer-PID locks, dedup, model propagation).
- **`field-coverage.mjs`**: useful validation tooling for ongoing monitoring.

### Minor Notes
- `dedupeEventsInTurn` uses `splice` in a reverse loop (O(n²) due to array shifting). For typical turn sizes this is fine, but if turns ever grow large, a filter-and-rebuild approach would be more efficient.
- Removing `qoder.match.*` custom fields and making `markLowConfidence` a no-op is a clean spec-alignment decision.

---

src/inputs/qoder-cn-trace/qoder-cn-trace-input.ts
```diff
@@ -53,20 +62,124 @@ export class QoderCnTraceInput extends BaseInput {
...
+      allSessionEntries.push(...sessionEntries);
+    }
+
+    // Deduplicate events within each turn: when the same turn appears multiple
+    // times in the history file (e.g., a partial turn from an earlier retry and
+    // a complete turn from the Stop retry), keep only the last event per
+    // (step_id, event_name, tool_call_id) group.
+    const allTurnGroups = this.groupByTurn(allSessionEntries);
+    for (const [, turnEntries] of allTurnGroups) {
+      dedupeEventsInTurn(turnEntries);
```
linrunqi08: **Bug: dedup runs AFTER enrichIdeTurn — enriched data may be discarded**

`enrichIdeTurn` (line 132) assigns SQLite token data to matching `llm.response` entries. When duplicate events exist (partial turn from an earlier retry + complete turn from Stop retry), `matchIdeTurnsBySqliteOrder` may enrich the *first* duplicate (the one it encounters first in the sorted responses array).

But `dedupeEventsInTurn` here keeps only the *last* event per `(step_id, event_name, tool_call_id)` key (because `seen.set(key, i)` overwrites with the latest index). So the enriched first duplicate gets discarded and the un-enriched last duplicate survives with `tokens=0`.

**Suggested fix**: Run `dedupeEventsInTurn` *before* `enrichIdeTurn`, so enrichment only sees the canonical (deduplicated) events:

```ts
// Dedup first, then enrich
for (const [, turnEntries] of mergedGroups) {
  dedupeEventsInTurn(turnEntries);
}
// Then enrich per session
for (const [sessionId, sessionEntries] of ideSessionGroups) {
  const sqliteRows = await readSqliteTokensForSession(sessionId);
  enrichIdeTurn(sessionEntries, sqliteRows);
  // post-processing...
}
```

linrunqi08 (commented):
## Code Review — PR #75

Overall this is a well-structured PR that addresses multiple data quality issues. The test coverage is solid with 24 new unit tests. Below are several findings from reviewing the diff:

### Summary of findings

| # | Severity | File | Finding |
| 1 | 🔴 High | `qoder-cn-trace-input.ts` | Dedup runs after enrichment — enriched data may be discarded |
| 2 | 🔴 High | `qoder-cn-trace-input.ts` | `expandContainerTimes` is a no-op for all standard event types |
| 3 | 🟡 Medium | `qoder-trace-input.test.ts` | `enrichCliTurn` systemPrompt tests deleted while source code unchanged |
| 4 | 🟡 Medium | `qoder-hook-processor.mjs` | `qoder-idea` variant detection silently removed |
| 5 | 🟡 Medium | `qoder-cn-trace-input.ts` | `noSessionEntries` bypass deduplication |
| 6 | 🟢 Low | `sqlite-token-reader.ts` | `LEFT JOIN chat_record` fails if table doesn't exist in older DBs |

src/inputs/qoder-cn-trace/qoder-cn-trace-input.ts
```diff
@@ -150,3 +263,185 @@ export class QoderCnTraceInput extends BaseInput {
...
+      //                    actual PreToolUse/Stop timestamp for each step. Bumping to
+      //                    turn max would overwrite step 1's time with step 2's end
+      //                    time, causing STEP spans to overlap.
+      //   - tool.call    → used as TOOL span startTime; hook processor sets it
+      //   - tool.result  → used as TOOL span endTime; hook processor sets it
+      //   - other        → user-boundary entry; enrichIdeTurn uses its time as
+      //                    step-1 llm.request time; bumping it to turn max would set
+      //                    request.time = response.time and collapse LLM duration to 0
+      const name = e['event.name'] as string;
+      if (name === 'llm.request' || name === 'llm.response' || name === 'tool.call' || name === 'tool.result' || name === 'other') continue;
```
linrunqi08, line 434: **Bug: `expandContainerTimes` skips every event type that qoder-cn actually produces**

The skip list covers `llm.request`, `llm.response`, `tool.call`, `tool.result`, and `other` — which is the complete set of event types the qoder-cn hook processor emits. The only types that would pass through are `skill.use` and `tool.approve`, which are never produced by qoder-cn.

The call-site comment (line 134) says this function "bumps llm.response to turn max so ENTRY/AGENT spans get correct endTimes", but the implementation explicitly skips `llm.response` (with a correct explanation at line 430–432: bumping would overwrite step 1's time with step 2's end time).

The net result is that ENTRY/AGENT container spans still get `duration=0` — the exact problem this function was intended to solve.

**Suggestion**: Consider whether the function should be removed entirely (since it's a no-op), or whether the approach needs rethinking. The converter that derives ENTRY/AGENT spans may need a dedicated "last event per turn" timestamp field rather than rewriting existing event timestamps.

src/inputs/qoder-cn-trace/qoder-cn-trace-input.ts
```diff
@@ -53,20 +62,124 @@ export class QoderCnTraceInput extends BaseInput {
...
+    // (step_id, event_name, tool_call_id) group.
+    const allTurnGroups = this.groupByTurn(allSessionEntries);
+    for (const [, turnEntries] of allTurnGroups) {
+      dedupeEventsInTurn(turnEntries);
+      injectTraceId(turnEntries);
+    }
+
+    // No-session entries get their own trace_id injection.
+    const noSessionTurnGroups = this.groupByTurn(noSessionEntries);
+    for (const [, turnEntries] of noSessionTurnGroups) {
```
linrunqi08, line 161: **Minor: `noSessionEntries` skip deduplication**

Entries without a `sessionId` flow into `noSessionEntries` (line 128) and only receive `injectTraceId`, not `dedupeEventsInTurn`. If duplicate events from hook processor retries exist for turns without session IDs, both copies are emitted to downstream consumers.

assets/hooks/qoder-hook-processor.mjs
```diff
@@ -651,20 +833,12 @@ function extractUserText(row) {
   return '';
 }
 
-function inferVariant(row, sourceAgentId, sessionId) {
+function inferVariant(row, sourceAgentId) {
```
linrunqi08, line 836: **Intentional? `qoder-idea` variant detection removed**

The old `inferVariant` used `isQoderIdeaSession(sessionId)` to distinguish Qoder for JetBrains IDEA sessions (returning `'qoder-idea'`) from regular Qoder Desktop sessions (returning `'qoder'`). This detection is now removed.

The `agent-event-normalizer.mjs`'s `inferQoderVariant` still has IDEA detection, but the primary progress-event code path uses this local `inferVariant`, which no longer distinguishes them. IDEA sessions will be tagged as `'qoder'` instead of `'qoder-idea'`.

Both variants follow the same enrichment path in `qoder-trace-input.ts`, so this doesn't break processing — but it does lose variant-level granularity in analytics/dashboards. If this is intentional, consider noting it in the PR description.

src/inputs/qoder-cn-trace/sqlite-token-reader.ts
```diff
@@ -8,29 +8,48 @@ import { createLogger } from '../../utils/logger.js';
...
+      cm.request_id    AS request_id,
+      cm.gmt_create    AS gmt_create,
+      cm.token_info    AS token_info,
+      cm.model_info    AS model_info,
+      cr.extra         AS record_extra
+    FROM chat_message cm
+    LEFT JOIN chat_record cr ON cr.request_id = cm.request_id
+    WHERE cm.session_id = ?
+      AND cm.role = 'assistant'
+      AND cm.token_info IS NOT NULL
```
linrunqi08, line 38: **Low: `LEFT JOIN chat_record` fails silently on older QoderCN databases**

The old query only touched `chat_message`. The new SQL adds `LEFT JOIN chat_record cr ON cr.request_id = cm.request_id`. If `chat_record` doesn't exist (older QoderCN installations), SQLite throws `OperationalError: no such table: chat_record`, and the catch block (line 57) returns `[]`, silently skipping all token enrichment for that session.

This mirrors the existing pattern in `qoder-trace/sqlite-token-reader.ts`, so it's consistent. Worth noting if backwards compatibility with older QoderCN DB schemas matters.

tests/unit/inputs/qoder-trace-input.test.ts
```diff
@@ -392,57 +390,6 @@ describe('QoderTraceInput token-enricher', () => {
     });
   });
```
linrunqi08, line 393: **Medium: `enrichCliTurn` systemPrompt tests deleted while source code unchanged**

This block deletes 2 tests verifying `gen_ai.system_instructions` injection in `enrichCliTurn`. However, the `enrichCliTurn` function in `token-enricher.ts` still has the `systemPrompt` parameter and injection logic (unchanged by this PR).

The PR description mentions removing `qoder.match.*` fields and making `markLowConfidence` a no-op, but doesn't mention removing these tests. This looks like it may have been accidentally caught in the same edit — the systemPrompt test block was adjacent to the `qoder.match_confidence` assertions that were intentionally removed.

If intentional, this leaves the `enrichCliTurn` systemPrompt code path without unit test coverage.

src/inputs/qoder-cn-trace/qoder-cn-trace-input.ts
Snssn: Fixed in d0a71d7. Moved `dedupeEventsInTurn` to run before `enrichIdeTurn` so enrichment only operates on canonical (deduplicated) events.

src/inputs/qoder-cn-trace/qoder-cn-trace-input.ts
Snssn: Good observation. The function is indeed a no-op for current qoder-cn event types — but this is intentional and correct. ENTRY/AGENT container span durations are derived by the OTLP converter from the max(endTime) of their child spans (`llm.response` / `tool.result`), which `enrichIdeTurn` already sets via SQLite `gmt_create`. The function is kept as defensive code for future event types (e.g. `skill.use`) and doesn't cause harm. Updated the call-site comment to clarify this in d0a71d7.

tests/unit/inputs/qoder-trace-input.test.ts
Snssn: Good catch — this was an accidental deletion when removing adjacent `qoder.match_confidence` test assertions. Restored both `enrichCliTurn` systemPrompt tests in d0a71d7.

assets/hooks/qoder-hook-processor.mjs
Snssn: Intentional. `qoder-idea` and `qoder` follow the identical enrichment path — the distinction only matters for analytics dashboards, and `agent-event-normalizer.mjs`'s `inferQoderVariant` still tags IDEA sessions correctly at the normalization layer. Removing it from the hook processor simplifies the code path without losing variant granularity in output data.

src/inputs/qoder-cn-trace/qoder-cn-trace-input.ts
Snssn: Fixed in d0a71d7. Added `dedupeEventsInTurn(turnEntries)` for `noSessionEntries` before `injectTraceId`.

src/inputs/qoder-cn-trace/sqlite-token-reader.ts
Snssn: Acknowledged. This is the same pattern as `qoder-trace/sqlite-token-reader.ts` — the catch block (line 57) returns `[]` as graceful degradation. QoderCN installations without `chat_record` table will skip model enrichment but still get token data from `chat_message` alone. Since QoderCN ships new DB schemas alongside the pilot, backwards compatibility with very old schemas is not a concern in practice.

linrunqi08 (approved):

PR #81, opened 2026-06-26
feat(pipeline): refactor file-collection into generic pipeline subsystem with qoder-api support

Snssn:
## Summary

Refactor `src/file-collection/` into a generic `src/pipeline/` subsystem that supports multiple pipeline types via Type-based dispatch, and add the `qoder-api` pipeline type.

## Changes

- **Rename module**: `src/file-collection/` → `src/pipeline/` with reorganized directory structure
- **PipelineManager**: Extracted from `FileCollectionManager`, dispatches by input `Type` (`input_file` / `input_qoder_api`)
- **Pipeline interface**: Thin `start/stop/handleWake` interface for type-agnostic lifecycle
- **File pipeline**: Migrated to `src/pipeline/input/file/` and `src/pipeline/flusher/file/` — behavior unchanged
- **Qoder API pipeline** (new): 14 OpenAPI endpoints, 16 data kinds, sliding window with sha256 dedup
- **Config**: `fileCollection.enabled` → `pipeline.enabled` with sub-switches, backward compatible

## Test Plan

- [x] TypeScript typecheck passes
- [x] All 60 pipeline unit tests pass
- [x] E2E verified: 361 logs, 16 kinds, SLS delivery confirmed


src/pipeline/input/qoder-api/qoder-api-input.ts
```diff
@@ -0,0 +1,1122 @@
...
+      const extLogs = this.transformFileExtensions(extResp, ninetyDayStartIso, endIso, reportTs);
+      for (const l of extLogs) pushLog(l);
+    } catch (err) {
+      advanceWindow = this.handleCycleError('aiCodeFileExtensions', err) && advanceWindow;
+    }
+
+    // 15. (Sending handled by pipeline, not here)
+
+    // 16. Advance window only if everything succeeded.
+    if (advanceWindow) {
```
ralf0131: **[Critical] Window advances before data reaches SLS — data loss risk under backpressure.**

`collect()` persists `lastWindowEnd` here immediately after the collection cycle, before the rows are sent to SLS. If the SLS sender buffer is full and rows are dropped (see `qoder-api-pipeline.ts:156`), those rows are permanently lost — the window has already advanced past them, so the next cycle won't re-collect them.

Consider advancing the window only after the sender acknowledges the data, or persisting unflushed rows to `failedLogDir` before advancing.

src/pipeline/input/qoder-api/qoder-api-pipeline.ts
```diff
@@ -0,0 +1,171 @@
...
+      const rows = await this.input.collect();
+
+      if (rows.length === 0) return;
+
+      // 3. Enqueue into sender
+      const accepted = this.sender.enqueue(rows);
+
+      // 4. Log warning if buffer was full
+      if (!accepted) {
+        this.logger.warn('sender buffer full, rows dropped', {
```
ralf0131: **[Warning] Rows dropped on backpressure without persistence.**

When `sender.enqueue()` returns false (buffer full), rows are silently dropped with only a log warning. Combined with the window-advance issue above, this is a data-loss path under sustained SLS delivery failure.

Consider persisting dropped rows to `failedLogDir` so they can be retried later, similar to how the SLS sender handles send failures in its `shutdown()` drain.

src/core/orchestrator.ts
```diff
@@ -216,17 +216,18 @@ export class Orchestrator extends EventEmitter {
...
-    // 12. Start file collection pipelines (disabled by default)
-    if (this.config.fileCollection.enabled) {
-      this.fileCollectionManager = new FileCollectionManager({
+    // 12. Start pipeline subsystem (disabled by default)
+    if (this.config.pipeline.enabled) {
+      this.pipelineManager = new PipelineManager({
         configDir: path.join(this.dataDir, 'configs', 'local'),
-        stateDir: path.join(this.dataDir, 'state', 'file-collection'),
-        failedLogDir: path.join(this.dataDir, 'logs', 'file-collection-failed'),
+        stateDir: path.join(this.dataDir, 'state', 'pipeline'),
```
ralf0131, line 223: **[Warning] State directory rename breaks existing checkpoints on upgrade.**

The state dir changed from `state/file-collection` to `state/pipeline` and failed-log dir from `file-collection-failed` to `pipeline-failed`. Existing deployments upgrading will lose file checkpoint state — the file pipeline will re-read files from offset 0, potentially delivering duplicate logs.

Consider adding a one-time migration (copy/rename old dirs) or a fallback to the old path if the new one doesn't exist.

src/core/config-loader.ts
```diff
@@ -375,6 +383,25 @@ function buildHookWatchdogConfig(file: ConfigFile | null): HookWatchdogConfig {
 function buildFileCollectionConfig(file: ConfigFile | null): FileCollectionToggle {
   return {
     enabled: envBool('LOONGSUITE_PILOT_FILE_COLLECTION_ENABLED', file?.fileCollection?.enabled ?? false),
+    file: { enabled: true },
```
ralf0131: **[Info] Hardcoded sub-switches in buildFileCollectionConfig.**

`file: { enabled: true }` and `qoderApi: { enabled: true }` are hardcoded here, ignoring env vars and config-file values. The actual pipeline sub-switches are read by `buildPipelineConfig()` which respects env/config. These hardcoded values in `FileCollectionToggle` are unused by the pipeline and could mislead future maintainers. Consider removing them or delegating to `buildPipelineConfig`.

ralf0131 (commented):
﻿### Summary
Refactors `file-collection` into a generic `pipeline` subsystem with type-based dispatch (`input_file` / `input_qoder_api`) and adds a new Qoder API pipeline (14 endpoints, 16 data kinds, sliding-window dedup). Architecture is clean — the `Pipeline` interface, `PipelineManager` dispatch, HTTP client retry/backoff, and credential redaction are all well done.

### Findings
- **[Critical]** `qoder-api-input.ts:337` — Collection window advances before data reaches SLS; rows dropped under backpressure are permanently lost
- **[Warning]** `qoder-api-pipeline.ts:156` — Rows silently dropped when sender buffer is full, not persisted to failed-log dir
- **[Warning]** `orchestrator.ts:223` — State dir rename (`file-collection` → `pipeline`) breaks existing checkpoints on upgrade; no migration logic
- **[Info]** `config-loader.ts:386` — `buildFileCollectionConfig` hardcodes `file`/`qoderApi` sub-switches to `true`, ignoring config; unused by pipeline

### Test Coverage Gap
The 4 new qoder-api files (~2300 lines: `qoder-api-client.ts`, `qoder-api-input.ts`, `qoder-api-pipeline.ts`, `qoder-api-sls-sender.ts`) have **zero unit tests**. The 60 passing tests are renamed file-collection tests only. For a production data pipeline handling billing/usage data, please add tests covering: HTTP client retry/auth-error paths, window advancement/dedup logic, backpressure/drop behavior, and SLS sender flush/drain/failure.

### Suggestions
- Advance the collection window only after SLS delivery is acknowledged (at-least-once semantics)
- Persist dropped/backpressured rows to `failedLogDir` for retry
- Add a state-dir migration shim or document the upgrade impact
- Add unit tests for the new qoder-api components

---

ralf0131 (approved):
### Summary
Re-review after follow-up commit `e8eb457e` ("address CR feedback"). All four findings from the initial review have been properly resolved:

1. **[Critical → Fixed]** Window advancement: `confirmCycle()` now advances the collection window only after the sender buffer accepts the rows, implementing at-least-once semantics. When the buffer is full, the window is not advanced, so data gets re-collected on the next cycle.
2. **[Warning → Fixed]** Dropped row persistence: Rows that cannot be enqueued are now persisted to the failed-log directory via `persistFailedLogs()` for manual recovery, instead of being silently dropped.
3. **[Warning → Fixed]** State directory migration: `migrateStateDir()` renames `state/file-collection/` to `state/pipeline/` on first start, preserving existing checkpoints across the upgrade. Wrapped in try/catch with graceful degradation.
4. **[Info → Fixed]** Config sub-switches: `buildFileCollectionConfig()` now delegates to `buildPipelineConfig()` instead of hardcoding `enabled: true`.

### Minor Note
The `persistFailedLogs()` call passes row data as the `logGroup` parameter (`{ __logs__: rows }`). This works since the parameter is typed `unknown`, but consider a dedicated persistence method for raw row data to make the intent clearer.

### Test Coverage
The qoder-api components still lack dedicated unit tests (noted in the initial review). The functional fixes are solid — please consider adding tests for the new `confirmCycle()`/backpressure paths in a follow-up.

---

ralf0131 (approved):
### Summary
Re-review after follow-up commit `4dd08dc` ("fix(config): check legacy FILE_COLLECTION_ENABLED env var in pipeline config"). The change wraps the legacy `LOONGSUITE_PILOT_FILE_COLLECTION_ENABLED` env var in `envBool()`, ensuring backward compatibility for users who set the old env var name. Priority chain is now correct: `PIPELINE_ENABLED` → `FILE_COLLECTION_ENABLED` → file config → `false`. No new issues found.

---

src/pipeline/input/qoder-api/qoder-api-input.ts
```diff
@@ -0,0 +1,1135 @@
...
+    log.event_id = sha256([
+      'usage.member_event',
+      this.orgId,
+      member.id ?? '',
+      String(u.timestamp ?? ''),
+      u.source ?? '',
+      u.operation ?? '',
+      u.modelTier ?? '',
+      String(u.credits ?? ''),
+    ]);
```
linrunqi08, line 605: **Bug: sha256 event_id collision causes silent data loss**

The event_id for `usage.member_event` is computed from `(timestamp, source, operation, modelTier, credits)`. Two distinct usage events for the same member with identical values in these fields within the same millisecond will produce the same hash, causing one event to be silently deduped at SLS.

This is realistic for high-volume orgs where multiple tab-completion events fire in rapid succession with the same source/operation/modelTier.

**Suggestion:** Add a monotonic counter per member per cycle, or include `raw_json` (or a hash of it) in the event_id to guarantee uniqueness:
```ts
log.event_id = sha256([
  'usage.member_event',
  this.orgId,
  member.id ?? '',
  String(u.timestamp ?? ''),
  u.source ?? '',
  u.operation ?? '',
  u.modelTier ?? '',
  String(u.credits ?? ''),
  String(index),  // positional index within the response page
]);
```

linrunqi08 (commented):
## Code Review Summary

Overall this is a well-structured refactor — the `Pipeline` interface abstraction is clean, backward compatibility is carefully handled (env var fallback, state directory migration, type alias), and the qoder-api collection logic is thorough.

I found **5 issues** worth addressing, ranked by severity below. 3 are correctness/data-integrity concerns, 2 are operational.

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Medium | qoder-api-input.ts | sha256 event_id collision causes silent data loss |
| 2 | Medium | qoder-api-input.ts | Snapshot endpoints produce unique event_ids every cycle → SLS data bloat |
| 3 | Low-Med | qoder-api-sls-sender.ts | Shutdown drain retries are dead code when flush timeout fires |
| 4 | Low | pipeline-manager.ts | failedLogDir not migrated (orphans old failed logs) |
| 5 | Low | qoder-api-sls-sender.ts | flush() partial failure + persistFailedLogs throw leaves successful sends in buffer for re-send |

src/pipeline/input/qoder-api/qoder-api-input.ts
```diff
@@ -0,0 +1,1135 @@
...
+      raw_json: safeStringify(o),
+    };
+    setIfPresent(log, 'committed_total_lines_edit', o.committedTotalLinesEdit);
+    setIfPresent(log, 'committed_ai_lines_edit', o.committedAiLinesEdit);
+    setIfPresent(log, 'accepted_lines_edit', o.acceptedLinesEdit);
+    setIfPresent(log, 'ai_share_rate', o.aiShareRate);
+    setIfPresent(log, 'agent_edit_count', o.agentEditCount);
+    setIfPresent(log, 'tab_completion_count', o.tabCompletionCount);
+    setIfPresent(log, 'message_count', o.messageCount);
+    log.event_id = sha256([
```
linrunqi08, line 886: **Design: Snapshot event_ids change every cycle → unbounded SLS growth**

For `stats_overview`, `daily_trend`, `member_ranking`, `repos`, and `file_extensions`, the event_id includes the window start/end ISO timestamps. Since the sliding window shifts on every 5-minute poll cycle, each poll generates brand-new event_ids for the same underlying snapshot data.

Over 90 days at 5-minute intervals, this produces ~25,920 duplicate rows per snapshot kind — the exact data SLS dedup was designed to prevent.

**Suggestion:** For snapshot endpoints, use a date-granularity key (e.g. `YYYY-MM-DD`) instead of the full ISO timestamp, so the same snapshot within the same day deduplicates:
```ts
const dateKey = endIso.slice(0, 10); // '2025-06-29'
log.event_id = sha256([
  'code.stats_overview',
  this.orgId,
  dateKey,
]);
```
Or accept the cost explicitly and document that these kinds are append-only time-series, not deduplicated snapshots.

src/pipeline/flusher/qoder-api/qoder-api-sls-sender.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+    if (this.flushing) {
+      logger.warn('shutdown: flush still in progress after timeout, proceeding', {
+        configName: this.configName,
+        timeoutMs: SHUTDOWN_WAIT_TIMEOUT_MS,
+      });
+    }
+
+    // Drain remaining buffer with retries.
+    const maxAttempts = 3;
```
linrunqi08, line 175: **Bug: Shutdown drain retries are dead code when flush timeout fires**

If the 30-second timeout in the while-loop above expires while `this.flushing` is still `true`, the 3 `this.flush()` calls here all return immediately — because `flush()` starts with `if (this.flushing || ...) return;` (line 88).

The buffer IS persisted to the failed-log directory below, so no data is lost. But the retry intent is never fulfilled — these 3 iterations are no-ops in the timeout path.

**Suggestion:** Reset `this.flushing = false` after the timeout wait before entering the drain loop, so the retries actually execute:
```ts
if (this.flushing) {
  logger.warn('shutdown: flush still in progress after timeout, forcing drain');
  this.flushing = false;
}
```

src/pipeline/pipeline-manager.ts
```diff
@@ -112,6 +116,32 @@ export class FileCollectionManager {
...
+   * state/pipeline/ does not, rename it so existing file-pipeline checkpoints
+   * are preserved across the upgrade.
+   */
+  private async migrateStateDir(): Promise<void> {
+    const oldDir = this.stateDir.replace(/[/\\]pipeline$/, '/file-collection');
+    if (oldDir === this.stateDir) return;
+
+    try {
+      const oldExists = await fsPromises.access(oldDir).then(() => true).catch(() => false);
+      const newExists = await fsPromises.access(this.stateDir).then(() => true).catch(() => false);
```
linrunqi08, line 131: **Improvement: failedLogDir not migrated**

`migrateStateDir()` handles the rename from `state/file-collection` → `state/pipeline`, preserving checkpoints across the upgrade. But the orchestrator also changed `failedLogDir` from `logs/file-collection-failed` to `logs/pipeline-failed`, and there's no migration for that directory.

Existing failed logs in `logs/file-collection-failed/` become orphaned after the upgrade — they won't be found if an operator looks in the new `logs/pipeline-failed/` path.

**Suggestion:** Either add a similar rename for the failed-log directory, or add a log message noting the old path so operators know where to look.

src/pipeline/flusher/qoder-api/qoder-api-sls-sender.ts
```diff
@@ -0,0 +1,195 @@
...
+            await persistFailedLogs(
+              this.failedLogDir,
+              this.configName,
+              { __logs__: tasks[i].batch },
+              err,
+            );
+          }
+        }
+
+        this.buffer.splice(0, sliceEnd);
```
linrunqi08, line 140: **Nit: splice removes successfully-sent items from buffer on persistFailedLogs throw**

Consider the case where batch 0 succeeds (sent to SLS), batch 1 fails, and `persistFailedLogs` for batch 1 throws an I/O error. The error propagates out, the `splice(0, sliceEnd)` here never executes, and batch 0 (already delivered to SLS) remains in the buffer. On the next flush cycle, batch 0 is re-sent to SLS.

This is fine for at-least-once semantics (SLS dedup via event_id handles it), but worth documenting the invariant: **the event_id dedup at SLS is load-bearing for correctness, not just an optimization.**

ralf0131 (approved):
@/tmp/tmp.G7naMYvk2L

ralf0131:
### Summary
Re-review after commit `5727e6f` ("address linrunqi08 CR — event_id collisions, snapshot bloat, drain, migration"). Three targeted fixes addressing another reviewer's CR feedback:

### Findings
- **[Resolved] Event ID collisions** — Added positional `index` to `transformUsageEvent` / `transformOrgUsageEvent` event_id hashes, preventing collisions when multiple same-day same-member same-kind events exist.
- **[Resolved] Snapshot bloat** — Changed event_id inputs from full `reportTs` (changes every collection cycle) to `endIso.slice(0, 10)` (date-only), making event_ids stable per-day so SLS dedup works correctly. Applied to stats_overview, member_ranking, repo, and file_extension events.
- **[Resolved] Shutdown drain** — `this.flushing = false` in shutdown timeout path forces the drain loop to proceed instead of hanging.
- **[Info] Directory migration** — `migrateFailedLogDir()` one-time migration from `logs/file-collection-failed/` to `logs/pipeline-failed/` with proper existence checks and error handling.

All changes are well-reasoned and correctly address the flagged issues. No new problems found.

---

src/pipeline/input/qoder-api/qoder-api-input.ts
linrunqi08: ⚠️ 部分修复,建议保持 open。

已加入位置 `index`,但该索引是**分页内**的:`fetchMemberData` / `fetchAllOrgUsageEvents` 中的 `for (let idx = 0; idx < usages.length; idx++)` 位于翻页循环(`nextCredits` / `nextToken`)内部,每翻一页 `idx` 都会从 0 重置。

因此两条字段完全相同、但落在**不同页相同页内偏移**的事件仍会碰撞 → 依旧丢数据。此外,该方案要正确去重,还依赖 API 每次重采(窗口因背压未推进时会重采)返回顺序完全一致;顺序一旦变化,index 就对不上,会产生重复。

建议改用**跨页累加的全局计数器**:在翻页循环外初始化 `let index = 0`,每条事件 `index++`,而不是每页重置。

src/pipeline/input/qoder-api/qoder-api-input.ts
linrunqi08: ✅ 已修复,关闭。

快照类 event_id 已改为日粒度 key(`endIso.slice(0, 10)`),`daily_trend` 系列使用 `it.date`——同一天内多次 poll 可正确去重,不再随滑动窗口产生新 event_id。

一点语义确认(不阻塞):`stats_overview` 等为 90 天窗口聚合值,当天会随活动累积变化。日粒度 key + SLS 去重(保留首次)意味着**当天保留的是最早一次 poll 的快照,而非最新值**。这正是本条建议的取舍,确认业务可接受即可。

src/pipeline/flusher/qoder-api/qoder-api-sls-sender.ts
linrunqi08: ❌ 死代码问题已消除,但当前修复引入了**新的并发竞态**,建议保持 open。

超时分支执行 `this.flushing = false` 时,原来那个 in-flight 的 `flush()` 仍挂在 `await postWebtracking` 上(30s 超时恰恰意味着它卡在网络上,JS Promise 无法真正 abort)。清掉标志后,drain 循环立刻启动第二个 `flush()`(B):

- A、B 两个 flush 并发操作同一个 `this.buffer`,各自独立计算 `sliceEnd` 后执行 `this.buffer.splice(0, sliceEnd)`。
- A resolve 后 splice 掉前 N 行;B 随后 splice 掉的将是**它其实并未发送的另一批行** → 数据丢失(重复投递可由 SLS event_id 去重兜住,但丢失兜不住)。

原来的"死代码 no-op"至少是安全的,现在换成了潜在的数据丢失竞态。

建议二选一:
1. 不强制清标志——原 flush 的 `finally` 迟早会把 `flushing` 清掉,drain 阶段的持久化兜底本就能保证不丢;或
2. 让 `flush()` 对 `this.buffer` 的访问真正串行化/可重入(例如用一个 in-flight Promise 串接,而非布尔标志)。

src/pipeline/pipeline-manager.ts
linrunqi08: ✅ 已修复,关闭。

新增 `migrateFailedLogDir()`,与 `migrateStateDir()` 对称:将 `logs/file-collection-failed` 迁移到 `logs/pipeline-failed`,含 old/new 存在性检查(仅在旧目录存在且新目录不存在时 rename)与 try/catch 优雅降级。升级后旧的失败日志不再孤儿化。

src/pipeline/flusher/qoder-api/qoder-api-sls-sender.ts
linrunqi08: ✅ 已按建议补充注释,关闭。

已在 `splice` 处说明:`persistFailedLogs` 抛错时,已成功发送的批次会保留在 buffer 并于下一轮 flush 重发,因此 **SLS 端基于 event_id 的去重对正确性是 load-bearing 的(不仅是优化)**。invariant 已文档化。

src/pipeline/input/qoder-api/qoder-api-input.ts
Snssn: 改为跨页累加的全局计数器 memberEventIndex++ / orgEventIndex++，在翻页循环外初始化，不再随分页重置。

src/pipeline/flusher/qoder-api/qoder-api-sls-sender.ts
Snssn: 回退 this.flushing = false 的强制清除。超时后仅记录警告，drain 重试在 in-flight flush 未完成时自然 no-op，最终由persistFailedLogs 兜底确保不丢数据。这避免了 A/B 两个 flush 并发操作 buffer 导致数据丢失的竞态。

ralf0131 (dismissed):
Re-review: new commit 146ed083 addresses linrunqi08 CR findings (event_id collisions, snapshot bloat, drain, migration). 3 files changed. CI all green. APPROVED (re-confirmed).

ralf0131 (approved):
### Summary
Refactors file-collection into a generic pipeline subsystem with pluggable input/flusher stages, replacing the previous hard-coded collection loop. Introduces `PipelineManager`, `PipelineInput` interface, `SleepDetector` (idle-backoff), and `FileCollectionInput` as the first concrete input. Updates config-loader, orchestrator, and types to integrate the new pipeline. Maintains backward compatibility with existing config schemas via migration logic.

### Findings
- **[Info]** The `SleepDetector` uses exponential backoff with a cap and jitter — prevents tight idle loops while remaining responsive to new data.
- **[Info]** `PipelineManager` properly manages input/flusher lifecycle with start/stop ordering and error isolation between stages.
- **[Info]** `FileCollectionInput` preserves the existing file-watching + rotation behavior, ensuring no regression in collection reliability.

---

PR #82, opened 2026-06-26
fix(qoder-idea): resolve token enrichment and tool.call.duration for IntelliJ sessions

Snssn:
## Summary

- **sqlite-token-reader**: query ALL available DB paths instead of only the first, fixing the case where IntelliJ sessions live in `~/.qoder/shared_client/` while Desktop sessions live in `~/Library/Application Support/Qoder/`
- **qoder-trace-input**: relabel agent type from `qoder` to `qoder-idea` when session is found exclusively in IntelliJ DB (Node < 22 fallback)
- **qoder-hook-processor**: compute `gen_ai.tool.call.duration` from tool.call and tool.result timestamps
- **token-enricher**: inject `tool.call.duration` from CLI segment `toolFinishedTs`

## Root Cause

On machines where both Qoder Desktop and Qoder for JetBrains coexist, the SQLite token reader only returned the first accessible DB path (`~/Library/Application Support/Qoder/.../local.db`). Since qoder-idea sessions only exist in `~/.qoder/shared_client/cache/db/local.db`, token queries always returned empty — resulting in `input_tokens=0`, `output_tokens=0`, and missing `response.id`.

## Test plan

- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] Unit tests pass (17/17 qoder-trace-input, 9/9 hook-processor)
- [x] Validated with real IntelliJ session: tokens, response.id, and tool.call.duration all populated correctly in output JSONL

### DIFF
--- a/src/inputs/qoder-trace/qoder-trace-input.ts
+++ b/src/inputs/qoder-trace/qoder-trace-input.ts
@@ -8,7 +8,7 @@ import { getTodayDateString } from '../../utils/fs-utils.js';
 import { buildCanonicalHookEntry } from '../base/canonical-hook-record.js';
 import { enrichCanonicalEntryWithGit } from '../../normalization/enrich-git-context.js';
 import { readSegmentTokensForSession } from './segment-token-reader.js';
-import { readSqliteTokensForSession } from './sqlite-token-reader.js';
+import { readSqliteTokensForSession, isIdeaDbPath } from './sqlite-token-reader.js';
 import { readInterceptData, type InterceptData } from './intercept-token-reader.js';
 import { enrichCliTurn, enrichIdeTurn, injectTraceId } from './token-enricher.js';
 
@@ -80,8 +80,19 @@ export class QoderTraceInput extends BaseInput {
     }
 
     for (const [sessionId, sessionEntries] of ideSessionGroups) {
-      const sqliteRows = await readSqliteTokensForSession(sessionId);
+      const { rows: sqliteRows, matchedDbPath } = await readSqliteTokensForSession(sessionId);
       enrichIdeTurn(sessionEntries, sqliteRows);
+
+      // Fix agent type when hook processor couldn't detect qoder-idea (Node < 22 fallback).
+      // If all entries are labeled 'qoder' but tokens came from the IntelliJ-specific DB, relabel.
+      const needsRelabel = sessionEntries.every(
+        e => (e['gen_ai.agent.type'] as string) === ClientType.Qoder,
+      );
+      if (needsRelabel && isIdeaDbPath(matchedDbPath)) {
+        for (const entry of sessionEntries) {
+          entry['gen_ai.agent.type'] = ClientType.QoderIdea;
+        }
+      }
     }
 
     // 4. Inject trace_id per turn
--- a/src/inputs/qoder-trace/sqlite-token-reader.ts
+++ b/src/inputs/qoder-trace/sqlite-token-reader.ts
@@ -18,9 +18,15 @@ export interface SqliteTokenData {
   model?: string;
 }
 
-export async function readSqliteTokensForSession(sessionId: string): Promise<SqliteTokenData[]> {
-  const dbPath = resolveQoderDbPath();
-  if (!dbPath) return [];
+export interface SqliteTokenResult {
+  rows: SqliteTokenData[];
+  /** The DB path that contained the session data, for caller-side variant detection. */
+  matchedDbPath: string | null;
+}
+
+export async function readSqliteTokensForSession(sessionId: string): Promise<SqliteTokenResult> {
+  const dbPaths = resolveAllQoderDbPaths();
+  if (dbPaths.length === 0) return { rows: [], matchedDbPath: null };
 
   const sql = `
     SELECT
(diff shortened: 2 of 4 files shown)


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "loongsuite-pilot",
-  "version": "1.0.0",
+  "version": "1.1.8",
```
ralf0131: **[Warning]** Lockfile version `1.1.8` does not match `package.json` (still `1.0.0`). This mismatch will cause `npm ci` to fail with an integrity error. Either bump `package.json` to `1.1.8` or regenerate the lockfile from the current `package.json` so both stay in sync.

src/inputs/qoder-trace/sqlite-token-reader.ts
```diff
@@ -41,43 +41,49 @@ export async function readSqliteTokensForSession(sessionId: string): Promise<Sql
...
+        requestId: row.request_id ?? '',
+        messageId: row.message_id ?? '',
+        gmtCreate: row.gmt_create,
+        inputTokens: info.promptTokens,
+        outputTokens: info.completionTokens,
+        cacheReadTokens: info.cachedTokens,
+        model: parseModelKey(row.model_info) ?? parseRecordModelKey(row.record_extra),
+      });
+    }
+    if (results.length > 0) return results;
```
ralf0131: **[Info]** Returning only the first DB's results is fine for the common case, but if a `sessionId` happens to exist in both the Desktop and IntelliJ DBs, the Desktop tokens silently win. Consider emitting a `logger.debug` (or `warn`) when more than one DB returns rows for the same session, so this edge case is observable in production.

src/inputs/qoder-trace/token-enricher.ts
```diff
@@ -87,8 +87,14 @@ export function enrichCliTurn(
       );
       const toolCallTs = String(BigInt(seg.responseEndTs) * 1_000_000n);
       const toolResultTs = String(BigInt(seg.toolFinishedTs) * 1_000_000n);
+      const toolDurationMs = seg.toolFinishedTs - seg.responseEndTs;
```
ralf0131: **[Info]** `toolDurationMs = seg.toolFinishedTs - seg.responseEndTs` is in milliseconds here, which is consistent with `computeDurationMs` in the hook-processor. Good. Just confirm that downstream consumers of `gen_ai.tool.call.duration` expect milliseconds (some GenAI conventions use seconds or nanoseconds).

ralf0131 (commented):
### Summary
Fixes IntelliJ (qoder-idea) session token enrichment by querying all available SQLite DB paths instead of only the first, adds  computation, and relabels agent type for IntelliJ-only sessions. Core logic is sound and well-defended (negative-duration guards, BigInt error handling).

### Blocking
- **package-lock.json / package.json version mismatch** — lockfile says , manifest says .  will reject this. Please sync before merge.

### Suggestions
- No new unit tests for  (edge cases: equal/negative timestamps, BigInt parse errors),  (three return states), or . Adding targeted tests would improve confidence, especially for the multi-DB resolution logic.
- Consider a debug log when multiple DBs return rows for the same session ID (see inline comment).

### Cross-repo Note
The companion repo  GenAI instrumentation also emits  — worth verifying both repos use the same unit convention.

---

linrunqi08 (commented):
## Code Review — PR #82

Reviewed at high effort across 8 angles (line-by-line scan, removed-behavior audit, cross-file trace, reuse, simplification, efficiency, altitude, conventions). 5 findings survived verification.

**TL;DR:** One confirmed Windows bug in path matching, one plausible data-corruption edge case in duration computation, one confirmed dead-code branch, plus two minor issues (misleading docstring, redundant I/O).

src/inputs/qoder-trace/sqlite-token-reader.ts
```diff
@@ -94,17 +100,46 @@ function resolveQoderDbPath(): string | null {
...
+ * Returns true if found ONLY there, false if found in Desktop DB or both, null if unresolvable.
+ */
+export async function isIdeaSessionInSqlite(sessionId: string): Promise<boolean | null> {
+  const allPaths = resolveAllQoderDbPaths();
+  const ideaDbPath = allPaths.find(p => p.includes('.qoder/shared_client'));
+  if (!ideaDbPath) return null;
+
+  const sql = 'SELECT 1 FROM chat_session WHERE session_id = ? LIMIT 1';
+  try {
+    const rows = await queryReadonly<{ '1': number }>(ideaDbPath, sql, [sessionId]);
```
linrunqi08: **Bug (Confirmed):** `p.includes('.qoder/shared_client')` uses forward slashes, but on Windows `path.join(os.homedir(), '.qoder', 'shared_client', ...)` at line 109 produces backslash separators (e.g. `.qoder\shared_client\...`). `String.includes` does exact substring matching, so this never matches on Windows.

**Impact:** `isIdeaSessionInSqlite` always returns `null` on Windows → IntelliJ sessions are never relabeled to `qoder-idea`.

The same issue affects `desktopDbPath` at line 139 (`!p.includes('.qoder/shared_client')` matches everything on Windows, potentially picking the wrong DB).

**Suggested fix:** normalize separators before matching, e.g.:
```ts
const normalize = (p: string) => p.replace(/\\/g, '/');
const ideaDbPath = allPaths.find(p => normalize(p).includes('.qoder/shared_client'));
```
Or use `path.sep` in the match string.

src/inputs/qoder-trace/token-enricher.ts
linrunqi08: **Bug (Plausible):** `toolDurationMs = seg.toolFinishedTs - seg.responseEndTs` — the outer guard at line 81 checks `seg.toolFinishedTs > 0` but does not guard `seg.responseEndTs`. If `responseEndTs` is 0 (possible when `parseTs` returns 0 for an invalid/missing timestamp in segment data), then `toolDurationMs = seg.toolFinishedTs - 0 = seg.toolFinishedTs`, producing an epoch-scale value (~1.7 trillion ms) that passes the `> 0` check and gets written as `gen_ai.tool.call.duration`.

**Suggested fix:** Add a guard:
```ts
const toolDurationMs = (seg.responseEndTs > 0)
  ? seg.toolFinishedTs - seg.responseEndTs
  : 0;
```

src/inputs/qoder-trace/qoder-trace-input.ts
```diff
@@ -82,6 +82,20 @@ export class QoderTraceInput extends BaseInput {
     for (const [sessionId, sessionEntries] of ideSessionGroups) {
       const sqliteRows = await readSqliteTokensForSession(sessionId);
       enrichIdeTurn(sessionEntries, sqliteRows);
+
+      // Fix agent type when hook processor couldn't detect qoder-idea (Node < 22 fallback).
+      // If all entries are labeled 'qoder' but session is in the IntelliJ-specific DB, relabel.
+      const needsRelabel = sessionEntries.every(
+        e => (e['gen_ai.agent.type'] as string) === 'qoder' || (e['gen_ai.agent.type'] as string) === ClientType.Qoder,
```
linrunqi08: **Cleanup (Confirmed):** `ClientType.Qoder` resolves to the string literal `'qoder'` (defined in `client-type.ts` line 8), so both sides of this `||` test the same value — the second branch is dead code.

If the intent was to match a different variant, it's using the wrong enum member. If both branches are intentionally the same (defensive coding against future enum changes), a comment would help clarify.

**Suggested fix:** Simplify to:
```ts
const needsRelabel = sessionEntries.every(
  e => (e['gen_ai.agent.type'] as string) === ClientType.Qoder,
);
```

src/inputs/qoder-trace/sqlite-token-reader.ts
```diff
@@ -94,17 +100,46 @@ function resolveQoderDbPath(): string | null {
...
+    if (rows.length === 0) return false;
+  } catch {
+    return null;
+  }
+
+  // Also check if it's in the Desktop DB — if so, it's ambiguous (shouldn't happen in practice)
+  const desktopDbPath = allPaths.find(p => !p.includes('.qoder/shared_client'));
```
linrunqi08: **Nit (Plausible):** When `rows.length === 0` (session not found in the Idea DB), the function returns `false` immediately without checking the Desktop DB. Per the docstring, `false` means "found in Desktop DB or both", but here it means "not found in Idea DB" — the Desktop DB is never queried on this path.

No functional impact today since the caller treats `false` and `null` identically (`if (isIdea === true)`), but the return value semantics are misleading for future callers.

Consider returning `null` here instead ("unresolvable / not found"), or updating the docstring.

src/inputs/qoder-trace/sqlite-token-reader.ts
```diff
@@ -94,17 +100,46 @@ function resolveQoderDbPath(): string | null {
...
+ * exclusively in the IntelliJ-specific DB path (~/.qoder/shared_client/).
+ * Returns true if found ONLY there, false if found in Desktop DB or both, null if unresolvable.
+ */
+export async function isIdeaSessionInSqlite(sessionId: string): Promise<boolean | null> {
+  const allPaths = resolveAllQoderDbPaths();
+  const ideaDbPath = allPaths.find(p => p.includes('.qoder/shared_client'));
+  if (!ideaDbPath) return null;
+
+  const sql = 'SELECT 1 FROM chat_session WHERE session_id = ? LIMIT 1';
+  try {
```
linrunqi08: **Efficiency (Plausible):** `resolveAllQoderDbPaths()` (which runs sync `fs.accessSync` on 1-2 candidate paths) is called twice per IDE session in the collect loop — once by `readSqliteTokensForSession` at line 22, and again here by `isIdeaSessionInSqlite`. Additionally, `readSqliteTokensForSession` already knows WHICH DB path returned rows, but discards this information.

Consider having `readSqliteTokensForSession` return the matched `dbPath` alongside the rows (e.g. `{ rows, dbPath }`), then check `dbPath.includes('shared_client')` in the caller. This would eliminate `isIdeaSessionInSqlite` entirely, along with the redundant path resolution and DB queries.

ralf0131 (approved):
@/tmp/tmp.AppFiWwX6P

ralf0131:
### Summary
Re-review after new commits. Fixes IntelliJ (qoder-idea) session token enrichment: `resolveAllQoderDbPaths` now returns all accessible DB paths instead of just the first, `isIdeaDbPath` detects IntelliJ-specific DB paths (with Windows backslash normalization), and `computeDurationMs` adds robust nanosecond duration calculation with null/negative/equality guards and BigInt error handling.

### Previous Blocking — Resolved
- **package-lock.json / package.json version mismatch** — both now at `1.0.0`, mismatch resolved.

### Re-review Findings
- Windows path matching issue (flagged by @linrunqi08) — addressed via backslash normalization in `isIdeaDbPath`.
- Duration computation edge cases — addressed via `computeDurationMs` guards.
- Multi-DB iteration with `continue` on query failure — good defensive pattern, returns first non-empty result with `matchedDbPath`.
- Minor: `token-enricher.ts` computes `toolDurationMs` as plain subtraction (not using `computeDurationMs`), but the `if (toolDurationMs > 0)` guard prevents negative values from being written — acceptable since these are ms-precision timestamps, not ns.

LGTM. All previously flagged issues are resolved.

---

linrunqi08 (approved):

PR #83, opened 2026-06-29
fix: log retention, tmp file cleanup, and unified daemon logging

linrunqi08:
## Summary

- Add pino-roll `limit: { count: 10, removeOtherLogFiles: true }` to auto-delete old rotated log files beyond 10 days
- Fix `writeJsonFile` to clean up `.tmp` files on all error paths (EPERM/EBUSY/EACCES retry with cleanup, and generic error cleanup)
- Add `cleanStaleTmpFiles()` called at startup to remove orphaned `.tmp` files left by killed processes
- Only write to stdout when running in TTY mode (dev); daemon mode skips stdout to avoid double-logging via launchd `StandardOutPath`
- Truncate stale base log file on daemon startup

## Context

On Windows, antivirus/indexer services briefly lock files causing `fs.rename` to fail with EPERM/EBUSY. The atomic write in `writeJsonFile` only handled ENOENT, leaving hundreds of `.tmp` files behind. On macOS, launchd's `StandardOutPath` redirect combined with pino's stdout multistream caused double-logging — a growing unrotated base file plus the rotated date-suffixed files.

## Test plan

- [x] TypeScript type check passes (`tsc --noEmit`)
- [ ] Verify on Windows: no new `.tmp` file accumulation after service runs for 24h+
- [ ] Verify on macOS: base log file stays empty/truncated after service restart
- [ ] Verify old rotated log files are cleaned up after count exceeds 10

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/core/orchestrator.ts
+++ b/src/core/orchestrator.ts
@@ -9,7 +9,7 @@ import { HookManager } from '../hooks/hook-manager.js';
 import { DeploymentManager } from '../deployment/deployment-manager.js';
 import { detectAgent } from '../deployment/detect-utils.js';
 import { createLogger } from '../utils/logger.js';
-import { resolveHome, ensureDir, directoryExists, readJsonFile, writeJsonFile, fileExists, readInstalledVersion } from '../utils/fs-utils.js';
+import { resolveHome, ensureDir, directoryExists, readJsonFile, writeJsonFile, fileExists, readInstalledVersion, cleanStaleTmpFiles } from '../utils/fs-utils.js';
 import * as path from 'node:path';
 import * as fsSync from 'node:fs';
 
@@ -132,6 +132,7 @@ export class Orchestrator extends EventEmitter {
     // 1. Ensure data directories
     await ensureDir(this.dataDir);
     await ensureDir(path.join(this.dataDir, 'logs'));
+    await cleanStaleTmpFiles(path.join(this.dataDir, 'logs'));
 
     // 2. Load state & agent-control config
     this.stateStore = new StateStore(path.join(this.dataDir, 'logs', 'input-state.json'));
--- a/src/types/pino-roll.d.ts
+++ b/src/types/pino-roll.d.ts
@@ -9,7 +9,7 @@ declare module 'pino-roll' {
     mkdir?: boolean;
     extension?: string;
     symlink?: boolean;
-    limit?: { count?: number };
+    limit?: { count?: number; removeOtherLogFiles?: boolean };
   }
 
   function build(options: PinoRollOptions): Promise<WriteStream>;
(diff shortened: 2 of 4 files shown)


linrunqi08 (commented):
## Code Review

Overall: **LGTM with minor notes**. The tmp-file cleanup logic is well-structured with proper error handling. No high-severity issues found.

---

### Findings

**1. [Medium] `fs-utils.ts` — EPERM/EBUSY/EACCES retry assumes error comes from `rename`**

The retry path does `await fsp.rename(tmp, path)` after 50ms. However, `fsp.writeFile(tmp, ...)` can also throw EACCES/EPERM (e.g., if the directory itself is locked). In that case, the tmp file may not exist or be incomplete, the retry `rename` would fail, and cleanup proceeds safely — so no data corruption, but the original error message could be misleading. Narrow edge case, acceptable as-is.

**2. [Low] `fs-utils.ts` — `cleanStaleTmpFiles` regex `/\.\d+\.\d+\.tmp$/` is broad**

Could match non-orphan files from other tooling whose names end with `.<digits>.<digits>.tmp`. Since this is only called on the controlled `logs/` directory, risk is minimal.

**3. [Low] `fs-utils.ts` — `cleanStaleTmpFiles` could delete an in-flight tmp from a concurrent process**

No PID-alive check before deleting. Given this runs only at startup and the daemon is a singleton, risk is low.

**4. [Low] `logger.ts` — `process.stdout.isTTY` is false in piped scenarios too**

E.g., `node app | tee out.log` or CI pipelines. Users in those contexts won't get stdout logs. Acceptable for the daemon deployment model, but worth noting.

**5. [Low] `logger.ts` — truncating `logFilePath` after pino-roll init**

Unlikely to conflict since pino-roll writes to date-suffixed files, not the base path. The `.catch(() => {})` safely handles any edge case.

---

No blocking issues. Approve after confirming the Windows `.tmp` accumulation and macOS base-file truncation work as expected in production.

ralf0131 (approved):
### Summary
LGTM — well-structured fix for log retention, tmp-file cleanup, and daemon stdout logging. Error handling in `writeJsonFile` is now comprehensive across all paths (ENOENT retry, EPERM/EBUSY/EACCES retry with 50ms backoff, and default-path tmp cleanup). `cleanStaleTmpFiles` at startup is a good defensive measure. The pino-roll `limit.removeOtherLogFiles` option and stdout-is-TTY guard correctly address the double-logging and log-accumulation issues described in the PR context.

The author's self-review already identified all the minor edge cases (rename-vs-writeFile error attribution, regex breadth, concurrent-process tmp deletion, TTY detection in piped scenarios). None are blocking. CI is green across Node 18/20/22, CodeQL, and Analyze. CLA is signed.

---

PR #89, opened 2026-06-29
feat(macos-menubar): support token consumption breakdown by model

fangxiu-wf:
## Summary

Adds a **MODELS** section to the macOS Menu Bar status panel, displaying per-model token consumption (model name, share %, compact token count, progress bar) alongside the existing AGENTS / PROVIDERS / REPOSITORIES groupings.

Tracks AGE-461. Supersedes #86 (closed for commit-authorship hygiene; identical code, only `author` metadata rewritten so both commits are attributed to me instead of the upstream developer-agent identity used during automated iteration).

## Changes (3 files, +166/-1 plus boundary-test hardening)

- `app/macos-status-bar/Sources/LoongSuitePilotMenuBarApp/PilotMetricsStore.swift`
  - New `ModelShareItem` struct + `PilotMetricsSnapshot.modelShares` field
  - New `ModelShareFile` Decodable, parsed in `buildSnapshot(...)`
  - `ProviderShareItem` / `ModelShareItem` get custom inits that clamp `share` to `[0, 1]` via `clampedShare(_:)` (NaN/inf → 0)
- `app/macos-status-bar/Sources/LoongSuitePilotMenuBarApp/PanelContentView.swift`
  - New `modelsSection` view, inserted between PROVIDERS and REPOSITORIES, reusing the PROVIDERS row template (HStack + share% + compact tokens + 4pt RoundedRectangle progress bar)
  - Extracts `progressBarWidth(share:totalWidth:)` helper
- `app/macos-status-bar/Tests/LoongSuitePilotMenuBarAppTests/MetricsSnapshotTests.swift`
  - 9 new test cases: `modelShares` decode order, missing-field default, NaN/Infinity clamping, progress-bar width bounds, malformed JSON degradation, `ModelShareItem` formatters

No changes to the JS-side `MetricsSummaryWriter` — the `modelShares` field has been written to `metrics-summary.json` since earlier; this PR only adds the Swift-side reader + UI.

## Test plan

- [x] `cd app/macos-status-bar && swift test` — **28/28 passing** (Apple Swift 6.3.3, Xcode 26.x, ~2.5s)
- [x] `npx vitest run tests/unit/status-bar/metrics-summary-writer.test.ts` — **9/9 passing** (covers `modelShares` writer-side aggregation)
[...]

PR #90, opened 2026-06-30
fix(macos-menubar): dedup clampedShare, add progress-bar width guard, rebuild binary

fangxiu-wf:
## Summary

Follow-up to #89. Three small improvements:

- **Deduplicate `clampedShare`**: Extract the identical static method from `ProviderShareItem` and `ModelShareItem` into a single file-level free function
- **Defense-in-depth on progress bar width**: Wrap the `.frame(width:)` calculation with `min(geo.size.width, ...)` so the bar can never exceed its container, even if a future code path bypasses the data-layer clamp
- **Rebuild bundled binary**: The darwin-arm64 binary shipped with #89 was stale (850504 bytes, pre-MODELS). Rebuilt from current source (914552 bytes) so the bundled binary includes the MODELS section

## Test plan

- [x] `swift test` — 28/28 passing (no test changes needed; existing clamp + progress-bar tests cover the refactor)
- [x] `nm | swift demangle` confirms `modelsSection`, `ModelShareItem`, `clampedShare` symbols present in rebuilt binary
- [x] `npx vitest run tests/unit/status-bar/metrics-summary-writer.test.ts` — 9/9 passing (JS side unaffected)


linrunqi08 (approved):
LGTM — clean refactoring, all three changes verified:

1. **`clampedShare` dedup**: the two removed `static func` methods had identical bodies; the file-level `private func` replacement is correct (Swift `private` at file scope = file-private, both structs access it from the same file). Verified no external callers exist across all 8 source files and 2 test files.

2. **Progress bar `min()` guard**: `min(geo.size.width, max(4, geo.size.width * item.share))` is a sound belt-and-suspenders layer on top of the data-layer clamp. Share is already ∈ [0,1] from `clampedShare`, so `geo.size.width * share ≤ geo.size.width` under normal conditions — the `min()` catches hypothetical bypass paths.

3. **Binary rebuild**: trusting the `nm | swift demangle` verification in the PR description.

Existing test suite (28 tests) covers the edge cases well: NaN, Infinity, share > 1, share = 0, null JSON fields, malformed JSON.

linrunqi08 (approved):

PR #93, opened 2026-07-01
fix: Windows compatibility for installer-opensource.ps1 and agent detection

linrunqi08:
## Summary
- Wrap all `& $script:NODE_BIN` calls with `$prevEAP` pattern to prevent PowerShell `NativeCommandError` when `$ErrorActionPreference = "Stop"` and node writes to stderr
- Switch PROBE_RESULT JSON passing from `process.argv[1]` to stdin pipe (`readFileSync(0,'utf-8')`) to avoid Windows command-line length/escaping issues — aligns with `installer.ps1`
- Add node directory to `$env:PATH` before `npm install` for nvm-windows setups where node is installed under a different user profile
- Check npm exit code and fail fast on install errors instead of silently continuing with broken `node_modules`
- Replace deprecated `--no-optional` with `--omit=optional`
- Use `where.exe` instead of `which` on Windows for command detection in `cli-probe.ts` and `detect-utils.ts`

## Test plan
- [x] Verified diff aligns with working `installer.ps1` patterns
- [ ] Run `installer-opensource.ps1` on Windows with nvm-windows under Administrator
- [ ] Verify agent probe detects agents (should match `installer.ps1` count)
- [ ] Verify npm install fails fast if node is not in PATH
- [ ] Verify cli-probe command detection works on Windows (where.exe) and macOS/Linux (which)

### DIFF
--- a/deploy/installer-opensource.ps1
+++ b/deploy/installer-opensource.ps1
@@ -50,6 +50,7 @@ param(
 )
 
 $ErrorActionPreference = "Stop"
+[Console]::OutputEncoding = [System.Text.Encoding]::UTF8
 
 # ============================================================
 # Constants
--- a/deploy/installer-opensource.ps1
+++ b/deploy/installer-opensource.ps1
@@ -189,7 +189,9 @@ function Check-Deps {
         exit 1
     }
 
+    $prevEAP = $ErrorActionPreference; $ErrorActionPreference = "Continue"
     $nodeMajor = & $script:NODE_BIN -e "process.stdout.write(String(process.versions.node.split('.')[0]))"
+    $ErrorActionPreference = $prevEAP
     if ([int]$nodeMajor -lt 18) {
         $nodeVer = & $script:NODE_BIN --version
         Msg "❌ 需要 Node.js >= 18，当前版本: $nodeVer" "❌ Requires Node.js >= 18, current: $nodeVer"
@@ -214,8 +216,10 @@ function Check-Deps {
         }
     }
 
+    $prevEAP = $ErrorActionPreference; $ErrorActionPreference = "Continue"
     $nodeVer = & $script:NODE_BIN --version
     $npmVer = & $script:NPM_BIN --version
+    $ErrorActionPreference = $prevEAP
     Msg "    ✅ node $nodeVer  npm $npmVer" "    ✅ node $nodeVer  npm $npmVer"
     Msg "    node pinned: $($script:NODE_BIN)" "    node pinned: $($script:NODE_BIN)"
     Write-Host ""
@@ -280,16 +284,20 @@ $script:PROBE_RESULT = "[]"
 function Probe-Agents {
     Msg "==> 探测 AI Agent..." "==> Probing AI Agents..."
     $probeScript = Join-Path $script:INSTALL_SRC "dist\cli-probe.cjs"
+    $prevEAP = $ErrorActionPreference; $ErrorActionPreference = "Continue"
     if (Test-Path $probeScript) {
         try {
-            $script:PROBE_RESULT = & $script:NODE_BIN $probeScript 2>$null
-            if (-not $script:PROBE_RESULT) { $script:PROBE_RESULT = "[]" }
+            $raw = & $script:NODE_BIN $probeScript 2>$null
+            if ($raw) {
+                $script:PROBE_RESULT = if ($raw -is [array]) { $raw -join "" } else { $raw }
+            }
         } catch {
             Msg "    ⚠️  Agent 探测失败，将跳过选择" "    ⚠️  Agent probe failed, skipping selection"
             $script:PROBE_RESULT = "[]"
         }
     }
-    $count = & $script:NODE_BIN -e "const r=JSON.parse(process.argv[1]);process.stdout.write(String(r.length))" $script:PROBE_RESULT 2>$null
+    $count = $script:PROBE_RESULT | & $script:NODE_BIN -e "const r=JSON.parse(require('fs').readFileSync(0,'utf-8'));process.stdout.write(String(r.length))" 2>$null
+    $ErrorActionPreference = $prevEAP
     if (-not $count) { $count = "0" }
     Msg "    ✅ 探测到 ${count} 个 Agent 定义" "    ✅ Found ${count} agent definitions"
     Write-Host ""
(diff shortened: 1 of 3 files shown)


deploy/installer-opensource.ps1
```diff
@@ -280,16 +284,20 @@ $script:PROBE_RESULT = "[]"
 function Probe-Agents {
     Msg "==> 探测 AI Agent..." "==> Probing AI Agents..."
     $probeScript = Join-Path $script:INSTALL_SRC "dist\cli-probe.cjs"
+    $prevEAP = $ErrorActionPreference; $ErrorActionPreference = "Continue"
```
linrunqi08, line 288: **Suggestion: wrap `$prevEAP` save/restore in `try/finally`**

The `$prevEAP = $ErrorActionPreference; $ErrorActionPreference = "Continue" … $ErrorActionPreference = $prevEAP` pattern is repeated ~12 times in this PR, all without `try/finally` protection. If a terminating .NET exception (e.g., `CommandNotFoundException`) occurs between save and restore, `$ErrorActionPreference` permanently leaks as `"Continue"` for the rest of the script, silently swallowing errors.

Concrete example in `Prompt-UserId` (~line 401): the restore is inside `try`, but the `catch {}` is empty — if `& $script:NODE_BIN` throws, the catch block runs without restoring EAP.

Consider a helper function to eliminate the repetition and guarantee the restore:
```powershell
function Invoke-NodeSafe {
    param([scriptblock]$Script)
    $prev = $ErrorActionPreference
    $ErrorActionPreference = 'Continue'
    try { & $Script } finally { $ErrorActionPreference = $prev }
}
```

Risk is low (the guarded code runs under "Continue"), but .NET-level exceptions are terminating regardless of EAP.

src/cli-probe.ts
```diff
@@ -23,8 +23,9 @@ async function findDetectionReason(detection: { paths: string[]; commands: strin
   }
   for (const cmd of detection.commands) {
     try {
+      const bin = process.platform === 'win32' ? 'where.exe' : 'which';
```
linrunqi08: **Suggestion: reuse `commandExists` from `detect-utils.ts` instead of duplicating**

This duplicates the platform-aware command lookup logic that `detect-utils.ts` also implements at line 33 (`commandExists`). Since `cli-probe.ts` already imports from that module (`import { detectAgent }`), exporting `commandExists` and reusing it here would eliminate the duplication.

Two independent copies of the same platform logic is a maintenance risk — a future fix to one (e.g., adding `Get-Command` fallback, or `where.exe /Q` quiet flag) could miss the other.

deploy/installer-opensource.ps1
```diff
@@ -280,16 +284,20 @@ $script:PROBE_RESULT = "[]"
...
+            if ($raw) {
+                $script:PROBE_RESULT = if ($raw -is [array]) { $raw -join "" } else { $raw }
+            }
         } catch {
             Msg "    ⚠️  Agent 探测失败，将跳过选择" "    ⚠️  Agent probe failed, skipping selection"
             $script:PROBE_RESULT = "[]"
         }
     }
-    $count = & $script:NODE_BIN -e "const r=JSON.parse(process.argv[1]);process.stdout.write(String(r.length))" $script:PROBE_RESULT 2>$null
+    $count = $script:PROBE_RESULT | & $script:NODE_BIN -e "const r=JSON.parse(require('fs').readFileSync(0,'utf-8'));process.stdout.write(String(r.length))" 2>$null
```
linrunqi08, line 300: **Note: stdin pipe encoding on non-UTF-8 Windows consoles**

Switching from `process.argv[1]` to stdin pipe changes the encoding path. PowerShell encodes piped strings using `[Console]::OutputEncoding`, which defaults to the OEM codepage (e.g., CP936/GBK on Chinese Windows), not UTF-8. Node reads stdin as UTF-8 via `readFileSync(0, 'utf-8')`.

If `PROBE_RESULT` ever contains non-ASCII characters (e.g., detection paths with CJK usernames like `C:\Users\张三\...`), the encoding mismatch could garble the JSON, causing silent fallback to 0 agents.

I understand this aligns with `installer.ps1` — if that works in production, this may be fine in practice. But consider adding `[Console]::OutputEncoding = [System.Text.Encoding]::UTF8` near the top of the script to ensure stdin pipes use UTF-8, especially given the target audience includes Chinese Windows environments.

linrunqi08 (commented):
## Code Review Summary

Overall this PR makes solid improvements for Windows compatibility — the `where.exe`/`which` platform check, npm exit code validation, PATH manipulation for nvm-windows, and `--omit=optional` modernization are all good changes.

A few suggestions below, mostly around defensive robustness of the `$ErrorActionPreference` save/restore pattern and code deduplication.

**3 findings** (0 blocking, 3 suggestions)

deploy/installer-opensource.ps1
linrunqi08, line 288: **建议：将 `$prevEAP` save/restore 用 `try/finally` 保护**

`$prevEAP = $ErrorActionPreference; $ErrorActionPreference = "Continue" … $ErrorActionPreference = $prevEAP` 这个模式在本 PR 中重复了约 12 次，但全部没有 `try/finally` 保护。如果在 save 和 restore 之间发生 .NET 终止性异常（如 `CommandNotFoundException`），`$ErrorActionPreference` 会永久泄露为 `"Continue"`，后续脚本中的错误会被静默吞掉。

具体例子：`Prompt-UserId` 函数（约第 401 行），restore 在 `try` 块内，但 `catch {}` 是空的 — 如果 `& $script:NODE_BIN` 抛出异常，catch 执行但不会恢复 EAP。

建议封装一个辅助函数，消除重复并确保恢复：
```powershell
function Invoke-NodeSafe {
    param([scriptblock]$Script)
    $prev = $ErrorActionPreference
    $ErrorActionPreference = 'Continue'
    try { & $Script } finally { $ErrorActionPreference = $prev }
}
```

风险较低（受保护代码运行在 "Continue" 模式下），但 .NET 级别的异常无论 EAP 设置如何都是终止性的。

src/cli-probe.ts
linrunqi08: **建议：复用 `detect-utils.ts` 中的 `commandExists`，避免重复实现**

这里的平台感知命令查找逻辑与 `detect-utils.ts` 第 33 行的 `commandExists` 函数实现完全一致。`cli-probe.ts` 已经从该模块导入了 `detectAgent`，将 `commandExists` 导出并在此复用即可消除重复。

两个独立的同一平台逻辑副本存在维护风险 — 如果将来需要修改某一处（例如添加 `Get-Command` 回退，或 `where.exe /Q` 静默标志），另一处容易遗漏。

deploy/installer-opensource.ps1
linrunqi08, line 300: **提示：stdin 管道在非 UTF-8 Windows 控制台上的编码问题**

从 `process.argv[1]` 切换到 stdin 管道改变了编码路径。PowerShell 使用 `[Console]::OutputEncoding` 编码管道字符串，该值在中文 Windows 上默认为 OEM 代码页（如 CP936/GBK），而非 UTF-8。Node 端通过 `readFileSync(0, 'utf-8')` 按 UTF-8 读取 stdin。

如果 `PROBE_RESULT` 包含非 ASCII 字符（例如检测路径中含有中文用户名 `C:\Users\张三\...`），编码不匹配可能导致 JSON 乱码，从而静默回退到 0 个 Agent。

理解这里是对齐 `installer.ps1` 的做法 — 如果生产环境运行正常，实际影响可能不大。但考虑到目标用户群包含大量中文 Windows 环境，建议在脚本头部添加 `[Console]::OutputEncoding = [System.Text.Encoding]::UTF8`，确保 stdin 管道使用 UTF-8 编码。

linrunqi08 (commented):
## 代码评审总结

整体来看，这个 PR 对 Windows 兼容性做了很好的改进 — `where.exe`/`which` 平台判断、npm 退出码校验、nvm-windows 的 PATH 处理、`--omit=optional` 替换废弃参数，这些改动都很实用。

以下是几点建议，主要涉及 `$ErrorActionPreference` save/restore 模式的健壮性和代码去重。

**3 个发现**（0 个阻塞项，3 个建议）

src/cli-probe.ts
linrunqi08: Fixed in 23c80c6 — exported `commandExists` from `detect-utils.ts` and reused it in `cli-probe.ts`, removing the duplicated logic.

deploy/installer-opensource.ps1
linrunqi08: Fixed in 23c80c6 — added `[Console]::OutputEncoding = [System.Text.Encoding]::UTF8` near the top of the script (after `$ErrorActionPreference = "Stop"`).

deploy/installer-opensource.ps1
linrunqi08: Valid suggestion. The current pattern matches `installer.ps1` (internal version), which also uses bare save/restore without `try/finally`. Since the guarded code runs under `"Continue"`, only .NET-level terminating exceptions could leak — risk is low in practice. Refactoring to `Invoke-NodeSafe` helper would be a good follow-up that applies to both scripts.

messixukejia (approved):

ralf0131 (approved):
LGTM. Comprehensive Windows compatibility fix addressing the core issues:

- **UTF-8 encoding** (`[Console]::OutputEncoding`) — fixes CJK garbled output
- **ErrorActionPreference management** — correct pattern to prevent `Stop` mode from terminating on child process stderr
- **stdin piping** for probe results — avoids Windows `MAX_CMD_LEN` (8191 chars) limitation
- **npm flag** (`--no-optional` → `--omit=optional`) — correct for npm >= 7
- **PATH management** during npm install — ensures DLL resolution
- **Cross-platform `commandExists`** — `where.exe` / `which` abstraction

All changes are well-scoped and consistently applied.

---

PR #110, opened 2026-07-07
Fix hook log collection across timezone rollover

jaymie9019:
## Summary
- Read hook JSONL data from a small rolling set of recent daily files instead of only the collector-local today file.
- Persist per-file byte offsets in input state while keeping the existing `lastFile` / `lastOffset` fields for compatibility.
- Migrate existing state by draining the previous `lastFile` tail without re-uploading old hook history.

## Why
Claude Code hook writers can inherit a different timezone from the collector process. A common setup is running Claude Code under a US timezone such as `America/Los_Angeles`, while the collector process uses the machine-local date. Around local date rollover, new Claude Code hook records may still be appended to the previous-day JSONL file from the collector's point of view.

The previous collector only read `${prefix}-${today}.jsonl`, so once the collector date moved forward it could stop reading the still-active previous-day file and miss those records.

## Validation
- `mise exec node@20.20.2 -- npm test -- tests/unit/inputs/base-hook-input.test.ts`
- `mise exec node@20.20.2 -- npm run typecheck`
- `mise exec node@20.20.2 -- npm test`
- Local live check with a timezone-lagged Claude Code hook file drained the previously unread tail after applying the collector patch.

### DIFF
--- a/src/inputs/base/base-hook-input.ts
+++ b/src/inputs/base/base-hook-input.ts
@@ -1,9 +1,14 @@
 import * as fs from 'node:fs/promises';
 import * as path from 'node:path';
 import { CollectionMethod } from '../../types/index.js';
-import type { AgentActivityEntry } from '../../types/index.js';
+import type { AgentActivityEntry, InputState } from '../../types/index.js';
 import { BaseInput, type InputOptions } from './base-input.js';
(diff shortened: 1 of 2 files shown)


jaymie9019:
Closing in favor of #111, which recreates the same patch with a GitHub-linked commit author for CLA detection.

PR #133, opened 2026-07-13
fix(metrics): reduce not-running alarm false positives

Snssn:
## Summary
- Switch collector/updater liveness checks to identity-first process matching, using PID files as a fast path and diagnostic signal.
- Debounce collector service-not-running alarms with startup grace, consecutive failure threshold, and cooldown.
- Keep updater watchdog failure semantics for command mismatch and heartbeat issues, while accepting PID changes when updater identity is found.

## Test plan
- [x] npm run typecheck --silent
- [x] npx vitest run tests/unit/utils/pid-utils.test.ts tests/unit/utils/process-discovery.test.ts tests/unit/updater/updater-metrics.test.ts tests/unit/metrics/metrics-collector.test.ts tests/unit/metrics/metrics-writer.test.ts tests/unit/core/updater-watchdog.test.ts --silent

### DIFF
--- a/src/core/updater-watchdog.ts
+++ b/src/core/updater-watchdog.ts
@@ -1,11 +1,12 @@
 import { execFile } from 'node:child_process';
-import * as fs from 'node:fs/promises';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import { promisify } from 'node:util';
 import type { AlarmManager } from '../metrics/alarm-manager.js';
 import { readJsonFile } from '../utils/fs-utils.js';
 import { createLogger } from '../utils/logger.js';
+import { checkProcessLiveness, UPDATER_PROCESS_PATTERNS } from '../utils/pid-utils.js';
+import type { ProcessLiveness } from '../utils/pid-utils.js';
 import { updaterRuntimePath, type UpdaterRuntimeState } from '../updater/runtime-state.js';
 
 const execFileAsync = promisify(execFile);
@@ -56,6 +57,7 @@ export interface UpdaterWatchdogOptions {
   sleepWakeGraceMs?: number;
   restartCooldownMs?: number;
   alarmManager?: AlarmManager;
+  updaterLiveness?: (pidFile: string) => ProcessLiveness;
 }
 
 /**
@@ -75,6 +77,7 @@ export class UpdaterWatchdog {
   private readonly sleepWakeGraceMs: number;
   private readonly restartCooldownMs: number;
   private readonly alarmManager: AlarmManager | null;
+  private readonly updaterLiveness: (pidFile: string) => ProcessLiveness;
   private timer: ReturnType<typeof setInterval> | null = null;
   private startedAt = Date.now();
   private lastTickAt = 0;
@@ -91,6 +94,8 @@ export class UpdaterWatchdog {
     this.sleepWakeGraceMs = opts.sleepWakeGraceMs ?? DEFAULT_SLEEP_WAKE_GRACE_MS;
     this.restartCooldownMs = opts.restartCooldownMs ?? DEFAULT_RESTART_COOLDOWN_MS;
     this.alarmManager = opts.alarmManager ?? null;
+    this.updaterLiveness = opts.updaterLiveness
+      ?? ((pidFile: string) => checkProcessLiveness(pidFile, UPDATER_PROCESS_PATTERNS));
   }
 
   start(): void {
@@ -148,7 +153,7 @@ export class UpdaterWatchdog {
       return this.restart('missing-heartbeat', reason);
     }
 
-    if (heartbeat.pid !== processState.pid && process.platform !== 'win32') {
+    if (processState.pid !== undefined && heartbeat.pid !== processState.pid && process.platform !== 'win32') {
       const reason = `updater heartbeat pid ${heartbeat.pid} does not match running pid ${processState.pid}`;
       if (this.inGraceWindow(now)) return { status: 'grace', reason };
       this.recordFailureAlarm(reason);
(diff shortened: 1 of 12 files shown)


ralf0131 (approved):
### Summary
Refactors process liveness checking to reduce false-positive "not-running" alarms. Introduces a centralized `checkProcessLiveness()` in `pid-utils.ts` that uses identity-first process matching (command-line patterns) with PID file as a fast path, and adds debouncing (startup grace + consecutive failure threshold + alarm cooldown) for collector health alarms in `updater-metrics.ts`.

### Findings
- **[Info]** src/utils/pid-utils.ts — The fallback to `findProcessByCommand` when the PID file is stale/missing is a solid design choice that recovers from stale PID files without false alarms.
- **[Info]** src/utils/pid-utils.ts:`isProcessAlive` — Correctly treats EPERM from `kill(pid, 0)` as alive (process exists but caller cannot signal it), which is better than the old inline code that returned false on any exception.
- **[Info]** src/core/updater-watchdog.ts — The `processState.pid !== undefined` guard before the PID mismatch check is a good fix — prevents false restarts when the liveness result has no PID.
- **[Info]** src/updater/updater-metrics.ts — The debouncing pattern (3-min startup grace, 2 consecutive failure threshold, 1-hour alarm cooldown) is well-structured and directly addresses the false-positive alarm issue.
- **[Info]** src/utils/pid-utils.ts — `findUnixProcessByCommand` correctly skips `process.pid` to avoid self-matching. The 5s timeout on `execFileSync` is reasonable for a 60s health-check interval.

### Suggestions
- Consider adding a `MAX_PROCESS_SCAN_COUNT` or limiting the `ps` output size in `findUnixProcessByCommand` for environments with very large process tables, though the 5s timeout provides a reasonable bound.

---

src/utils/pid-utils.ts
```diff
@@ -1,19 +1,206 @@
...
+export function findProcessByCommand(patterns: readonly ProcessCommandPattern[]): ProcessLiveness {
+  if (process.platform === 'win32') {
+    return findWindowsProcessByCommand(patterns);
+  }
+  return findUnixProcessByCommand(patterns);
+}
+
+export function checkProcessLiveness(pidFile: string, patterns: readonly ProcessCommandPattern[]): ProcessLiveness {
+  const pid = readPidFile(pidFile);
+  const pidFileStateWhenMissing: PidFileState = fs.existsSync(pidFile) ? 'invalid' : 'missing';
```
linrunqi08, line 73: **[Medium]** Empty command treated as "matched" — optimistic assumption may mask real command mismatch

**影响:** When `readProcessCommand(pid)` fails (returns `''`), `pidFileCommandMatched` becomes `true` due to `!pidFileCommand || isCommandMatch(...)`. A PID-reuse scenario combined with transient `ps` failure could incorrectly identify an unrelated process as the target service, defeating the identity-first guarantee.

**建议:** Treat empty command as "unknown" and fall through to process scan:
```typescript
pidFileCommandMatched = pidFileCommand ? isCommandMatch(pidFileCommand, patterns) : undefined;
if (pidFileCommandMatched === true) { return { running: true, ... } }
// undefined or false: fall through to findProcessByCommand
```

src/updater/updater-metrics.ts
```diff
@@ -50,16 +54,22 @@ export class UpdaterMetrics {
...
   private readonly collectorPidFile: string;
+  private readonly collectorLiveness: (pidFile: string) => ProcessLiveness;
   private healthTimer: ReturnType<typeof setInterval> | null = null;
   private flushTimer: ReturnType<typeof setInterval> | null = null;
   private eventQueue: UpdaterEvent[] = [];
   private alarmQueue: AlarmEntry[] = [];
   private userIdAlarmEmitted = false;
+  private readonly startedAt = Date.now();
+  private collectorConsecutiveFailures = 0;
+  private lastCollectorAlarmAt = 0;
```
linrunqi08, line 65: **[Medium]** `startedAt` is set at construction time but `start()` may be called later

**影响:** The 3-minute startup grace window (`COLLECTOR_HEALTH_STARTUP_GRACE_MS`) begins counting from object construction (`private readonly startedAt = Date.now()`). If the `UpdaterMetrics` instance is constructed during early initialization but `start()` is called after other async setup completes, the effective grace window is shortened by the delay between construction and start.

**建议:** Move `startedAt` assignment into `start()`:
```typescript
async start(): Promise<void> {
  this.startedAt = Date.now();
  await ensureDir(this.logsDir);
  ...
}
```

src/core/updater-watchdog.ts
```diff
@@ -75,6 +77,7 @@ export class UpdaterWatchdog {
   private readonly sleepWakeGraceMs: number;
   private readonly restartCooldownMs: number;
   private readonly alarmManager: AlarmManager | null;
+  private readonly updaterLiveness: (pidFile: string) => ProcessLiveness;
   private timer: ReturnType<typeof setInterval> | null = null;
```
linrunqi08, line 81: **[Medium]** `readUpdaterProcess()` changed from async I/O to synchronous `execFileSync` blocking

**影响:** The default `updaterLiveness` uses `checkProcessLiveness` which calls `execFileSync` (5s timeout for `ps`/PowerShell). Previously this method used async `execFileAsync`. Now the collector's event loop is blocked during each 60s watchdog check (typically <100ms, worst case 5s). During this time, no other I/O callbacks can execute.

**建议:** Acceptable trade-off for correctness, but worth noting as a known regression from the previous async behavior. Consider a future refactor to `checkProcessLivenessAsync` if event-loop latency becomes observable in production metrics.

linrunqi08:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 3 |
| Low      | 3 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | PASS |
| 死锁/卡死风险 | PASS |
| 状态恢复正确性 | PASS |

All timers are properly cleaned up in `stop()`, `execFileSync` calls have explicit timeouts preventing deadlock, and `collectorConsecutiveFailures` resets correctly on recovery.

### 总体结论

**不阻断合入。** 该 PR 从架构上正确实现了 identity-first 进程匹配和告警去抖动，有效解决了 `SERVICE_NOT_RUNNING_ALARM` 误报问题。主要风险点是 `readProcessCommand` 失败时的乐观匹配语义（empty command → matched），建议合入前修复以避免极端场景下的身份误判。其余 Medium 发现（`startedAt` 时机、同步阻塞）和 Low 发现可作为后续改进。

### Key Findings

1. **[Medium] Empty command = matched fallback** (`pid-utils.ts:73`): When `readProcessCommand` fails and returns `''`, the code treats it as "matched" due to `!pidFileCommand || isCommandMatch(...)`. A PID-reuse scenario combined with a transient `ps` failure could incorrectly identify an unrelated process as the target. Suggest falling through to process scan on empty command.

2. **[Medium] `startedAt` at construction vs `start()`** (`updater-metrics.ts:65`): The 3-min startup grace window begins at construction time. If `start()` is called significantly later, the effective grace is shortened.

3. **[Medium] Sync blocking in event loop** (`updater-watchdog.ts`): `checkProcessLiveness` uses `execFileSync` (up to 5s timeout), blocking the collector event loop on each 60s check. Previously async. Low practical impact but a known regression.

4. **[Low] Dropped bare `loongsuite-pilot.ps1` pattern** (`pid-utils.ts:29`): The old code matched any `loongsuite-pilot.ps1` invocation; the new regex requires `run-updater` subcommand.

### Highlights（正向实践）

- Identity-first process matching is fundamentally more robust than PID-only checks
- Excellent DI pattern for testability (`updaterLiveness`/`collectorLiveness` injection)
- Well-structured 3-layer debouncing (grace → threshold → cooldown)
- Rich `ProcessLiveness` diagnostics metadata aids debugging
- Centralized pattern constants eliminate drift between consumers

---
*评审报告详见: `code-review/pr-133/final-report.md`*

src/utils/pid-utils.ts
```diff
@@ -1,19 +1,206 @@
...
+export const COLLECTOR_PROCESS_PATTERNS: readonly ProcessCommandPattern[] = [
+  'collector-daemon.js',
+  '/bin/collector-daemon',
+  '\\bin\\collector-daemon',
+  /(?:^|\s)loongsuite-pilot(?:\.ps1)?\s+run(?:\s|$)/,
+];
+
+export const UPDATER_PROCESS_PATTERNS: readonly ProcessCommandPattern[] = [
+  'updater-daemon.js',
+  '/bin/updater-daemon',
```
linrunqi08, line 28: **[Medium]** Regex CLI patterns use `(?:^|\s)` anchor which won't match full-path invocations

**影响:** The pattern `/(?:^|\s)loongsuite-pilot(?:\.ps1)?\s+run-updater(?:\s|$)/` requires start-of-string or whitespace before `loongsuite-pilot`. A full-path invocation like `/home/user/.local/bin/loongsuite-pilot run-updater` has `/` before the binary name, not `\s`. Similarly on Windows: `C:\Users\foo\.local\bin\loongsuite-pilot.ps1 run-updater` has `\` before it.

In practice this has **low current impact** because `loongsuite-pilot.sh` uses `exec` so `ps` shows the Node.js command (matched by string patterns like `updater-daemon.js`), not the shell wrapper. But if the launch method changes or on platforms without `exec`, the regex fallback won't trigger.

**建议:** Change anchor from `(?:^|\s)` to `(?:^|[\s/\\\\])` to match path separators:
```typescript
/(?:^|[\s/\\])loongsuite-pilot(?:\.ps1)?\s+run-updater(?:\s|$)/
```

ralf0131:
📋 **Ready to merge** — approved 125h ago, CI passing, no merge conflicts.

This PR has been approved and all checks pass, but hasn't been merged yet. Pinging maintainers for follow-up.

---
*Automated notification by github-manager-bot*

ralf0131 (approved):
### Summary
Re-reviewed incremental commit `a9d4b5b2` ("treat unreadable process command as unverified identity") on top of the previously approved base.

The new commit is a focused correctness fix with two key changes:

1. **Tri-state identity verification** (): When `readProcessCommand(pid)` returns empty (e.g., permission denied, platform limitation), the old code treated it as `pidFileCommandMatched = true` — falsely reporting the process as healthy. Now it correctly returns `undefined` (unverified) and falls through to the system-wide process scan. Only `=== true` (explicitly matched) short-circuits to `running: true`. The `reason` string was simplified accordingly.

2. **Startup grace accuracy** (`updater-metrics.ts`): `startedAt` moved from construction-time (`Date.now()` in field initializer) to `start()` method. Ensures the startup grace period begins at actual service start, not at class instantiation — preventing a shortened grace window when async init delays exist.

3. **Regex robustness** (`pid-utils.ts`): Process patterns now match `/` and `\` as command boundaries (e.g., `/usr/local/bin/loongsuite-pilot run`), not just whitespace.

Tests added cover both new paths: unreadable-command falls through to scan, and reports unhealthy when scan also finds nothing. Existing watchdog test updated with `execFileSync` mock for command reading.

Looks good — re-approving.

---

src/utils/pid-utils.ts
linrunqi08: fixed — 已改为 `pidFileCommandMatched = pidFileCommand ? isCommandMatch(...) : undefined`,仅 `=== true` 返回 running,否则 fall through 到进程扫描,并新增两个覆盖测试。

src/updater/updater-metrics.ts
linrunqi08: fixed — `startedAt` 已改为 0 并在 `start()` 内赋值 `this.startedAt = Date.now()`,启动宽限窗口从 start 起算。

src/utils/pid-utils.ts
linrunqi08: fixed — COLLECTOR/UPDATER 两处正则锚已改为 `(?:^|[\s/\\])`,可匹配全路径调用的路径分隔符。

linrunqi08 (approved):

PR #151, opened 2026-07-15
fix: use opensource release package script

Monsooooon:
## Summary
- switch the release workflow package step to deploy/package-opensource.sh
- reuse the existing build output with --skip-build
- attach the Windows zip package and PowerShell installer to GitHub Releases

## Root cause
The GitHub release workflow called deploy/package.sh --opensource, but the open-source checkout only has deploy/package-opensource.sh. Tag-triggered releases would fail before creating the GitHub Release assets.

## Validation
- git diff --check
- verified deploy/package-opensource.sh supports --skip-build and produces both tar.gz and zip assets

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
       - run: npm test
       - run: npm run build
       - name: Package
-        run: bash deploy/package.sh --opensource
+        run: bash deploy/package-opensource.sh --skip-build
       - name: Create GitHub Release
         env:
           GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


ralf0131 (approved):
LGTM. Clean fix — the release workflow was referencing `deploy/package.sh --opensource` which does not exist in the open-source checkout. Switching to `deploy/package-opensource.sh --skip-build` reuses the build output from the prior `npm run build` step (no redundant rebuild), and adding the Windows zip + PowerShell installer (`installer.ps1`) to release assets improves cross-platform distribution. Well-scoped, minimal change.

---

PR #164, opened 2026-07-20
Fix sqlite3 installation under npm 12

Monsooooon:
## Summary

- approve the audited `sqlite3@5.1.7` install script through npm 12's `allowScripts` policy
- load `sqlite3` from the staged candidate with the updater's Node.js runtime before changing version pointers
- keep `current` and `previous` unchanged and skip the collector restart when the native binding cannot load
- add an integration regression test that covers the failure path on Unix and Windows command forms

## Root cause

npm 12 can skip unreviewed dependency lifecycle scripts while `npm install` still exits successfully. `sqlite3@5.1.7` needs its install script to create or download `node_sqlite3.node`, so the updater could previously activate an unusable candidate and send the collector into a crash loop.

## Impact

The package explicitly allows the reviewed sqlite3 script, and the updater now treats a missing or incompatible native binding as a deployment failure before activation.

## Validation

- `npm run typecheck`
- `AGENT_DATA_COLLECTION_CONFIG=/tmp/... npm test -- --run tests/unit/updater/updater.test.ts tests/integration/updater-flow.test.ts`
- 46 updater unit/integration tests passed
- the equivalent packaged artifact was verified with Node.js 22.22.2 and npm 12.0.1: installation exited 0, `node_sqlite3.node` was created, and `require('sqlite3')` succeeded

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,9 @@
   "optionalDependencies": {
     "@rollup/rollup-darwin-arm64": "^4.60.2"
   },
+  "allowScripts": {
+    "sqlite3@5.1.7": true
+  },
   "files": [
     "dist/",
     "assets/",
--- a/src/updater/updater.ts
+++ b/src/updater/updater.ts
@@ -514,6 +514,13 @@ export class Updater {
         shell: process.platform === 'win32',
       });
 
+      logger.info('checking sqlite3 runtime');
+      await execFileAsync(process.execPath, ['-e', "require('sqlite3')"], {
+        cwd: stagingDir,
+        env: childEnv,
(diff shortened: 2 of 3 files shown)


ralf0131 (approved):
### Summary
LGTM. Solid fix for the npm 12 `sqlite3` lifecycle-script gap — combines prevention (`allowScripts` whitelist for the audited `sqlite3@5.1.7`) with defense-in-depth (runtime `require('sqlite3')` health check before activation). The failure path is correctly tested: when the native binding cannot load, `current`/`previous` pointers stay unchanged and the collector is not restarted.

---

PR #174, opened 2026-07-21
fix(qoderwork): load host app's own worker runtime, never a foreign one

fangxiu-wf:
## Background

Token interception for QoderWork is injected by setting `QODER_WORKER_RUNTIME_PATH` to `qoderwork-runtime-wrapper.mjs`. On macOS this is done with `launchctl setenv`, which sets the variable in the **launchd user domain — global to every GUI app**. The entire `@qoder-ai/qoder-agent-sdk` family (QoderWork, QwenWorkCN, QoderWork CN, …) honours the *same* variable name and uses it as its worker entry.

## The bug

The old wrapper hardcoded `QoderWork.app` in `RUNTIME_CANDIDATES`. So when a **sibling** app (e.g. QwenWorkCN) starts its worker, it inherits the global env, loads our wrapper, and the wrapper imports **QoderWork's** runtime into the sibling's worker — a cross-version / cross-site (`global` vs `cn`) mismatch that breaks the sibling app (the reported "千问办公挂掉 / two CLIs get confused").

Verified on a machine with both apps installed:
- `launchctl getenv QODER_WORKER_RUNTIME_PATH` → the wrapper (and the var even leaks into unrelated Electron apps like Cursor/Wukong).
- QwenWorkCN bundles its own `qoder-worker-runtime.obf.mjs` (v1.0.46 / `cn`) and reads the same env var; QoderWork's is v1.0.41 / `global` (different md5).

## The fix

Make the wrapper a transparent, **app-agnostic** shim:
- Locate the **host app's own** runtime purely from the running process — the first `.app` in `process.execPath`, then `process.resourcesPath` — resolving to whichever app is hosting the worker. No app name is hardcoded.
- Install the JSON.parse/stringify interception only right before importing that runtime.
- **Remove the hardcoded candidates entirely.** If the host app's own runtime cannot be located with certainty, install nothing and load nothing.

Design priority: **never break the host app.** Losing token interception is acceptable; loading a foreign runtime is not.

`workerData.qoderWorkerRuntime.runtimeRoot` / `QODER_WORKER_RUNTIME_ASSET_ROOT` are intentionally not used — the SDK materialises the wrapper, so those point at the materialised wrapper copy, not the app.

## Test plan

- [x] Static resolution validated against real bundles: QwenWorkCN main binary, QwenWorkCN nested `*Helper.app`, QoderWork main binary → each resolves to its OWN runtime; nested Helper correctly falls back to the outer bundle; plain `node` (no clue) → not found → safe no-op.
- [x] `node --check` passes.
- [x] Live: after deploying, QoderWork ran a worker and token collection still works end-to-end (fresh intercept record → `gen_ai.usage` in the final output), with no `qoderwork-wrapper-error.log` entries — confirming self-location works in the real Electron worker thread.
- [ ] QwenWorkCN end-to-end once its beta allows a conversation (root cause is structurally removed regardless).

## How to verify locally (before release)

No need to wait for a release — this can be grey-tested by hand on any machine that can open QwenWorkCN:

1. Replace the **live** wrapper (the one the env points at), not the packaged copy:
   `~/.loongsuite-pilot/hooks/qoderwork-runtime-wrapper.mjs`
2. Confirm the env is actually pointing at it: `launchctl getenv QODER_WORKER_RUNTIME_PATH` should print that path. (It is normally set because QoderWork.app is installed; if empty, the app never loads the wrapper and there is nothing to test.)
3. **Fully quit** QwenWorkCN (Cmd+Q, not just close the window) and reopen it, then have a real conversation. The worker only reads the env / loads its runtime at startup; a content change is re-materialised by the SDK, so there is no stale-cache concern.
4. Pass criteria: QwenWorkCN works normally **and** `~/.loongsuite-pilot/logs/qoderwork-wrapper-error.log` gets no new `host app runtime not found` / `host runtime import failed` lines for QwenWorkCN.

This is the same procedure already confirmed for QoderWork above, just exercised from QwenWorkCN.

## How the fix reaches users (after release)

Once merged and released, the fix propagates through the normal install/update path — with one caveat:

- Both the installer (`deploy/installer-opensource.sh` → `scripts/postinstall.js`) and the auto-updater (`src/updater/updater.ts` runs the staged `postinstall.js`) copy `assets/hooks/` → `~/.loongsuite-pilot/hooks/` via `fs.cpSync(..., { recursive: true })`. Node's `cpSync` overwrites by default, so the live `qoderwork-runtime-wrapper.mjs` is replaced with the new version (no "skip if exists").
- The env path (`~/.loongsuite-pilot/hooks/qoderwork-runtime-wrapper.mjs`) is stable, so it keeps pointing at the now-updated file, and the SDK re-materialises on the content change.
- **Caveat:** reinstalling/updating pilot does **not** restart the GUI apps, and a worker only picks up the new wrapper when its host app starts. So the effective statement is: *reinstall or auto-update **+** fully quit & reopen the affected app → fixed.* Already-running workers are not hot-patched.

### DIFF
--- a/assets/hooks/qoderwork-runtime-wrapper.mjs
+++ b/assets/hooks/qoderwork-runtime-wrapper.mjs
@@ -1,108 +1,170 @@
-// QoderWork worker runtime wrapper — intercepts token data via JSON.parse hook.
-// Loaded via: QODER_WORKER_RUNTIME_PATH=<this-file>
+// QoderWork-family worker runtime wrapper — transparent, app-agnostic shim.
 //
-// QoderWork runs its agent SDK in a Node.js worker_thread (not Bun), so the
-// qodercli BUN_OPTIONS --preload trick does not apply. The SDK honors
-// QODER_WORKER_RUNTIME_PATH as the worker entry, so we wrap it: install a
-// JSON.parse/JSON.stringify hook, then `await import()` the real runtime.
+// Loaded via the SHARED env var QODER_WORKER_RUNTIME_PATH. The ENTIRE
+// @qoder-ai/qoder-agent-sdk family honours this variable (QoderWork,
+// QwenWorkCN, QoderWork CN, ...), and on macOS we set it with `launchctl
+// setenv`, which is GLOBAL to the launchd user domain. Consequences:
+//   • Every GUI app inherits the variable, but only apps that actually run the
+//     @qoder-ai SDK ever load this file as their worker entry.
+//   • Therefore this wrapper CAN be the worker entry of ANY sibling app, not
+//     just QoderWork. It MUST NOT assume which app loaded it.
 //
-// Writes to: ~/.loongsuite-pilot/logs/qoderwork-intercept.jsonl
-// The captured `id` (chatcmpl-xxx) matches the hook processor's
-// gen_ai.response.id (derived from transcript message.id), enabling direct
-// token matching in qoder-work-trace-input without a transcript mapping module.
+// Design priority (do NOT weaken): NEVER break the host app. We only ever hand
+// control to the *host app's OWN* bundled runtime, located dynamically from the
+// running process. There is intentionally NO hardcoded/app-specific fallback:
+// loading a foreign runtime (e.g. QoderWork's runtime inside QwenWorkCN) is
+// exactly what corrupts the app. If we cannot locate the host app's own runtime
+// with certainty, we install nothing and load nothing — token interception is
+// sacrificed, the app is never handed a wrong runtime.
+//
+// On the success path only, token/system-prompt records are appended to
+// ~/.loongsuite-pilot/logs/qoderwork-intercept.jsonl.
 
 import { createRequire } from 'module';
+import { fileURLToPath } from 'url';
 const require = createRequire(import.meta.url);
 const fs = require('node:fs');
 const path = require('node:path');
 
 const INTERCEPT_DIR = path.join(process.env.HOME || '/tmp', '.loongsuite-pilot', 'logs');
 const INTERCEPT_FILE = path.join(INTERCEPT_DIR, 'qoderwork-intercept.jsonl');
+const ERROR_LOG = path.join(INTERCEPT_DIR, 'qoderwork-wrapper-error.log');
 const MIN_SYSTEM_PROMPT_LENGTH = 100;
-
-try { fs.mkdirSync(INTERCEPT_DIR, { recursive: true }); } catch {}
 
 const origParse = JSON.parse;
 const origStringify = JSON.stringify;
 let lastId = null;
 let systemPromptCaptured = false;
 
-// Global override of JSON.parse to intercept SSE-parsed token usage.
-JSON.parse = function (text, reviver) {
-  const result = origParse.call(JSON, text, reviver);
+function logDiag(msg) {
   try {
-    if (result && typeof result === "object"
-        && result.usage && result.choices !== undefined
-        && result.id !== lastId) {
-      lastId = result.id;
-      const u = result.usage;
-      const rec = {
-        type: "token",
-        ts: Date.now(),
-        id: result.id,  // chatcmpl-xxx, matches transcript message.id
-        model: result.model || "",
-        prompt_tokens: u.prompt_tokens || 0,
-        cached_tokens: (u.prompt_tokens_details && u.prompt_tokens_details.cached_tokens) || 0,
-        completion_tokens: u.completion_tokens || 0,
-        reasoning_tokens: (u.completion_tokens_details && u.completion_tokens_details.reasoning_tokens) || 0,
-        total_tokens: u.total_tokens || 0,
-      };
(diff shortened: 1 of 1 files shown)


linrunqi08:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 0 |
| Low      | 3 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | PASS |
| 死锁/卡死风险 | PASS |
| 状态恢复正确性 | PASS |

### 总体结论

**未发现阻断合入的问题，建议合入。** 本 PR 将 wrapper 从“硬编码 `QoderWork.app` 候选”改为“仅从运行进程（`process.execPath` 首个 `.app` + `process.resourcesPath`）定位并加载宿主应用自有运行时，否则 no-op”，从结构上消除了跨应用运行时污染（QoderWork 运行时被导入 QwenWorkCN）的根因，并严守“绝不弄坏宿主应用”的设计红线。CI 全绿、`node --check` 通过、作者已用真实 bundle 静态验证 + QoderWork live 端到端验证。

以下为 3 条 Low 级残余风险/提示（均不阻断）：

- **[Low] `.app` 定位假设**（L115）：非贪婪取 `execPath` 首个 `.app`，若祖先目录名以 `.app` 结尾会误解析 root(1)，仅 `process.resourcesPath` 存在时才兜底；worker_thread 中 `resourcesPath` 不保证存在，此时退化为 no-op（**丢采集但不加载外来运行时、不破坏宿主**，与设计红线一致）。极小概率。
- **[Low] 采集覆盖收敛**（L146）：完全依赖运行进程派生 root；在“`execPath` 无 `.app` 且 `resourcesPath` 缺失”的异常启动形态下，QoderWork 自身也可能退化 no-op（丢采集）。打包 Electron 下 `execPath` 必含 `.app`，实际风险很低。
- **[Low] `mkdirSync` 冗余**（L50 与 L41）：`logDiag` 与 `installInterceptHooks` 各调用一次，幂等无副作用，仅整洁度提示。

### Highlights（正向实践）

- 设计红线清晰且贯穿实现：定位失败即 no-op，宁失采集不加载外来运行时，结构性消除根因。
- 防自引用（`real === selfPath`）、`seen` 去重、`realpathSync` 归一，边界处理到位。
- 注释充分解释了 `launchctl setenv` 全局性、首个 `.app` 匹配、不使用 `workerData.runtimeRoot` 等关键决策。
- JSON 钩子安装从“模块顶层无条件”收敛为“仅命中宿主运行时前”，降低对未命中 worker 的副作用。

---
*评审报告详见: `code-review/pr-174/final-report.md`*

linrunqi08 (approved):

ralf0131 (approved):
﻿### Summary

This PR fixes a critical cross-app runtime contamination bug in the QoderWork worker runtime wrapper. The old code hardcoded `/Applications/QoderWork.app` paths, which caused sibling apps (e.g. QwenWorkCN) sharing the same `QODER_WORKER_RUNTIME_PATH` env var to load the wrong runtime — breaking the sibling app entirely.

The fix replaces hardcoded paths with dynamic process-based discovery (`process.execPath` → first `.app` in path → resolve runtime relative to that bundle). The design priority — "never break the host app" — is correctly enforced: if the host app's own runtime cannot be located, the wrapper installs nothing rather than guessing.

### Review

- **Correctness** ✅ — Dynamic discovery via `process.execPath` is sound for Electron worker threads. The anti-recursion check (comparing resolved wrapper path against candidate runtime paths) correctly prevents self-import. The `\.app/.*?\.app` regex properly handles nested Helper.app bundles by matching the outermost `.app`. Error handling covers all failure modes (no `.app` in path, runtime not found, import failure) without crashing the worker.
- **Performance** ✅ — All new operations (`realpathSync`, regex match, `readdirSync`) are one-time startup costs. No runtime overhead on the hot path (JSON.parse/stringify hooks).
- **Compatibility** ✅ — Fully backward compatible. The wrapper still intercepts tokens via the same JSON hooks and writes to the same file. The env var contract (`QODER_WORKER_RUNTIME_PATH`) is unchanged.
- **Tests** ⚠️ — Manual test plan is thorough (real bundle validation, `node --check`, live deployment verification). However, the new `candidateResourceRoots()` function has multiple code paths (execPath resolution, Helper.app fallback, readdir fallback) that would benefit from unit tests to prevent regression.

### Suggestions

1. **Consider adding unit tests for `candidateResourceRoots()`** — The function has 3 distinct resolution strategies with different fallback chains. Mock `process.execPath` and `process.resourcesPath` to verify each path resolves correctly. This would catch regressions if the Electron runtime changes how it reports these values.

2. **`ERROR_LOG` write in catch block** (line 168) — Consider adding a timestamp prefix to error log entries for easier debugging: `new Date().toISOString() + ' | ' + err.stack || err`.

Overall, this is a well-designed fix with excellent documentation of the root cause and design constraints. The "fail safe, never break the host" principle is correctly implemented. LGTM.

---

PR #187, opened 2026-07-25
feat(workbuddy): add transcript collection and quality gates

yyuuttaaoo:
## Summary

This PR adds WorkBuddy Desktop collection to Pilot and turns the existing JSONL validator into an enforceable onboarding quality gate.

- Adds automatic WorkBuddy discovery, hook deployment, transcript collection, normalization, and uninstall cleanup.
- Keeps Pilot's existing wire contract: `llm.request`, `llm.response`, `tool.call`, and `tool.result`; tool duration remains milliseconds.
- Uses WorkBuddy hooks only as fail-open wake-up/path hints. The transcript is the authoritative audit source; Pilot does not read WorkBuddy SQLite.
- Preserves native JSON values in JSONL output and strengthens event topology, identity, type, privacy, and completeness validation.
- Expands the existing English and Chinese onboarding/schema documentation instead of introducing a separate quality spec or validator mode.

### WorkBuddy support

- Added the `workbuddy` agent declaration and automatic discovery/configuration.
- Added cross-platform hook wrappers and a low-sensitive hook journal that records only structural identifiers and transcript path hints.
- Added a checkpointed transcript input with file watching, polling fallback, cold-start baselining, and restart-safe replay behavior.
- Added deterministic audit event construction for:
  - text-only turns;
  - single and parallel tool waves;
  - tool result correlation;
  - final responses;
  - token/cache usage;
  - turn/step boundaries;
  - workspace and Git context;
  - content collection controls.
- Added stable `event.id` generation and omitted source facts that are unavailable rather than inventing `unknown`, fake tokens, fake status, or zero duration.
- Added installer and uninstaller handling for the WorkBuddy hooks and journal.

### Shared quality gate improvements

- Extended the existing agent onboarding checklists in English and Chinese with required/conditional field coverage, topology, replay, privacy, fixture, and real-agent acceptance requirements.
- Enhanced the existing JSONL validator to:
  - accept only Pilot's existing event names;
  - report field coverage grouped by event;
  - validate required and conditionally required fields;
  - validate native types, nanosecond timestamps, trace/span identifiers, tokens, and positive millisecond durations;
  - detect duplicate event IDs;
  - validate request/response, tool call/result, and turn-boundary topology;
  - report recommended-field gaps and suspicious placeholders without exposing field values.
- Kept `E2E_JSONL_STRICT` as the sole failure-control switch and enabled it by default for the formal L1 install smoke test.
- Added one synthetic high-coverage WorkBuddy fixture plus table-driven validator failures. No real transcript, personal path, prompt, tool argument, or tool result is committed.

## Implementation overview

```text
WorkBuddy hook
  -> structural wake-up journal (no content)
  -> WorkBuddyInput watches/scans referenced transcripts
  -> transcript assembler merges text and parallel tool waves
  -> WorkBuddy event builder applies content policy and stable identity
  -> Pilot MultiFlusher emits the existing wire protocol
```

The hook is deliberately not an audit source: it avoids duplicating partial events and sensitive payloads. Hook journal and transcript file changes trigger immediate collection; a 30-second local transcript poll provides a fallback consistent with other Pilot listeners. The transcript provides ordering, model responses, tool calls/results, usage, and workspace context. File checkpoints provide incremental processing and restart safety. SQLite is not queried because the transcript is sufficient for the emitted contract and avoids a third source of truth.

For events where WorkBuddy records a call and result at the same timestamp, duration is omitted instead of emitting a fabricated `0 ms`.

### Automated checks

- `npm run typecheck`: passed
- `npm test`: 181 test files passed; 2,004 tests passed; 2 skipped
- `npm run build`: passed, including the macOS status-bar application
- `git diff --cached --check`: passed

### Installed-product acceptance

The built package was installed locally and exercised through the WorkBuddy GUI with two read-only tool calls. Only rows appended after the pre-test line baseline were analyzed:

- JSONL baseline: 58 rows
- JSONL after interaction: 66 rows
- New rows: 8
- Event counts: 2 `llm.request`, 2 `llm.response`, 2 `tool.call`, 2 `tool.result`
- Enhanced strict validator: passed with zero parse, protocol, required-field, type, duplicate-ID, pairing, turn-boundary, or suspicious-placeholder errors

### Observed event field coverage

| Event | Count | Required envelope | Trace/span | Session/turn/step | Git/workspace |
|---|---:|---:|---:|---:|---:|
| `llm.request` | 2 | 2/2 (100%) | 2/2 (100%) | 2/2 (100%) | 2/2 (100%) |
| `llm.response` | 2 | 2/2 (100%) | 2/2 (100%) | 2/2 (100%) | 2/2 (100%) |
| `tool.call` | 2 | 2/2 (100%) | 2/2 (100%) | 2/2 (100%) | 2/2 (100%) |
| `tool.result` | 2 | 2/2 (100%) | 2/2 (100%) | 2/2 (100%) | 2/2 (100%) |

The required envelope covers `time_unix_nano`, `event.id`, `event.name`, user, session, agent, and provider identity.

| Event | Event-specific fields | Coverage |
|---|---|---:|
| `llm.request` | request ID/model and input message delta | 2/2 (100%) |
| `llm.response` | response ID/model/finish reason/output and input/output/total tokens | 2/2 (100%) |
| `tool.call` | tool name, call ID, and native arguments | 2/2 (100%) |
| `tool.result` | tool name, matching call ID, native result, and status | 2/2 (100%) |

Conditional boundary fields also matched the observed topology:

| Conditional field | Coverage | Interpretation |
|---|---:|---|
| `turn.start` on first request | 1/2 (50%) | Present only on the first step of the turn |
| `turn.end` on final response | 1/2 (50%) | Present only on the final step of the turn |
| `gen_ai.tool.call.duration` | N/A (0/2) | Source call/result timestamps were equal, so no fake zero was emitted |

Both request/response pairs and both tool call/result pairs were one-to-one; tool names and call IDs matched. JSON arguments/results remained native objects rather than serialized strings.

## Scope note

The exploratory WorkBuddy-specific spec is intentionally excluded. This repository already treats the onboarding and output-schema documents as the maintained public contract, so this PR updates those locations directly.


ralf0131:
⚠️ **Merge conflict detected**

This PR has conflicts with the `main` branch and cannot be merged. Please rebase or merge `main` into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout feat/workbuddy
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

This is a one-time reminder. Feel free to @mention me for a re-review after conflicts are resolved.

---
*Automated notification by github-manager-bot*

fangxiu-wf:
Thank you again for the substantial implementation and for addressing the earlier feedback. I performed another complete review of the current head (`a0073b9c06f4a8af1f1c1ae1f92b2120a3b30020`), including an installed-product E2E test on a real Windows environment rather than relying only on fixtures or unit tests.

## Windows installed-product E2E validation

I built and installed the PR merge candidate on **Windows 11** and exercised **WorkBuddy Desktop 5.3.5.0 / Electron 37.10.3** through its real GUI. The test asked WorkBuddy to:

1. call `Write` to create a file with a fixed payload;
2. call `Read` to read it back and verify the content;
3. return a fixed completion marker.

I then queried the exported trace from the backend by trace ID. The resulting trace contained 10 spans with the expected hierarchy:

```text
ENTRY  enter_ai_application_system
└── AGENT  invoke_agent workbuddy
    ├── STEP  react step
    │   ├── LLM   chat auto  [finish=tool_call]
    │   └── TOOL  execute_tool Write
    ├── STEP  react step
    │   ├── LLM   chat auto  [finish=tool_call]
    │   └── TOOL  execute_tool Read
    └── STEP  react step
        └── LLM   chat auto  [finish=stop]
```

The following all passed:

- all spans shared the same trace ID and every non-root parent existed;
- both `Write` and `Read` spans contained tool name, call ID, native arguments, result, and successful status;
- message capture worked when temporarily enabled and was disabled again after the test;
- the three LLM spans reported input `108641`, output `332`, and total `108973` tokens, and the AGENT aggregation matched exactly;
- there were no OTLP export failures and no credential-like values were detected in the inspected server-side data.

This confirms that the main WorkBuddy transcript → Pilot events → OTLP → backend path works end to end on Windows, including a real multi-step tool-call flow. However, the successful trace primarily covers the normal path and polling fallback. It does not invalidate the issues below; in fact, the server-side trace exposed two of them directly.

## Remaining findings

### [P1] A temporarily unreadable transcript clears its checkpoint and replays all history

`WorkBuddyInput.collect()` creates `nextOffsets` as an empty map and replaces the persisted map with it. If project discovery or a transcript read temporarily fails, the old checkpoint disappears. Once the path becomes readable again, `initialized=true` plus the missing offset selects baseline `0` and rebuilds the entire transcript.

I reproduced this against the current head: after checkpointing the 9-line fixture, I temporarily made the projects directory unavailable. The stored offset map became empty, and restoring the directory emitted all 8 historical events again.

Please preserve existing offsets for paths that could not be scanned/read, and only prune an offset after a separately confirmed permanent deletion. A regression test should cover temporary directory/file unavailability and recovery.

### [P1] Fallback IDs do not correlate `tool.call` with `tool.result` when `callId` is absent

The call fallback is generated from `call:${sourceId(call)}:${index}`, while the result fallback uses `call:${sourceId(result)}`. These are different seeds. Removing the native call IDs from both sides of the current fixture produced different IDs for every call/result pair.

The existing test only removes the ID from one call and compares the response tool part with `tool.call`; it never asserts that the corresponding `tool.result` receives the same generated ID. Please derive the fallback once through a shared correlation map and add a parallel-tool test where both call and result IDs are missing.

### [P1] Every WorkBuddy LLM span currently has zero duration

For both tool waves and final responses, the builder creates `llm.request` and `llm.response` from the same response record, so both events have the same `time_unix_nano`.

This was directly visible in the Windows E2E trace: all 3 LLM spans had duration `0`, the final STEP also had duration `0`, and the first two STEP durations were exactly equal to their TOOL durations. This makes LLM latency and the trace waterfall inaccurate.

Please use the user message / previous tool-result boundary as the request timestamp and the assistant/function-call record as the response timestamp, then add a Builder → OTLP conversion test asserting positive and properly nested LLM/STEP durations.

### [P2] Missing tool names are accepted by the builder but rejected by the strict quality gate

The builder and its unit test intentionally omit `gen_ai.tool.name` when unavailable. The new strict validator and output schema require it for both `tool.call` and `tool.result`.

I passed the builder's missing-name output through JSONL and the current strict validator; it exited with status 1 and reported 6 issues. Please first copy a name from the paired call/result when one side has it. If neither side has a name, the contract must explicitly choose whether to skip the incomplete tool events or permit omission. Please cover the decision with a Builder → JSONL → strict-validator integration test.

### [P2] `tool_call` still conflicts with the documented `tool_calls` finish reason

WorkBuddy emits `tool_call`, while the output schema documents only `tool_calls`. The strict validator checks only that `gen_ai.response.finish_reasons` is an array, not that each element is a valid documented value, so the mismatch passes strict validation and also reached the real Windows trace.

The earlier Copilot thread was marked resolved/outdated after the code moved, but the semantic mismatch remains at the current head. This is also inconsistent across existing adapters, so please select one canonical repository contract and align the schema, builders, and validator enum checks.

### [P2] The Windows hook ignores the Node path pinned by the installer

`workbuddy-loongsuite-pilot-hook.ps1` only calls `Get-Command node`. The Windows installer records its selected executable in `~/.loongsuite-pilot/node-bin`, and the other Windows hooks resolve that pinned path before falling back to PATH.

In this E2E setup, portable Node was added only to the installer process environment. An already-running WorkBuddy GUI does not inherit that temporary PATH. Therefore the successful trace proves the polling fallback, but does not prove that the immediate hook wakeup ran. Please reuse the existing Windows `Resolve-NodeBin` behavior and add a wrapper test where PATH has no Node but `node-bin` is present.

### [P2] The actual DeepSeek model is reported with `provider=workbuddy`

The real trace reported:

```text
gen_ai.request.model=auto
gen_ai.response.model=deepseek-v4-pro
gen_ai.provider.name=workbuddy
```

`auto` is a valid requested routing model, but `gen_ai.provider.name` is defined by the output schema as the model provider. WorkBuddy is already represented by the agent type/system. `inferProvider()` currently lacks DeepSeek and Qwen mappings and falls back to `workbuddy`, which makes provider aggregation inaccurate. Please prefer an explicit upstream provider when available and cover the known model families, including DeepSeek and Qwen.

### [P2] The input filters old output but still rereads every full transcript every 30 seconds

Each collection cycle scans every project directory, uses `readFile()` for every transcript, parses all historical lines, and rebuilds the full session before `minTerminalIndex` suppresses old output. This does not satisfy the new onboarding requirement to incrementally read only new records, and CPU, memory, and disk work will grow with accumulated history.

Please use byte/inode checkpoints and retain only the minimal pending turn/tool-wave state required to process appended records.

### [P2] Strict validation succeeds when the selected time window contains zero events

If matching JSONL files exist but every row is excluded by `E2E_JSONL_SINCE_SECONDS`, the validator reports `windowed=0` and exits 0 even in strict mode. I reproduced this with the current validator:

```text
status=0
files=1 entries=1 windowed=0
```

This permits an E2E run that collected no new events to pass. Strict mode should fail when a requested time window contains no rows, ideally together with explicit required-agent/event coverage.

### [P2] Windows working directories cannot be used for Git enrichment

The WorkBuddy builder passes `C:\...` cwd values to the shared Git enrichment helper, but `extractProbeDir()` accepts only paths beginning with `/`. Consequently, Windows events retain `workspace.path`, but a Windows Git workspace cannot populate `git.repo`, `git.branch`, `git.domain`, or `workspace.current_root`.

Please use a cross-platform absolute-path check and add a Windows path test.

## Documentation and current PR state

- The WorkBuddy agent documentation still says only macOS WorkBuddy 5.2.6 was verified. After the Node wakeup fix is exercised, please record the Windows 11 / WorkBuddy 5.3.5 validation as well.
- Since WorkBuddy is intentionally outside the default headless L1 filter, the GUI acceptance instructions should explicitly use `E2E_JSONL_AGENT_FILTER=workbuddy`.
- The PR is currently `CONFLICTING/DIRTY` against the latest `main`, with a conflict in `deploy/installer-opensource.ps1`. The resolution should preserve current main's separate Codex cleanup and add WorkBuddy to the general hook-config cleanup list.
- The Windows E2E candidate was based on an earlier merge result, not the current `main`, so the relevant validation should be rerun after rebasing.

The relevant WorkBuddy, hook, flusher, deployment, and validator tests pass, typecheck passes, and the Node 18/20/22 CI checks are green. Those results and the Windows E2E demonstrate that the implementation is close and that the main path is working. Given the replay/correlation defects and the systematic zero-duration trace data, however, I do not think the PR is ready to merge yet. Please address the P1 findings and the contract/Windows correctness items above, rebase, and rerun the installed Windows E2E.

src/inputs/workbuddy/workbuddy-input.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  private async scanTranscriptFiles(): Promise<string[]> {
+    let projectDirs: string[];
+    try { projectDirs = await fs.readdir(this.projectsDir); } catch { return []; }
+    const files: string[] = [];
+    for (const projectDir of projectDirs) {
+      const fullDir = path.join(this.projectsDir, projectDir);
+      let names: string[];
+      try { names = await fs.readdir(fullDir); } catch { continue; }
```
ralf0131: **[Warning]** `readCompleteJsonl` reads the entire transcript file into memory on every `collect()` cycle. For long WorkBuddy sessions with thousands of transcript records, this could become a memory concern. Consider streaming JSONL parsing or at least tracking the byte offset (not just record count) to skip already-processed portions more efficiently.

src/inputs/workbuddy/workbuddy-input.ts
```diff
@@ -0,0 +1,200 @@
...
+    const entries: AgentActivityEntry[] = [];
+
+    for (const transcriptPath of transcriptPaths) {
+      const parsed = await readCompleteJsonl(transcriptPath);
+      if (!parsed) continue;
+      const sessionId = path.basename(transcriptPath, '.jsonl');
+      const prior = offsets[transcriptPath];
+      const baseline = !initialized && prior === undefined ? parsed.records.length : prior ?? 0;
+      const built = await buildWorkBuddyEvents(parsed.records, {
+        sessionId,
```
ralf0131: **[Info]** The dual strategy of `fs.watch` + polling fallback is a solid approach for reliability. The `watchDirectory` error handling (catching and logging watch failures without crashing) is well done. One suggestion: consider adding a max-retry or backoff for the polling fallback when the watched directory doesn't exist yet (cold start scenario).

src/inputs/workbuddy/workbuddy-event-builder.ts
```diff
@@ -0,0 +1,450 @@
+import { createHash } from 'node:crypto';
```
ralf0131, line 1: **[Info]** The event builder's approach of omitting unavailable source facts rather than inventing `unknown`/fake values is excellent for data integrity. The stable `event.id` generation and the deterministic audit event construction for text-only turns, tool waves, and turn boundaries are well-structured.

ralf0131 (commented):
### Summary
This PR adds WorkBuddy Desktop collection to Pilot with automatic discovery, hook deployment, transcript collection, normalization, and uninstall cleanup. It also strengthens the JSONL validator into an enforceable quality gate with topology validation, duplicate event ID detection, and strict field coverage checks. The implementation is well-structured with clean separation of concerns.

**⚠️ This PR currently has merge conflicts that need to be resolved before it can be merged.**

### Findings
- **[Warning]** `workbuddy-input.ts:112` — `readCompleteJsonl` reads entire transcript into memory; could be a concern for long sessions
- **[Info]** `workbuddy-input.ts:78` — Dual fs.watch + polling fallback is solid; consider backoff for cold-start polling
- **[Info]** `workbuddy-event-builder.ts:1` — Excellent data integrity approach: omit unavailable facts rather than inventing placeholders
- **[Warning]** `workbuddy-event-builder.ts:87` — `pendingDelta` reset after tool wave may leave subsequent `llm.request` with empty input delta; verify this is intentional
- **[Info]** `workbuddy-hook-journal.mjs:1` — Privacy-conscious hook journal design, only structural identifiers recorded
- **[Info]** `agents.d/workbuddy.json:16` — `.*` event matchers are correct as default; users can narrow via settings override

### Suggestions
1. **Merge conflicts**: Resolve conflicts with `main` branch before merge. The `deployment-manager.ts` and `hook-strategy.ts` files appear to have conflicting changes.
2. **Memory efficiency**: For long WorkBuddy sessions, consider tracking byte offset alongside record count to enable incremental JSONL reading rather than re-parsing the entire file.
3. **Cold-start polling**: When the watched directories don't exist yet (fresh install), the polling fallback could benefit from exponential backoff to avoid unnecessary filesystem checks.
4. **Documentation**: The onboarding docs (English + Chinese) are well-updated. The quality gate requirements are clearly documented.

### Overall
Well-designed addition that follows Pilot's established patterns. The privacy-first hook journal, checkpointed transcript reading, and comprehensive quality gate are strong points. The merge conflicts are the only blocking issue.

---

src/inputs/workbuddy/workbuddy-event-builder.ts
```diff
@@ -0,0 +1,450 @@
...
+    );
+    request['gen_ai.request.id'] = step.requestId;
+    if (step.requestModel) request['gen_ai.request.model'] = step.requestModel;
+    if (firstStep) request['gen_ai.turn.start'] = true;
+    if (pendingDelta.length > 0) request['gen_ai.input.messages_delta'] = pendingDelta;
+
+    const response = baseEntry(
+      'llm.response',
+      responseSource,
+      opts.sessionId,
```
ralf0131, line 201: **[Warning]** In `emitToolWave`, the `pendingDelta` is reset to `[]` after emitting a tool wave. If a subsequent step within the same turn has no user message delta but does have an LLM response, the `gen_ai.input.messages_delta` on the next `llm.request` will be empty. Verify this is intentional — some downstream consumers may expect at least the tool results as input context.

assets/hooks/workbuddy-hook-event-writer.mjs
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env node
```
ralf0131, line 1: **[Info]** The hook journal design is privacy-conscious — only recording structural identifiers (event type, callId, transcript path hints) without capturing prompts, tool arguments, or results. This is the right approach for a telemetry collector.

agents.d/workbuddy.json
```diff
@@ -0,0 +1,30 @@
...
+    "commands": []
+  },
+  "hook": {
+    "settingsPath": "~/.workbuddy/settings.json",
+    "events": [
+      "SessionStart",
+      "UserPromptSubmit",
+      "PreToolUse",
+      "PostToolUse",
+      "Stop"
```
ralf0131, line 16: **[Info]** The `eventMatchers` with `PreToolUse: ".*"` and `PostToolUse: ".*"` will match all tool names. This is correct for a general-purpose collector but worth noting that if WorkBuddy adds high-frequency internal tools, the hook overhead could increase. The `.*` pattern is fine as a default; users can narrow it via settings override if needed.

src/inputs/workbuddy/workbuddy-input.ts
yyuuttaaoo: Fixed in f56da7b. WorkBuddy now persists per-transcript byte offsets plus file identity/size metadata and reads only the uncommitted `[offset, stable EOF)` range after a Stop boundary; it no longer reparses the full transcript on every poll. Truncation/replacement and partial JSONL records are handled explicitly, with regression coverage for chunked writes, stable Stop sealing, and restart behavior.

src/inputs/workbuddy/workbuddy-input.ts
yyuuttaaoo: Thanks for the suggestion. I kept the fixed polling fallback intentionally: missing directories fail fast, and adding backoff would delay detection when WorkBuddy is installed or creates its projects directory later. `fs.watch` remains the low-latency path when available, while the 30-second poll is the bounded reliability fallback.

src/inputs/workbuddy/workbuddy-event-builder.ts
yyuuttaaoo: Thank you. I kept this invariant in the follow-up: unavailable source facts are still omitted, and deterministic event IDs remain stable across reprocessing.

src/inputs/workbuddy/workbuddy-event-builder.ts
yyuuttaaoo: Verified in f56da7b. Clearing `pendingDelta` after emitting a tool wave is intentional because that delta has just been attached to the current request. Subsequent `function_call_result` records append `role=tool` messages back into `pendingDelta`, so the next request receives the tool results. The regression test now explicitly asserts both parallel tool-result messages and their call IDs on the next request.

assets/hooks/workbuddy-hook-event-writer.mjs
yyuuttaaoo: Thank you. This remains unchanged: the hook journal is only a structural wake-up/path hint and does not persist prompt, assistant, tool argument, or tool result content.

agents.d/workbuddy.json
yyuuttaaoo: Agreed. `.*` remains the deliberate default so collection is complete across WorkBuddy tools. If high-frequency internal tools become observable in practice, narrowing matchers can be considered from measured evidence rather than pre-filtering unknown tool names.

src/inputs/workbuddy/workbuddy-event-builder.ts
```diff
@@ -0,0 +1,442 @@
...
+      stepOrdinal++;
+      const step = stepContext(record, opts.sessionId, turnId, stepOrdinal);
+      const message = toMessage(record, 'assistant');
+      const request = baseEntry('llm.request', record, opts.sessionId, step, `request:${step.stepId}`);
+      request['gen_ai.request.id'] = step.requestId;
+      if (step.requestModel) request['gen_ai.request.model'] = step.requestModel;
+      if (firstStep) request['gen_ai.turn.start'] = true;
+      if (pendingDelta.length > 0) request['gen_ai.input.messages_delta'] = pendingDelta;
+
+      const response = baseEntry('llm.response', record, opts.sessionId, step, `response:${sourceId(record)}`);
```
ralf0131: The `finish_reason` is now always `['stop']` regardless of the original record status. The previous code distinguished error cases (`finishReason = 'error'`) and set `error.type = 'model_response_failed'`.

This means downstream consumers can no longer detect failed model responses from telemetry alone. If this is intentional (e.g., WorkBuddy does not surface error status in practice), consider adding a brief comment explaining the rationale. If not, the error classification should be preserved for observability.

src/inputs/workbuddy/workbuddy-input.ts
```diff
@@ -0,0 +1,377 @@
...
+      return false;
+    }
+  }
+
+  protected override async onStart(): Promise<void> {
+    this.watchDirectory(this.projectsDir);
+    this.watchDirectory(this.hookLogDir);
+  }
+
+  protected override async onStop(): Promise<void> {
```
ralf0131, line 79: The byte-offset-based incremental reading is a solid improvement over the previous full-file read. The `TranscriptFileMeta` tracking (size, mtime, stop detection) combined with the stability retry (`STABILITY_RETRY_MS = 5_000`) handles the common edge cases well (active writes, restarts, file rotation).

ralf0131 (commented):
### Summary
This is a re-review after new commits were pushed (a0073b9c → f56da7b8). The update addresses the previous memory concern by switching from full-file reads to byte-offset incremental reading, simplifies the event builder API by removing the `BuiltWorkBuddyEvent` wrapper, and adds test coverage for the new behavior.

**Note:** This PR still has merge conflicts with `main` — please rebase before merging.

### Findings
- **[Warning]** workbuddy-event-builder.ts:200 — `finish_reason` always `'stop'` now; error classification removed
- **[Info]** workbuddy-input.ts:71 — Byte-offset incremental reading is a solid improvement

---

yyuuttaaoo:
@fangxiu-wf Thank you for the real Windows installed-product validation and the detailed follow-up. I have pushed the general correctness fixes and merged the latest `main` into this branch with a regular merge (no rebase).

The current head addresses the non-Windows findings as follows:

- **Transient source failures / checkpoints:** existing byte offsets and file metadata are now copied forward when project discovery, `stat`, or transcript reads are temporarily unavailable. Checkpoints and Hook evidence are pruned only after a complete scan plus a definite not-found result. Replacement/truncation, partial JSONL records, restart, stable Stop sealing, and recovery after temporary directory loss have regression coverage.
- **Incremental transcript reads:** only the uncommitted `[byte offset, stable EOF)` range is parsed. The collector no longer rebuilds every full transcript every 30 seconds.
- **Missing tool identity/name:** Pilot no longer invents fallback IDs or `unknown` names. A TOOL event is omitted when its required identity, name, or timestamp cannot be established. Structural Hook data may repair a missing transcript field only when it matches the same session/transcript/tool boundary unambiguously; parallel calls are covered.
- **LLM and TOOL timing:** the first LLM request starts at prompt submission, later requests start after the preceding tool result, and responses use the earliest native model-output boundary. A transcript response timestamp takes precedence over a Stop Hook observation; Hook time is only a fallback when the transcript time is absent. TOOL duration is the positive difference between a matched result and call. Builder-to-OTLP tests now assert exact positive LLM and TOOL span durations.
- **Finish semantics:** WorkBuddy's generic `status=completed` is intentionally not treated as a model stop/error reason. A stable Stop boundary seals the final assistant response. `tool_call` remains aligned with the existing Claude Code/Pilot wire behavior; the broader repository documentation/enum inconsistency should be handled consistently across adapters rather than changed only for WorkBuddy.
- **Provider semantics:** `workbuddy` remains the routing provider because the transcript does not expose a proven upstream inference-service provider. Model labels such as DeepSeek/Kimi/MiniMax/GLM are recorded as response models but are not used to guess a direct provider relationship.
- **Strict validation:** a requested time window with zero matching events now fails strict mode, and the onboarding guide explicitly requires `E2E_JSONL_AGENT_FILTER=workbuddy` for WorkBuddy GUI acceptance.
- **Concurrent Hook logging:** the shared append-only Hook journal has been replaced by an ephemeral per-session spool. Each Hook process publishes one immutable event via exclusive temporary-file creation plus same-directory atomic rename. Consumed events are removed after the corresponding transcript range is checkpointed, and session data is removed only after transcript deletion is confirmed.
- **Merge conflict:** `deploy/installer-opensource.ps1` preserves current `main`'s specialized Codex cleanup and adds WorkBuddy to the general Hook-config cleanup list.

The reliability lessons are now part of the bilingual Agent onboarding guide and output schema (`5151b4e`): semantic LLM/TOOL time boundaries, transcript-over-Hook precedence, checkpoint behavior under transient I/O failures, and the multi-process per-session Hook spool lifecycle. `AGENTS.md` links directly to this guidance.

Validation on the merged head: `npm run build` passed; the full suite passed with **185 test files / 2,099 tests** (1 file / 5 tests skipped). The PowerShell-specific path still needs a Windows environment.

I have deliberately left these Windows-specific items open for the Windows follow-up rather than claiming unverified fixes:

1. resolve the installer-pinned `node-bin` in the WorkBuddy PowerShell Hook and prove immediate Hook wakeup with PATH lacking Node;
2. support Windows absolute paths in shared Git enrichment;
3. rerun installed-product E2E after those changes and then update the documented Windows version/support statement.

ralf0131 (approved):
Key design decisions look solid:
- **Hook-based wakeup + file polling**: The hook script (`workbuddy-hook-event-writer.mjs`) writes lightweight event hints to `~/.loongsuite-pilot/state/workbuddy/hook-events/`, which the input watches. The 5s stability wait before processing avoids partial transcript reads.
- **Transcript parser**: Handles both array and NDJSON formats, with proper fallback. Session identity resolution from transcript content (not just file path) is correct.
- **Quality gates**: `TOOL_CONTENT_MAX_CHARS` (64KB) truncation with `truncated=true` flag, `MAX_TOOL_PER_SESSION` (5000) safety limit, and `MAX_SESSIONS` (100) LRU eviction are all reasonable.
- **Hook event correlation**: The `matchHookEventToSession` logic correctly handles Stop events by matching `session_id` + `cwd`, and uses the hook as a wakeup signal rather than a data source.
- **Atomic file writes** in the hook script (write to `.tmp`, then rename) prevent partial reads.

One minor note: the `safeSessionDirName` function in the hook uses `path.basename(sessionId)` which is fine for sanitization, but the 64-char truncation could theoretically cause collisions for very long session IDs with common prefixes. The SHA-256 suffix (12 chars) mitigates this well.

LGTM.

---

yyuuttaaoo:
@fangxiu-wf The Windows follow-up is now complete and pushed in `04a7ac4`.

All three intentionally-open Windows items have been implemented and verified on Windows 11 build 26200 with WorkBuddy Desktop 5.3.5.0:

1. **Installer-pinned Node / immediate Hook wakeup**
   - The PowerShell wrapper now resolves `<pilot-data>\node-bin` before NVM/FNM/Volta/Program Files/PATH fallbacks and requires Node >= 18.
   - Installed-product verification used a PATH containing only Windows system and PowerShell directories, with no Node on PATH.
   - The installed WorkBuddy Hook exited 0, returned `{}`, wrote one wakeup event, and completed in 984 ms.

2. **Windows absolute paths in shared Git enrichment**
   - Shared enrichment now accepts both POSIX and Windows absolute paths.
   - Regression coverage includes drive-letter cwd paths and UNC workspace roots.

3. **Installed-product E2E and Windows support statement**
   - Rebuilt and freshly installed the product, then collected an existing real WorkBuddy transcript (63 source records; no transcript content or personal path committed).
   - Output contained 48 WorkBuddy events: 15 `llm.request`, 15 `llm.response`, 9 `tool.call`, and 9 `tool.result`.
   - The transcript included an interrupted turn; the collector now closes it deterministically with a cancelled turn/tool result before the next user turn.
   - Strict JSONL validation passed with zero parse, required-field, type, duplicate-ID, model-pair, tool-pair, turn-boundary, empty-window, or suspicious-placeholder errors.
   - Bilingual support docs now record WorkBuddy Desktop 5.3.5.0 and Windows 11 installed-product E2E coverage.

The installed run also exposed two Windows product issues that are fixed in the same commit: daemon bootstrap now honors `LOONGSUITE_PILOT_CACHE_DIR`, and uninstall uses extended-length paths plus the pinned Node runtime so deep package trees and WorkBuddy Hook config are removed even when PATH lacks Node.

After reboot, I repeated the focused Windows regression on Node v24.18.0: 6 test files / 90 tests passed, `npm run typecheck` passed, and `npm run build` passed. Test installations, scheduled tasks, collector processes, Pilot directories, and WorkBuddy Pilot Hook entries were removed afterward.

ralf0131 (approved):
### Summary
This PR adds comprehensive WorkBuddy agent support to loongsuite-pilot, including a new hook script for event collection, a transcript-based input processor, type definitions, agent system map registration, and deployment integration. The implementation follows existing patterns well with atomic file writes, path traversal protection, secure file permissions, and thorough test coverage across all new components.

**Verdict:** LGTM — well-structured feature addition with solid engineering practices.

### Findings
- **[Info]** `assets/hooks/workbuddy-hook-event-writer.mjs:28` — Consider stripping null bytes from session ID basename for extra safety
- **[Info]** `src/inputs/workbuddy/workbuddy-input.ts` — Good use of BigInt for time comparison and incremental offset tracking

### Highlights
- Atomic write pattern (`.tmp` + `rename`) with exclusive flag prevents race conditions
- Path traversal protection via `isPathWithin` in both hook and input processor
- Secure defaults: 0o700 directories, 0o600 files
- Comprehensive test suite covering edge cases and error paths

---

src/inputs/workbuddy/workbuddy-event-builder.ts
```diff
@@ -0,0 +1,442 @@
...
+      if (output !== undefined) result['gen_ai.tool.call.result'] = output;
+      const status = normalizeToolStatus(record.status);
+      if (status) result['tool.result.status'] = status;
+      const durationMs = tool ? timestampMs(record) - tool.callTimestamp : undefined;
+      if (durationMs !== undefined && durationMs > 0) {
+        result['gen_ai.tool.call.duration'] = durationMs;
+      }
+      if (status === 'failure') result['error.type'] = 'tool_execution_failed';
+      await push(result, record);
+      const resultPart: Record<string, JsonValue> = {
```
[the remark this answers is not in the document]
yyuuttaaoo: Addressed on the current head. Pilot no longer synthesizes a fallback tool-call ID or pairs a result with the earliest unmatched call. A tool event is emitted only when the call ID can be established from the transcript or an unambiguous matching Hook event, and a tool result must resolve to that same ID in the active tool map.

The positional fallback suggested here would be unsafe for parallel tool calls, so ambiguous calls/results are intentionally omitted instead. Regression coverage verifies ambiguous parallel identities are not guessed and strict tool-call/result topology remains valid.

src/inputs/workbuddy/workbuddy-event-builder.ts
yyuuttaaoo: This is intentional and is now documented next to the response construction. WorkBuddy's generic assistant `status=completed` is a lifecycle state, not a model-native finish or error reason, so it is not sufficient evidence for `model_response_failed`.

A stable Stop boundary seals a normal final response. Explicitly interrupted/incomplete turns are handled separately as `cancelled`, including cancellation of pending tool calls, as verified with a real Windows WorkBuddy transcript and strict topology validation.

src/inputs/workbuddy/workbuddy-event-builder.ts
```diff
@@ -0,0 +1,626 @@
...
+  sessionId: string,
+  step: StepContext,
+  eventSeed: string,
+  timestamp: number,
+): AgentActivityEntry {
+  const time = String(timestamp * 1_000_000);
+  return {
+    time_unix_nano: time,
+    observed_time_unix_nano: String(Date.now() * 1_000_000),
+    'event.id': stableId(sessionId, eventSeed),
```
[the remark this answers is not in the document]
yyuuttaaoo: Fixed in `842765b`. Millisecond timestamps are now converted to decimal nanosecond strings with BigInt arithmetic instead of multiplying beyond JavaScript's safe-integer range. The same conversion is used for `observed_time_unix_nano`.

The regression test uses a large but safe millisecond timestamp and verifies exact decimal output without scientific notation. The focused WorkBuddy/validator tests, typecheck, and build all pass.

src/inputs/workbuddy/workbuddy-input.ts
```diff
@@ -0,0 +1,620 @@
...
+
+  const records: WorkBuddyRecord[] = [];
+  let lineStart = 0;
+  let nextOffset = startOffset;
+  for (let index = 0; index < content.length; index++) {
+    if (content[index] !== 0x0a) continue;
+    parseJsonlLine(content.subarray(lineStart, index), records);
+    lineStart = index + 1;
+    nextOffset = startOffset + lineStart;
+  }
```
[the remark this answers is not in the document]
yyuuttaaoo: Fixed in `cea8d29`. `readJsonlRange()` now stops at the first newline-delimited record that fails to parse and returns the last successfully committed boundary, so the stability retry path cannot checkpoint past the failed record.

The regression test writes a malformed newline-delimited record followed by valid data, verifies that no offset is committed, repairs the record, and confirms the complete turn is collected on retry. The focused WorkBuddy/validator tests (46 tests), typecheck, and build all pass on Windows.

ralf0131 (approved):
### Summary

New commit `cea8d293` adds a defensive fix to `readJsonlRange`: when `parseJsonlLine` fails on a malformed JSONL line, the function now returns early without advancing the checkpoint. This prevents the parser from skipping past corrupted data and allows recovery when the file is later fixed.

**Changes reviewed:**
- `src/inputs/workbuddy/workbuddy-input.ts`: Early return on parse failure (+3/-1)
- `tests/unit/inputs/workbuddy-input.test.ts`: New test case for malformed JSONL handling (+48)

**Assessment:**
- ✅ Correctness: Logic is sound — checkpoint only advances past successfully parsed records
- ✅ Test coverage: Comprehensive test covering malformed input and recovery scenario
- ✅ Compatibility: No breaking changes
- ✅ Performance: No concerns

This is a solid defensive fix with good test coverage.

---

yyuuttaaoo:
@copilot resolve the merge conflicts in this pull request

fangxiu-wf:
## WorkBuddy Windows E2E 与生命周期复核结论

结论：**WorkBuddy Trace 主链已经通过真实 Windows E2E，但当前版本建议暂缓合入**。安装、采集和 Trace 质量没有发现回归；仍有 2 个代码问题和 1 组生命周期验收缺口需要闭环。

### 已通过

- PR Head：`cea8d2932f7a18717a110fdd116ff00c3b791f00`
- Windows 11 上完成候选版本构建、安装及运行态验证。
- WorkBuddy Hook 使用 Pilot 固定 Node 路径，在测试进程 PATH 不包含 Node 时仍可正常执行。
- 已从 SLS 回读包含真实 `Write`、`Read` 工具调用的 10 Span Trace：
  - `ENTRY=1 / AGENT=1 / STEP=3 / LLM=3 / TOOL=2`
  - 父子树、Tool Call ID 关联、Token 汇总、消息正文、正时长和 Resource 属性全部通过。
- PR Head 原有 Node 18/20/22 CI 和 CLA 均成功。

### 当前阻塞项

#### 1. PR 已与当前 main 冲突

当前 GitHub 状态为 `mergeable=CONFLICTING / mergeStateStatus=DIRTY`。冲突只有：

```text
tests/unit/normalization/agent-system-map.test.ts
```

原因是 main 新增 `mimo-code`、PR 同一区域新增 `workbuddy`。我在临时合并中同时保留二者后，`tsc`、build 及相关测试 `167 passed / 1 Windows-only skipped` 均通过，冲突本身是机械性的；但正式解冲突提交仍需重新跑 CI 和 Windows E2E。

#### 2. Windows 卸载清理可能“假成功”

`deploy/installer-opensource.ps1` 的 `Remove-HookConfigs` 已包含 `.workbuddy\settings.json`，并会保留第三方 Hook；但调用 Node 清理脚本后没有检查 `$LASTEXITCODE`：

```powershell
& $script:NODE_BIN -e ... $cfg $HOOK_MARKER 2>$null
$ErrorActionPreference = $prevEAP
Msg "    ✅ 已清理: $short" "    ✅ Cleaned: $short"
```

原生进程非零退出默认不会触发 PowerShell `catch`。如果 settings JSON 损坏、文件被锁定或写入失败，这里仍可能输出“已清理”，随后 Pilot 安装文件已被删除，但 WorkBuddy settings 中保留了指向不存在脚本的 Hook。

建议：

- 显式检查 `$LASTEXITCODE`，非零时抛错；
- 使用 `finally` 恢复 `$ErrorActionPreference`；
- 清理后回读 settings，确认 Pilot marker 已不存在，再输出成功；
- 增加真实行为测试，而不只是断言脚本文本包含 `.workbuddy\settings.json`。

#### 3. WorkBuddy Windows Hook 命令不完整支持含空格路径

`src/deployment/hook-strategy.ts` 的 `wrapLegacyPs1Command()` 使用 `cmd.split(' ')` 提取脚本路径，并且生成的 `-File` 路径未加引号。用户目录或自定义 Pilot dataDir 含空格时，WorkBuddy Hook 命令可能无法执行。

同样，`src/core/orchestrator.ts` 构建 watchdog marker 时使用：

```ts
path.basename(def.hook.hookCommand.split(' ')[0])
```

含空格路径下 marker 也可能错误，导致 watchdog 误判并重复 repair。

现有 WorkBuddy PowerShell 测试虽然使用了含空格的临时目录，但它直接通过 `spawnSync(powershell, ['-File', script])` 调用 wrapper，没有覆盖 `HookStrategy → settings.json command → WorkBuddy` 的真实路径。

### Watchdog 现状

设计上已经考虑了两层 watchdog：

1. Hook watchdog 默认开启，每 5 分钟检查所有启用的 `deployMode=hook` Agent；WorkBuddy 会被自动加入，缺失时调用 `deploySingle(workbuddy)` 恢复。
2. Windows Collector 计划任务包含 AtLogOn、每 5 分钟重复触发、`RestartCount=3` 和 `MultipleInstances=IgnoreNew`。

但当前 E2E 只验证了计划任务为 Running、collector 单实例和 Hook 正常，没有实际覆盖：

- 删除一个 WorkBuddy Hook 后自动恢复；
- 恢复时保留第三方 Hook；
- 杀掉 collector 后计划任务自动恢复且仍为单实例；
- 完整卸载后 Pilot Hook、计划任务和安装文件移除，同时保留用户设置与第三方 Hook。

### 建议合入门槛

- [ ] 解决当前 main 冲突并重新跑 CI
- [ ] 修复 Windows 卸载 `$LASTEXITCODE`/回读验证
- [ ] 修复含空格路径的 Hook 命令和 watchdog marker
- [ ] 补充 `install → tamper hook → watchdog repair → kill collector → restart → uninstall` Windows 生命周期验收
- [ ] 对新的 merge ref 再跑一次 WorkBuddy 工具调用 Trace，并从 SLS 回读验收

上一次 10 Span Trace 已充分证明 WorkBuddy 采集能力可用；不过它对应旧 merge ref `e1367d83`（当时 main 为 `076986d`），不能替代当前 main 解冲突后新候选的最终验收。

fangxiu-wf (approved):

PR #202, opened 2026-07-29
fix(opencode): apply configured log poll interval

fangxiu-wf:
## Summary

- register `opencode-log` in the default listener configuration
- pass `listeners["opencode-log"].pollInterval` to the actual `OpenCodeLogInput`
- add regression coverage for config loading and orchestrator wiring

## Motivation

The OpenCode JSONL collector always used its hard-coded 30-second polling interval because the listener configuration was only passed to agent discovery, not to the input itself. After this change, users can reduce local trace collection latency through `~/.loongsuite-pilot/config.json`, while the default remains 30 seconds.

## Testing

- `npm run typecheck`
- `npm test` — 183 test files passed, 2079 tests passed

### DIFF
--- a/src/core/config-loader.ts
+++ b/src/core/config-loader.ts
@@ -372,6 +372,7 @@ function buildListenersConfig(
     'cursor-hook': { enabled: true, pollInterval: 30_000 },
     'claude-code-log': { enabled: true, pollInterval: 30_000 },
     'codex-transcript': { enabled: true, pollInterval: 30_000 },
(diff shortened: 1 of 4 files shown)


rangemer333-cell (approved):

ralf0131 (approved):
LGTM. Clean fix — the `opencode-log` listener was missing from the defaults map, so the configured `pollInterval` was never propagated to `OpenCodeLogInput`. The fix correctly adds the default, passes it through in the orchestrator, and covers both paths with tests.

---

PR #218, opened 2026-08-04
fix(hook): write back winShell to already-installed hook config on upgrade

linrunqi08:
## Background

`agents.d/qoder*.json` declares `"winShell": "powershell"` (the nested inner hook's `shell` field) so that `.ps1` hooks run through PowerShell on Windows. The strategy layer (`hook-strategy.ts`) already passes it through as `def.shell`, but `HookManager` ignored it on the **upgrade path**, so users who installed the hook with an older pilot never got their config updated.

## Root cause

- `isHookInstalled` only compared the `command` string, not `shell`. An entry written by an older version (no `shell`) was still treated as installed → `needsDeploy` returned `false` → no redeploy.
- Even on redeploy, `installHook` returned early once `isCommandPresent` matched, so it never wrote back the new `shell` field.

Net effect: `"shell": "powershell"` was only written on a fresh install; upgrades never took effect, so `.ps1` hooks were not executed via PowerShell on Windows.

## Fix (`src/hooks/hook-manager.ts`)

- `isHookInstalled`: after the `command` matches, additionally require the matching nested inner entry to declare the required `shell` (only when `def.shell` is set). A stale entry without `shell` is treated as "not fully installed" → triggers redeploy.
- `installHook`: when an existing entry is matched, repair (write back) the `shell` field in place — idempotent, no duplicate entries.
- New helpers `hasRequiredShell` / `applyShellToEntries` are both no-ops when `shell` is unset, so non-Windows and agents that don't declare `winShell` (codex / cursor / kiro) are unaffected.

## Note

The jsonc install path (`installJsoncHook`) does not do shell repair: no jsonc agent currently declares `winShell`, so `hasRequiredShell` returns `true` for them and no redeploy loop occurs. If a jsonc agent gains `winShell` in the future, that path needs the same fix.

## Tests

- 3 new unit tests covering the upgrade scenarios (stale entry treated as not-installed / in-place repair without duplication / shell dimension ignored when winShell is not declared).
- `hook-manager` + `hook-strategy`: 58 passing; full `tests/unit/deployment`: 205 passing; `tsc --noEmit` passes.

### DIFF
--- a/src/hooks/hook-manager.ts
+++ b/src/hooks/hook-manager.ts
@@ -118,7 +118,11 @@ export class HookManager {
       const updatedArr = target[lastKey] as any[];
 
       if (this.isCommandPresent(updatedArr, def.hookCommand)) {
-        if (updatedArr !== arr) {
+        // Entry already present. It may predate a newly-added `shell` field
+        // (e.g. Qoder gained winShell after the hook was first installed) —
+        // repair the stale entry in place instead of leaving it untouched.
+        const shellRepaired = this.applyShellToEntries(updatedArr, def.hookCommand, def.shell);
+        if (updatedArr !== arr || shellRepaired) {
           await writeJsonFile(def.settingsPath, settings);
         }
         logger.debug('hook already installed', { agentId: def.agentId });
@@ -227,7 +231,12 @@ export class HookManager {
         return false;
       }
 
-      return this.isCommandPresent(hooks, def.hookCommand);
+      if (!this.isCommandPresent(hooks, def.hookCommand)) return false;
+
+      // A matching command whose nested entry lacks the required `shell` counts
+      // as not fully installed, so an upgrade that adds winShell (Qoder family
+      // on Windows) redeploys and repairs the entry rather than skipping it.
+      return this.hasRequiredShell(hooks, def.hookCommand, def.shell);
     } catch {
       return false;
     }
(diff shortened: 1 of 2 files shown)


linrunqi08:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 0 |
| Low      | 4 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| Resource release | PASS |
| Deadlock / hang risk | PASS |
| State-recovery correctness | PASS |

The main path (strict-json + nested Qoder family) is correct and **converges without a redeploy loop**: `isHookInstalled` detects the stale (shell-less) entry → `needsDeploy=true` → `installHook` repairs `shell` in place → next round `hasRequiredShell=true` → `needsDeploy=false`. Non-Windows / agents without `winShell` are strict no-ops.

### Verdict: not blocking ✅

All 4 findings are **latent — unreachable in the current config matrix**. Verified: the only agents declaring `winShell` are qoder / qoder-cn / qoder-work / qoder-work-cn (all `nested` + strict-json); the only jsonc agent (`qwen-code-cli`) does not declare `winShell` → **jsonc ∩ winShell = ∅**.

Suggested follow-ups (non-blocking):

- **[Low] `src/hooks/hook-manager.ts` — jsonc/strict-json asymmetry.** `hasRequiredShell` is applied in the *shared* `isHookInstalled`, but the in-place shell repair only exists in the strict-json `installHook`; `installJsoncHook` never writes back `shell`. If a jsonc agent ever gains `winShell`, an upgrade hitting a shell-less entry would make `isHookInstalled` permanently `false` → `needsDeploy` permanently `true` → `installJsoncHook` no-ops → **never converges across deploy cycles** (and the hook stays broken on Windows). Already documented in the PR description. Suggest symmetric repair via `editJsonc`, or a config-level assertion that `jsonc` and `winShell` are mutually exclusive with a `logger.warn`.
- **[Low] flat vs nested.** `hasRequiredShell` / `applyShellToEntries` only handle nested entries, while `isCommandPresent` also matches flat ones. A future `flat + winShell` combo would not converge. Suggest constraining `winShell` to `format: 'nested'`.
- **[Low] `some` vs `all`.** `hasRequiredShell` uses `some` while `applyShellToEntries` repairs all. With duplicate entries for the same command where only one carries `shell`, `isHookInstalled` returns true and the stale one is never repaired (normal install dedups, so very low probability).
- **[Low] test gaps.** Missing: batch repair of multiple entries; an idempotent "no write when shell already correct" assertion; a jsonc scenario; the `applyShellToEntries` stale-overwrite branch (`h.shell !== shell`).

### Highlights

- Fix targets the real root cause (command-only check + early return without write-back), not a surface patch.
- The "no-op when `shell` is unset" design cleanly isolates the blast radius to Windows + winShell agents.
- The jsonc limitation is proactively documented in the PR description.
- Comments match behavior; new helpers reuse the existing matching convention with no duplication.

---

ralf0131 (approved):
### Summary
Clean fix for stale hook entries that lack the `winShell` field after upgrade. The logic is correct: `isHookInstalled()` now returns `false` when `shell` is required but missing, triggering re-install; `applyShellToEntries()` repairs in-place for the already-installed path. Tests cover all three scenarios (detect, repair, no-op).

LGTM.

---

ralf0131 (approved):
LGTM. Well-designed upgrade path fix. `hasRequiredShell` correctly identifies stale entries missing the `shell` field, and `applyShellToEntries` repairs them in place without duplicating hooks. The `isHookInstalled` change to return false for missing `shell` ensures upgrades redeploy correctly. Comprehensive test coverage including the no-op case for non-Windows.

---

PR #247, opened 2026-08-12
fix(ps1): make hooks and CLI wrapper CLM-safe and pin JSON interchange to UTF-8

linrunqi08:
## Why

Windows validation machines run a WDAC application-control policy, so any `.ps1` not covered by the policy executes in **ConstrainedLanguage Mode** (CLM). Under CLM only CoreTypes may be cast or have *methods* invoked. The hooks wrap their work in a fail-open `catch`, so a CLM violation there doesn't fail loudly — it silently drops telemetry.

Worth recording: `about_Language_Modes`' allowed-types list is **not** authoritative for Windows PowerShell 5.1. `[pscustomobject]@{}` really does throw `ConversionSupportedOnlyToCoreTypes`, because the accelerator targets the internal type `LanguagePrimitives+InternalPSCustomObject`.

## What changed

**1. CLM violations removed**

| Site | Before | After |
|---|---|---|
| `codex-…-hook.ps1` | `[ordered]@{ … }` | `@{ … }` |
| `codex`/`workbuddy` hooks | `[System.IO.Path]::HasExtension($candidate)` | `$candidate -match '\.[^\\/.]+$'` |
| `scripts/loongsuite-pilot.ps1` | `[System.IO.Path]::IsPathRooted($p)` | `-match '^([A-Za-z]:[\\/]\|\\\\[^\\/]+[\\/])'` |
| `scripts/loongsuite-pilot.ps1` | `$state.PSObject.Properties.Remove('hermes-agent')` | `$state \| Select-Object -Property * -ExcludeProperty 'hermes-agent'` |

**2. JSON interchange pinned to UTF-8**

`Set-Content` / `Add-Content` / `Get-Content` default to the **ANSI codepage** on 5.1 (GBK/936 on Chinese Windows) while node reads and writes the same files as UTF-8. A `targetDir` under `C:\Users\<non-ASCII>\` round-tripped as mojibake, and the plugin at that path went uncleaned on rollback. Every JSON/JSONL interchange site now passes `-Encoding UTF8`.

The paired half is on the node side: 5.1 has no `utf8NoBOM`, so `-Encoding UTF8` **always** emits a BOM, and `JSON.parse` rejects a leading BOM. `readJsonFile()` swallows parse errors and returns `null`, which callers read as *"file absent"* — so a BOM alone would silently reset deployment state instead of failing. `readJsonFile()` and `parseJsonDocument()` now strip it.

`readJsonDocument()` deliberately does **not** strip: its callers reuse `raw` as a compare-and-swap baseline and as `editJsonc`'s base text, so stripping there would corrupt the round-trip. `tests/unit/utils/json-document.test.ts` pins both halves of that asymmetry.

Fixing only one side of this boundary would be worse than fixing neither.

## Guardrails

Three static ratchets, each sweeping every tracked `.ps1` via `git ls-files` with a reverse-asserted whitelist that may shrink but never grow:

- `tests/unit/scripts/ps1-clm-safe.test.mjs` — banned shapes (`[pscustomobject]`, `[ordered]`, `New-Object`, `Add-Type`, `Invoke-Expression`, `.PSObject`, `.ContainsKey()`, `class`/`enum`) plus an explicit allow-list for the .NET static accesses that are legitimate (property *gets*, core types, or calls inside a try with a working fallback).
- `tests/unit/scripts/ps1-comments-ascii.test.mjs` — comments stay ASCII. A BOM-less `.ps1` decodes with the ANSI codepage, and the documented `irm <url> | iex` decodes per the HTTP charset and never looks at a BOM; either way a mangled byte can carry a quote and take the parser down. Bilingual *output* strings stay exempt.
- `tests/unit/scripts/ps1-json-encoding.test.mjs` — `-Encoding UTF8` at every interchange site, plus the node-side BOM strip.

The encoding ratchet keys off the **target path**, not the converter. The four hook `Log-Error` helpers build their JSONL line by string interpolation and contain no `ConvertTo-Json` anywhere, so a converter-keyed rule stayed silent while all four appended with the ANSI codepage. That blind spot is pinned by its own test.

## Testing

- `tests/unit/scripts/*` + `tests/unit/utils/{fs-utils,json-document}.test.ts` — all pass.
- Full suite compared against clean `main`: identical pre-existing failures (`hermes-agent-plugin`, `hermes-agent-event-log-flow`, `plugin-inject-strategy`, `cursor/resource-context`, `qoder-hook-processor-cold-start`). The few hook-processor files that flake under parallel load pass in isolation. No regression from this change.

### DIFF
--- a/assets/hooks/claude-code-loongsuite-pilot-hook.ps1
+++ b/assets/hooks/claude-code-loongsuite-pilot-hook.ps1
@@ -32,7 +32,12 @@ function Log-Error {
         $time = (Get-Date -Format "yyyy-MM-ddTHH:mm:ssZ")
         $escapedMsg = $Message -replace '\\', '\\\\' -replace '"', '\"'
         $line = "{`"time`":`"$time`",`"gen_ai.agent.type`":`"claude-code`",`"stage`":`"$Stage`",`"error.type`":`"ps1_$Stage`",`"error.message`":`"$escapedMsg`"}"
-        Add-Content -Path $file -Value $line
+        # Explicit UTF8 and -LiteralPath: Add-Content defaults to the ANSI codepage, so a
+        # Chinese username in a path or message lands as mojibake -- and node's
+        # shared/error-logger.mjs appends UTF-8 to this very file, making it mixed-encoding.
+        # A BOM on the first line is harmless here: nothing machine-parses this log (the
+        # retention service only deletes it by age). Matches codex-loongsuite-pilot-hook.ps1.
+        Add-Content -LiteralPath $file -Value $line -Encoding UTF8
     } catch {}
 }
 
--- a/assets/hooks/codex-loongsuite-pilot-hook.ps1
+++ b/assets/hooks/codex-loongsuite-pilot-hook.ps1
@@ -36,14 +36,22 @@ function Log-Error {
             New-Item -ItemType Directory -Path $dir -Force | Out-Null
         }
         $file = Join-Path $dir "codex-error-$day.jsonl"
-        $record = [ordered]@{
+        # Plain @{}, never [ordered]@{}: on a WDAC machine this hook runs in
+        # ConstrainedLanguage mode, where the accelerator-converts-an-@{}-literal shape
+        # can throw (see the CLM header in deploy/installer.ps1 -- [pscustomobject] does
+        # exactly that on 5.1 despite being documented as allowed). The catch below would
+        # swallow it, so the symptom is an error log that is never written on precisely
+        # the machines where it is needed. Key order is cosmetic in a JSONL log line.
+        $record = @{
             time = (Get-Date).ToUniversalTime().ToString("o")
             "gen_ai.agent.type" = "codex"
             stage = $Stage
             "error.type" = "ps1_$Stage"
             "error.message" = $Message
         }
-        Add-Content -LiteralPath $file -Value ($record | ConvertTo-Json -Compress)
+        # Explicit UTF8: Add-Content defaults to the ANSI codepage, which garbles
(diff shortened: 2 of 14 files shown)


ralf0131 (approved):
### Summary

Excellent fix for Windows WDAC/CLM compatibility and UTF-8 encoding consistency. The changes are minimal, well-targeted, and thoroughly documented.

### Findings

- **[Info]** `codex-loongsuite-pilot-hook.ps1` — The `[ordered]@{}` → `@{}` change is correct; JSONL key order is cosmetic and the comment explains the rationale well.
- **[Info]** `loongsuite-pilot.ps1` — The `Select-Object -ExcludeProperty` replacement for `.PSObject.Properties.Remove()` is the right CLM-safe approach. One minor note: `Select-Object -Property *` creates a new object, so if downstream code relies on reference identity of the original hashtable, this could behave differently. Given the context (JSON serialization), this is not a concern here.
- **[Info]** `json-document.ts` — The deliberate BOM asymmetry between `readJsonFile()` (strips BOM) and `readJsonDocument()` (preserves raw for CAS) is well-designed and correctly pinned by tests.
- **[Info]** `ps1-clm-safe.test.mjs` — The ratcheting guardrail that scans all tracked `.ps1` files for banned CLM patterns is an excellent preventive measure. The allow-list for legitimate .NET static accesses is a good balance.

### Verdict

All four review dimensions check out:
- **Correctness**: CLM workarounds are semantically equivalent to the originals
- **Performance**: No perf impact (regex replaces are on cold paths)
- **Tests**: New guardrail tests prevent CLM regressions; BOM asymmetry is pinned
- **Compatibility**: No public API changes; purely internal implementation fixes

LGTM 👍

---

PR #257, opened 2026-08-14
feat: add multimodal collection for qoder

xiongyunn:
## Summary

This PR adds multimodal media collection for Qoder IDE and Qoder CLI to Pilot. Local image paths are read at write time, converted to storage `uri` parts via the shared `MultimodalProcessor.pathToUri` path, uploaded asynchronously to OSS or SLS PutObject, and attached as structured GenAI metadata without changing Pilot’s existing event wire contract.

- Reuses the existing `src/multimodal` stack (Processor, Uploader, OSS/SLS, retry, queue backpressure). Qoder does not add a new output surface or storage protocol.
- Qoder IDE: SQLite `attachedImagePaths` (input), `Image file:` / ImageGen paths (tool), and `` in assistant text (output).
- Qoder CLI: paste `[Image: source: <path>]` and `@path` (input); `Read image:` / `Image file:` / ImageGen paths (tool). No output surface — CLI assistant text does not embed images.
- Hook processor keeps `isMeta` paste-source rows on the same turn as the real user prompt so CLI local paths reach `messages_delta`.
- Fail-open: missing files, oversized images, and `pathToUri` throws do not drop the original text events; skips are logged.
- Adds unit coverage for IDE/CLI extractors, `uploadMode` gates, part cap, fail-open, and hook turn folding.
- Updates English and Chinese multimodal / agents docs for Qoder IDE + CLI coverage.

### Qoder IDE (`qoder` / `qoder-idea`)

- After IDE token enrich, `enrichIdeMultimodal` walks the session and converts local image paths with `pathToUri`.
- `uploadMode` surfaces:
  - `input` — SQLite `chat_record.extra.attachedImagePaths` (and context images) onto the matching `llm.request` / user `messages_delta`;
  - `tool` — `tool.result` text `Image file: <path>` and ImageGen `The absolute path of the image is: <path>` rewritten to text + `uri` parts;
  - `output` — `llm.response` markdown `` with an image extension;
  - `both` — all of the above; `none` skips.
- Glob listings that were never visually read are not collected.
- Process-local request_id → attached-path LRU plus processor path/sha256 cache avoid repeat disk reads and duplicate uploads.

### Qoder CLI (`qoder-cli`)

- Same `qoder-trace` collect path, after CLI token enrich. No SQLite. Config key remains `agents.qoder.multimodal`.
- `uploadMode` surfaces:
  - `input` — `[Image: source: <path>]` (strip `, original …` metadata) and `@path` (relative paths join `agent.qoder.cwd`);
  - `tool` — `Read image: <path>`, `Image file: <path>`, ImageGen absolute path;
  - `output` — unused (`both` = input + tool).
- Remote OSS / presigned URLs in transcript image blocks are not used as the upload source; local `pathToUri` is.
- Prompt text like `picture/pipeline.jpg这个是什么？` (no `@`) is not parsed as a path.
- Per-path convert is unique’d at extract time; `other` and `llm.request` that both carry the source text each get a `uri` part (upload still deduped by cache/sha256).
- Caps converted images per carrier at `MAX_MULTIMODAL_PARTS` (10).

### Hook processor (CLI paste path)

- Transcript paste is two user rows: the human prompt (`[Image #0]…` + remote image block) and an `isMeta` sidecar `[Image: source: <local path>]`.
- `isRealUserPrompt` is `type === user && !tool_result && !isMeta`. `isMeta` no longer starts a new turn (same as tool results).
- Used for turn split, incremental/triggered windows, and `userRow` selection (fallback to the old user-and-not-tool-result find).
- `buildUserMessageParts` (CLI only) keeps the primary prompt and appends other same-turn user texts (the Image:source sidecar) onto `other` / `llm.request` `messages_delta`. IDE parts stay a single primary text.

### Wiring

- `QoderTraceInput.collect`: IDE enrich per session; CLI enrich per turn; both inject `pathToUri` from the process-scoped processor when infra + `agents.qoder.multimodal.uploadMode` enable it.
- Supported agent ids: `codex`, `qoder` (`MULTIMODAL_SUPPORTED_AGENT_IDS`). `qoder-cli` / `qoder-idea` resolve to `qoder` via the existing agent alias map.

## Implementation overview

```text
Qoder transcript / hook history
  -> hook processor (CLI: fold isMeta Image:source into the same turn parts)
  -> QoderTraceInput.collect
       IDE: enrichIdeMultimodal (SQLite attach + tool/output paths)
       CLI: enrichCliMultimodal (Image:source / @path + Read/ImageGen)
  -> MultimodalProcessor.pathToUri
       stat / read bytes / hash / joinStorageUri
       enqueue Uploader.upload (async)
  -> message / tool.result parts: text + uri (no base64 in output)
  -> optional gen_ai.input.multimodal_metadata from uri parts
  -> Pilot MultiFlusher emits the existing wire protocol
```

Configuration split (unchanged from the Codex multimodal PR):

- Global infra: `config.multimodal` (`uploader`, credentials, `storageBasePath` for OSS; SLS derives `sls://{project}/{logstore}`).
- Per-agent policy: `agents.qoder.multimodal.uploadMode` (`none` default; `input` / `tool` / `output` / `both`).

### Automated checks

- `tests/unit/inputs/qoder-cli-multimodal.test.ts` — paste / `@` extract, tool prefixes, glob skip, `uploadMode` gates, part cap, fail-open, both carriers get uri
- `tests/unit/inputs/qoder-trace-input.test.ts` — IDE + CLI collect convert, Image file / ImageGen, part cap
- `tests/unit/hooks/qoder-hook-processor-retry.test.mjs` — CLI Image:source appended on the primary prompt; isMeta does not start a new turn; IDE does not append Image:source parts


ralf0131 (approved):
### Summary

This PR adds multimodal media collection for Qoder IDE and CLI, reusing the existing `src/multimodal` stack (Processor, Uploader, OSS/SLS, retry, queue backpressure). The implementation is clean, well-structured, and follows established patterns.

**Key observations:**
- **IDE multimodal** (`qoder-ide-multimodal.ts`): Correctly walks session entries, converts local image paths with `pathToUri`, handles input/tool/output surfaces with proper `uploadMode` gating. LRU cache avoids duplicate uploads.
- **CLI multimodal** (`qoder-cli-multimodal.ts`): Parses `[Image: source:]` paste and `@path` patterns, handles tool result images. No output surface (correct for CLI).
- **SQLite reader** (`sqlite-token-reader.ts`): Batch query with parameterized SQL, fail-open, proper JSON parsing of `extra.attachedImagePaths` with context fallback.
- **Hook processor** (`qoder-hook-processor.mjs`): `isRealUserPrompt()` correctly adds `isMeta` filtering to exclude paste-source rows from turn boundaries. `buildUserMessageParts()` attaches local image URIs to CLI user prompts.
- **Orchestrator wiring**: Clean integration with existing multimodal infrastructure, proper config gating.

**Design strengths:**
- Fail-open throughout — missing files, oversized images, and `pathToUri` throws do not drop original events
- Process-local LRU + sha256 cache avoid repeat disk reads and duplicate uploads
- `uploadMode` surfaces (input/tool/output/both/none) provide fine-grained control
- Remote URLs are not used as upload sources — only local `pathToUri`

LGTM ✅

---

ralf0131 (approved):
### Summary
Re-review after new commit `f5ebc6f` — adds remote/multi-profile SQLite path lookup for Qoder. Clean refactoring: extracts `QODER_DB_TAIL` constant, adds `listHashedProfileDbPaths()` for hashed profile directories, and deduplicates candidates via `Set`.

LGTM — extends path resolution correctly for Linux remote and multi-profile scenarios.

---

ralf0131 (approved):
### Summary
Re-review after new commit — adds remote/multi-profile SQLite path lookup for Qoder. Clean refactoring with hashed profile directory scanning and Set-based deduplication.

LGTM — extends path resolution correctly for Linux remote and multi-profile scenarios.

---

src/inputs/qoder-trace/qoder-ide-multimodal.ts
```diff
@@ -0,0 +1,382 @@
...
+        if (!part || typeof part !== 'object') continue;
+        const p = part as Record<string, unknown>;
+        if (p.type === 'text' && typeof p.content === 'string') texts.push(p.content);
+      }
+      if (texts.length === 0) continue;
+
+      const paths = extractMarkdownImagePaths(texts.join('\n'));
+      if (paths.length === 0) continue;
+
+      const uriParts = await convertPathsToUriParts(paths, pathToUri, timeMs, stats);
```
yyuuttaaoo, line 312: **[High]** Qoder transcript 和模型输出中的路径可直接触发任意本地文件、符号链接目标或 Windows UNC 路径读取与上传。

**影响:** 路径只按扩展名判定图片，`stat`/`readFile` 会跟随符号链接，也没有可信根目录、`realpath` containment、magic MIME、UNC/device-path 检查。工作区内容或提示注入可让模型输出 `x`，把指向敏感文件的 symlink 内容上传；Windows UNC 还会触发 SMB 访问、凭据泄露或长时间挂起。
**建议:** 只接受具有可信 provenance 的附件和工具结果；将 IDE/CLI 路径限制在已知 attachment/generated-image 根目录，`realpath` 后验证 containment，拒绝 symlink、UNC 和 device path，并校验内容 magic/MIME 后再上传。

src/inputs/qoder-trace/qoder-cli-multimodal.ts
```diff
@@ -0,0 +1,319 @@
...
+  isImageFilePath,
+  MAX_MULTIMODAL_PARTS,
+  type PathToUriFn,
+  type UriPart,
+} from '../../multimodal/index.js';
+import { createLogger } from '../../utils/logger.js';
+
+const logger = createLogger('QoderCliMultimodal');
+
+const IMAGE_SOURCE_RE = /\[Image:\s*source:\s*([^\]]+)\]/gi;
```
yyuuttaaoo: **[High]** 当前 Qoder CLI `1.1.28` 的真实图片粘贴 transcript 不包含实现依赖的 `[Image: source: path]` sidecar。

**影响:** 用户 15:59 的真实输入只有 `[Image #0] ...`，源图位于 session-scoped tmp image 目录；head extractor 实测返回空数组。最常用的 CLI paste input 不会生成 URI，尽管 Qoder 自身能识图；现有测试使用了人工构造的旧 sidecar。
**建议:** 用当前客户端的真实脱敏 fixture 固化契约，从稳定 transcript/hook 字段或 session/cwd-scoped asset mapping 关联 `[Image #N]`；若只能支持部分版本，增加显式版本门槛并收窄文档声明。

src/inputs/qoder-trace/qoder-trace-input.ts
```diff
@@ -116,6 +137,23 @@ export class QoderTraceInput extends BaseInput {
       }
     }
 
+    if (this.multimodalEnabled && this.multimodalProcessor) {
+      const pathToUri = (filePath: string, timeUnixMs?: number) =>
+        this.multimodalProcessor!.pathToUri(filePath, timeUnixMs);
+      for (const sessionEntries of ideSessionGroups.values()) {
+        await enrichIdeMultimodal(sessionEntries, {
```
yyuuttaaoo, line 206: **[Medium]** Qoder collection 串行等待没有 deadline 或取消机制的本地路径 `stat/readFile`。

**影响:** 网络盘、FUSE、失联挂载或 UNC 读取若不返回，当前批次的普通文本也不能交给 InputManager；停止时 `BaseInput.stop()` 会无限等待 cycle，使后续 processor、flusher 和 StateStore shutdown 不可达。
**建议:** 给每图文件操作加可观测 deadline/取消信号，超时后继续文本路径；停止流程先拒绝新媒体任务并使长批次可中断，增加 never-resolving `pathToUri` 的 collect/stop 回归测试。

src/inputs/qoder-trace/qoder-cli-multimodal.ts
```diff
@@ -0,0 +1,319 @@
...
+const AT_IMAGE_RE = /@(\S+?\.(?:png|jpe?g|gif|webp|bmp|svg|ico|tif|tiff))\b/gi;
```
yyuuttaaoo: **[Medium]** 图片正则和候选转换都缺少输入工作量上限，`MAX_MULTIMODAL_PARTS` 只限制成功结果而不限制尝试次数。

**影响:** `@a` 或 `![` 重复文本让同步正则呈二次方增长；Node 24 下 80 KB CLI/IDE 样本分别约 1.42 s/1.89 s。十万条 missing path 还会执行十万次串行 `stat` 并刷日志，阻塞同一 input 的正常文本采集和 offset 推进。
**建议:** 先限制载体文本与单路径长度，使用不跨下一个 marker 的线性/有界 parser；对去重后的 candidate 在转换前 `slice(0, MAX_MULTIMODAL_PARTS)`，限制 attempts 而非 successes，并把剩余数量聚合为一条日志。

src/inputs/qoder-trace/qoder-ide-multimodal.ts
```diff
@@ -0,0 +1,382 @@
...
+      missing.push(id);
+    }
+  }
+
+  if (missing.length > 0) {
+    try {
+      const fetched = await readAttachedImagePathsForRequestIds(missing);
+      for (const id of missing) {
+        const paths = fetched.get(id) ?? [];
+        attachedPathsByRequestId.set(id, paths);
```
yyuuttaaoo: **[Medium]** SQLite 查询失败、尚未落库与真实无附件都被永久负缓存为 `[]`。

**影响:** reader 吞掉单库查询错误并返回空 Map；caller 随即缓存所有 missing request ID。短暂锁冲突或落库延迟后不会重查，而 hook checkpoint 继续推进，IDE 图片会静默永久丢失。
**建议:** 返回 `{paths, complete}` 或等价三态；只有所有候选 DB 成功扫描后才负缓存。对不完整查询做有界重试/短 TTL，并增加“首次失败或空，随后成功”的恢复测试与计数指标。

assets/hooks/qoder-hook-processor.mjs
```diff
@@ -1311,16 +1312,14 @@ function assignContentToBoundaries(boundaries, contentEvents) {
...
+ * isMeta rows stay on the preceding prompt, same as tool results.
  */
-function splitContentEventsIntoTurns(contentEvents) {
+export function splitContentEventsIntoTurns(contentEvents) {
   const turns = [];
   let currentTurn = [];
 
   for (const row of contentEvents) {
-    if (row.type === 'user' && !isToolResult(row)) {
+    if (isRealUserPrompt(row)) {
```
yyuuttaaoo, line 1324: **[Medium]** leading `isMeta` 被拆成独立 turn，同时消息构造又把同 turn 所有非-tool user 元数据加入用户 parts，而非只折叠图片 sidecar。

**影响:** `[Image: source:]` 若位于窗口开头，bootstrap/recovery 保留 latest turn 时会丢路径，incremental 则把图片和 prompt 分到不同 turn；本机存在的 `isMeta` local-command caveat 还可能污染普通用户输入内容。
**建议:** 用 `pendingPrelude` 将 prompt 前的明确图片 meta 绑定到下一 real prompt；`buildUserMessageParts` 只附加已识别的 `isMeta && [Image: source:]` 行，并覆盖 incremental、missing-cursor、truncated 三种窗口测试。

src/inputs/qoder-trace/qoder-ide-multimodal.ts
```diff
@@ -0,0 +1,382 @@
...
+  const seen = new Set<string>();
+  for (const filePath of paths) {
+    if (parts.length >= MAX_MULTIMODAL_PARTS) {
+      stats.skipped += 1;
+      continue;
+    }
+    const trimmed = filePath.trim();
+    if (!trimmed || seen.has(trimmed)) continue;
+    seen.add(trimmed);
+    const result = await pathToUri(trimmed, timeMs);
```
yyuuttaaoo: **[Medium]** 同一路径的 URI cache 没有包含事件日，跨日事件会复用前一天 URI。

**影响:** URI 包含事件本地日，但 `pathUriCache` 只比较 path、mtime 和 size。实测同一图片分别传 8 月 24 日和 25 日，两次都返回 `.../20260824/<sha>.png`，违反输出 schema 并破坏按事件日查询。
**建议:** 将本地 `YYYYMMDD` 加入 cache/inflight key，或只缓存 hash/MIME 再按每次事件时间生成 URI；增加同一路径跨日的回归测试。

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo, line 312: **[Medium]** IDE output 的 Markdown parser 既漏掉合法空格/title 形式，也没有按 `agent.qoder.cwd` 解析相对路径。（部分为既有 Copilot comment `3841528581`）

**影响:** `x`、`x` 和 `x` 会漏采或错误地相对 collector cwd 查找；CLI 相对路径则已有 workspace-cwd 语义。
**建议:** 使用符合 Markdown destination/title 语义的有界 parser，提取后统一调用平台感知 resolver，并让 IDE tool/output 相对路径基于 entry 的 `agent.qoder.cwd`。

yyuuttaaoo:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 3 |
| Medium   | 8 |
| Low      | 1 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | ❌ FAIL |
| 死锁/卡死风险 | ❌ FAIL |
| 状态恢复正确性 | ❌ FAIL |

### 总体结论

发现 3 个阻断问题（0 Critical + 3 High），建议修复后再合入。

CLI 当前真实图片粘贴格式无法被提取；本地路径缺少可信根目录与 symlink/UNC 防护；`qoder-idea` relabel 后会绕过 `agents.qoder.captureMessageContent`。此外，未设 deadline 的文件读取、无界候选处理和 SQLite 永久负缓存会影响普通文本采集、关闭流程与恢复正确性。

### Highlights（正向实践）

- 复用了共享 `MultimodalProcessor` 和统一 URI part schema。
- 普通 missing/invalid file 路径保持 fail-open，远端 PUT 仍为异步。
- 本机真实 IDE SQLite 附件已通过 PR-head 提取与事件富化 POC。
- 现有 4 条 Copilot inline finding 已逐条复核，发布时不重复评论。

---
*评审报告详见: `code-review/pr-257/final-report.md`*  

src/inputs/qoder-trace/qoder-cli-multimodal.ts
yyuuttaaoo: 补充本机 source asset 的定位方法。这个方法用于验证当前客户端的数据事实，不建议直接把“按时间猜文件”当成生产关联协议：

1. 在 `~/.qoder/projects` 下搜索真实图片 marker，例如 `rg -l --glob '*.jsonl' '\[Image #[0-9]+\]' ~/.qoder/projects`，得到 `~/.qoder/projects/<project-slug>/<session-id>.jsonl`。
2. 从 transcript 文件名取得 `<session-id>`，从父目录取得 Qoder 编码后的 `<project-slug>`；该 user row 同时给出图片序号和事件时间。
3. 检查 `~/.qoder/tmp/<project-slug>/images/<session-id>/`。本次只有 `0-clipboard-<timestamp>.png.png`，前缀 `0` 对应 `[Image #0]`，文件 mtime 与 user row 在同一秒，Qoder 随后的回答也证明它实际读到了这张图。
4. 反向执行 `rg -n '\[Image:\s*source:' ~/.qoder/projects`，并按 session/prompt id 搜索 `~/.qoder`；本机 157 个 transcript 中 source sidecar 为 0，没有发现另一条持久化 source-path carrier。
5. 将真实 user text 直接传给 PR head 的 `extractInputImagePaths()`，结果为 `[]`。

因此 tmp 目录证明“源图存在且能按 session/index 定位”，但稳定修复应优先使用明确的 transcript/hook 字段；若只能回退到 asset mapping，需要同时校验 session、project、image index、时间窗口和文件唯一性，并用真实脱敏 fixture 固化。

src/inputs/qoder-trace/qoder-cli-multimodal.ts
yyuuttaaoo: 补充完整的本机性能数据和输入形状。环境是 macOS arm64、Node `24.19.0`，直接运行 PR head 的两个正则，对无匹配文本测同步 `exec` 总耗时：

| 输入大小 | CLI `('@a').repeat(...)` | IDE `'!['.repeat(...)` |
|---:|---:|---:|
| 20 KB | 约 91.5 ms | 约 122 ms |
| 40 KB | 约 333 ms | 约 482 ms |
| 80 KB | 约 1.39 s | 约 1.94 s |
| 160 KB | 约 5.72 s | 约 7.55 s |

输入翻倍、耗时接近 4 倍，符合二次方增长；这是同步 regex，会直接占住 collector event loop。另一个独立放大器是 `MAX_MULTIMODAL_PARTS` 只检查 `parts.length`：构造大量 `@/tmp/missing-N.png` 时，失败不会增加 `parts.length`，所以仍会逐个串行 `stat` 并逐条 warn。建议回归测试同时断言超长无匹配输入的耗时上界，以及前 N 个路径全部 missing 时 `pathToUri` 调用次数不超过 10。

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: 补充可复现步骤：使用同一个 `MultimodalProcessor` 和 recording uploader，对同一个未修改的真实 PNG 依次调用：

```text
pathToUri(path, 2026-08-24T12:00:00+08:00)
pathToUri(path, 2026-08-25T12:00:00+08:00)
```

两次文件的 path、mtime、size 都相同。第一次返回并 enqueue `.../20260824/<sha>.png`；第二次在 `pathUriCache.get(key)` 处命中完整 `UriResult`，仍返回 `.../20260824/<sha>.png`，recording uploader 也只看到第一天的 key。第二次根本没有走到 `bytesToUri(... time_unix_ms ...)`，所以传入的新事件时间无法生效。跨日测试应同时断言第二次 URI 为 `20260825/...`，以及 uploader 收到两个不同 day key。

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: 还有一个与本线程同源、之前未单独写进 inline finding 的 **Medium residual risk**：30 MB size cap 存在 pathname TOCTOU。

`pathToUri()` 先对 pathname `stat` 并检查 `fresh.size <= MAX_MULTIMODAL_DATA_SIZE`，随后 `readImagePathBytes()` 再按同一 pathname 调用无界 `fs.readFile()`。两步之间文件或 symlink target 可以被替换成大文件；Node 会先为整个新文件分配 Buffer，之后 `bytesToUri()` 才按 `bytes.length` 拒绝。结果是攻击者可绕过“读取前”的内存保护，造成 collector 内存峰值甚至 OOM；UNC/FUSE 上还会与无 deadline 问题叠加。

建议用同一个已打开 handle 完成 `lstat/open(O_NOFOLLOW) -> fstat -> bounded read(cap + 1)`，并在读取后再次校验类型、size 和 magic MIME；不要在检查后重新按 pathname 打开。

src/inputs/qoder-trace/qoder-cli-multimodal.ts
yyuuttaaoo: 操作方式校正：本机 macOS 上这条真实图片输入是通过 **`Ctrl+V`** 粘贴的，不是 `Alt+V`。这不改变上面的 carrier 结论，但 PR 当前 `docs/multimodal.md:128` 和 `docs/zh-CN/multimodal.md:128` 都写成了 `Alt+V paste`，与实际验证操作不一致。建议文档改为平台准确的 `Ctrl+V`，或只写“paste image”而不把未验证的快捷键固化成契约。

src/types/index.ts
```diff
@@ -196,7 +196,7 @@ export interface MultimodalRuntimeConfig {
 }
 
 /** Agent ids with multimodal extraction implemented. */
-export const MULTIMODAL_SUPPORTED_AGENT_IDS = ['codex'] as const;
+export const MULTIMODAL_SUPPORTED_AGENT_IDS = ['codex', 'qoder'] as const;
```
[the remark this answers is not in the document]
xiongyunn: `qoder` 多模态只覆盖桌面 IDE / CLI；`qoder-idea` 是 JetBrains，内容策略按设计走独立的 `agents.qoder-idea`，不继承 `agents.qoder`。本 PR 也不做 JetBrains 多模态。

src/inputs/qoder-trace/qoder-cli-multimodal.ts
xiongyunn: 1.1.28 粘贴的可见 user / last-prompt 只有 `[Image #0]`，本地路径在同 prompt 的 isMeta `[Image: source: ~/.qoder/tmp/<slug>/images/<session>/N-clipboard-*.png.png]`。生产关联应继续走这条 sidecar（Hook 转发 → Input 提取），不要用 tmp 时间猜文件。
此外，后续在测试 1.1.28 时发现：`@` / `--attachment` 会额外写入 `type: "attachment"` / `image_file`，Pilot 已补上对应提取（Hook 原样转发本轮 `image_file` → Input 与 `[Image: source:]` / `@` 合并去重）。

src/inputs/qoder-trace/qoder-ide-multimodal.ts
xiongyunn: Qoder是Stop才写 hook，这时sqlite已经落盘。查空按没附件缓存，避免同一 request_id 反复打库。

assets/hooks/qoder-hook-processor.mjs
xiongyunn: 当前qoder cli transcript提取这边不处理反序，Qoder 官方写的是按发生顺序往文件末尾追加。当前粘贴是先 prompt 后图，会留在同一轮，没观察到反序，暂不改。

xiongyunn: 之前 isMeta 是切开保留的，没有丢。多模态为了拿图才和 prompt 合在一轮。只留 `isMeta && [Image: source:]` 会删掉其它 isMeta；只让 image 跟轮也不行——中间若有 Caveat 开新轮，图会落到 Caveat 上。先维持现状，确认 caveat 污染后再统一丢。

src/inputs/qoder-trace/qoder-trace-input.ts
```diff
@@ -32,10 +77,26 @@ export class QoderTraceInput extends BaseInput {
...
+    this.multimodalEnabled = opts.multimodal?.enabled === true && !!opts.multimodal.processor;
+    this.multimodalUploadMode = opts.multimodal?.uploadMode ?? 'none';
+    this.multimodalProcessor = opts.multimodal?.processor ?? null;
+    this.allowedRootPaths = this.multimodalEnabled
+      ? resolveQoderAllowedRootPaths(opts.multimodal?.allowedRootPaths)
+      : [];
+  }
+
+  override async stop(): Promise<void> {
+    this.multimodalStopped = true;
```
yyuuttaaoo, line 106: **[Medium]** Multimodal stays disabled after this input is stopped and started again.

**Impact:** `stop()` sets `multimodalStopped=true`, but `onStart()` never resets it. AgentDiscovery reuses the same input instance across disable/enable or temporary-unavailability transitions, so text collection restarts while both multimodal loops keep breaking and `pathToUri` returns null forever.

**Suggestion:** Reset the stop flag at the start boundary and add a start-stop-start regression test that verifies image enrichment resumes.

src/multimodal/resolve.ts
```diff
@@ -45,35 +106,184 @@ export function mimeFromImagePath(filePath: string): string | null {
...
-    const bytes = await fs.readFile(stated.resolvedPath);
-    return { bytes, mime_type: stated.mime_type, size: bytes.length };
+    fh = await fs.open(resolvedPath, flags);
   } catch {
     return null;
   }
+  try {
+    const st = await fh.stat();
+    if (!st.isFile() || st.size <= 0) return null;
+    const toRead = Math.min(st.size, maxBytes + 1);
```
yyuuttaaoo, line 302: **[Medium]** A single `FileHandle.read` is accepted as the complete image even when it returns a short read.

**Impact:** Node does not guarantee that one read fills the requested buffer, especially on FUSE or remote filesystems. Any positive prefix with valid image magic is hashed and uploaded as though complete; an in-place size change can similarly produce a truncated or unrenderable object.

**Suggestion:** Loop bounded reads until the expected fstat size or EOF, then fstat again and reject or retry when size, mtime, or identity changed. Cover a mocked short read and concurrent size change.

src/multimodal/processor.ts
```diff
@@ -63,29 +107,57 @@ export class MultimodalProcessor {
...
     if (this.shuttingDown) {
       logger.warn('multimodal pathToUri rejected', { reason: 'shutting_down' });
       return null;
     }
 
-    const trimmed = filePath.trim().split(/[?#]/)[0] ?? '';
-    if (!trimmed || !isImageFilePath(trimmed)) return null;
-    const key = path.resolve(trimmed);
+    const deadlineMs = opts?.deadlineMs ?? PATH_TO_URI_DEADLINE_MS;
+    return withDeadline(
```
yyuuttaaoo, line 123: **[Medium]** The deadline detaches the caller but does not cancel the underlying filesystem work.

**Impact:** A timed-out conversion can later call `bytesToUri` and upload an orphan object after the event and checkpoint advanced without its URI. Permanently hung work also remains alive; initial `lstat` is before inflight admission, so those hangs are not covered by the 1,024-entry cap. Ten serial paths can still delay one carrier by about 30 seconds.

**Suggestion:** Admit before every filesystem await, pass real cancellation through realpath/open/read/close or use a killable helper, and enforce a per-turn or batch deadline. Test that completion after timeout never uploads.

src/multimodal/resolve.ts
```diff
@@ -45,35 +106,184 @@ export function mimeFromImagePath(filePath: string): string | null {
...
-export async function readImagePathBytes(stated: PathStat): Promise<PathBytes | null> {
+/** Open and read an already-normalized local image path. */
+export async function openNormalizedLocalImage(
+  resolvedPath: string,
+  maxBytes = MAX_MULTIMODAL_DATA_SIZE,
+  allowedRootPaths?: string[],
+): Promise<(PathBytes & { mtimeMs: number; resolvedPath: string }) | null> {
+  if (allowedRootPaths) {
+    if (allowedRootPaths.length === 0) return null;
+    if (!(await isRealPathInsideRoots(resolvedPath, allowedRootPaths))) return null;
```
yyuuttaaoo: **[High]** The allowed-root check can be raced because containment is verified by pathname before the file is opened.

**Impact:** `O_NOFOLLOW` protects only the final component. In a local macOS PoC, atomically swapping an allowed parent symlink after `realpath` validation caused this sequence to open and read an outside-root PNG after two flips. A process controlling a directory below an allowed root can therefore bypass the intended exfiltration boundary.

**Suggestion:** Bind authorization to the opened object: compare the opened handle identity with a post-open path stat/realpath and reject any mismatch, or isolate the read in a killable trusted helper. Add a parent-symlink-swap regression test.

src/multimodal/resolve.ts
```diff
@@ -45,35 +106,184 @@ export function mimeFromImagePath(filePath: string): string | null {
...
-export async function statImagePath(filePath: string): Promise<PathStat | null> {
-  const mime = mimeFromImagePath(filePath);
-  if (!mime) return null;
+/** UNC or Windows device path. */
+export function isUncOrDevicePath(filePath: string): boolean {
+  const trimmed = filePath.trim();
+  if (!trimmed) return false;
+  const win = trimmed.replace(/\//g, '\\');
+  if (/^\\\\[.?]\\UNC\\/i.test(win)) return true;
+  if (/^\\\\[.?]\\/i.test(win)) return false;
```
yyuuttaaoo: **[Medium]** This branch explicitly treats Windows device namespaces as safe paths.

**Impact:** Non-UNC `\\\\?\\...` and every `\\\\.\\...` value return false from `isUncOrDevicePath`, although the documentation says device paths are skipped. Values such as GLOBALROOT or pipe/device paths therefore reach `lstat` before allowed-root containment. The visible CI matrix does not execute these Windows branches.

**Suggestion:** Reject device, GLOBALROOT, and pipe namespaces before any filesystem call. If extended drive paths must be supported, strip and canonicalize only the `\\\\?\\C:\\...` form before normal containment, with Windows tests.

assets/hooks/qoder-hook-processor.mjs
```diff
@@ -1371,6 +1372,49 @@ function extractToolResults(rows) {
...
+  if (typeof att.mediaType === 'string' && att.mediaType.trim()) {
+    slim.mediaType = att.mediaType.trim();
+  }
+  return slim;
+}
+
+function cliAttachmentFields(agentType, userRow, allParsed) {
+  if (agentType !== 'qoder-cli') return {};
+  const attachments = collectTurnImageFileAttachments(allParsed, userRow);
+  return attachments.length > 0 ? { 'agent.qoder.attachments': attachments } : {};
```
yyuuttaaoo, line 1411: **[High]** The new attachment carrier persists an absolute local path outside the content-privacy policy.

**Impact:** `agent.qoder.attachments[].filename` is added even when multimodal is `none`. `applyAgentContentPolicy` does not remove this field when `captureMessageContent:false`, so usernames, workspace names, and filenames are still emitted despite the privacy setting. Successful enrichment also leaves the carrier on the event.

**Suggestion:** Treat this as transient extractor state and delete it before entries leave `QoderTraceInput`, or include it in content policy and masking. Add regression cases for multimodal none plus capture false.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: Round 2 status: **OPEN**. The new outer catch does not observe the real failure path because `readAttachedImagePathsForRequestIds` still catches each DB error and returns an ordinary empty Map. The caller then caches every missing request as `[]`, so transient query failure and late rows remain permanently suppressed.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
```diff
@@ -0,0 +1,382 @@
...
+    if (!result) {
+      stats.skipped += 1;
+      continue;
+    }
```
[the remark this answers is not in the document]
yyuuttaaoo: Round 2 status: **fixed**. Each IDE `pathToUri` call is caught independently, so one rejection no longer aborts later images or the metadata pass.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: Round 2 status: **PARTIAL; keep open**. Allowed roots, static symlink/ordinary UNC rejection, magic MIME, and bounded same-handle reads are good improvements. However containment is still checked before open. I reproduced a parent-directory symlink swap that read an outside-root PNG, so the authorization boundary is still bypassable. Windows device namespaces also reach filesystem access. I added focused follow-up comments on the new head.

src/inputs/qoder-trace/qoder-cli-multimodal.ts
yyuuttaaoo: Round 2 status: **fixed**. Captures, extracted paths, and conversion attempts are bounded at 10. Re-running the original adversarial shapes on Node 24 is now linear: about 0.2/22 ms at 80 KB and 0.4/43 ms at 160 KB for CLI/IDE.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: Round 2 status: **fixed**. IDE output extraction now resolves relative Markdown destinations against `agent.qoder.cwd`; the workspace still needs to be in configured allowed roots, as documented.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
```diff
@@ -0,0 +1,382 @@
...
+import { LruMap, MULTIMODAL_LRU_LIMIT } from '../../multimodal/uploader/lru-set.js';
+import { createLogger } from '../../utils/logger.js';
+import { readAttachedImagePathsForRequestIds } from './sqlite-token-reader.js';
+
+const logger = createLogger('QoderIdeMultimodal');
+
+// Path patterns found in current Qoder IDE surfaces; extend when new scenes appear.
+const IMAGE_FILE_RE = /Image file:\s*([^\n\r]+)/gi;
+const IMAGE_GEN_PATH_RE = /The absolute path of the image is:\s*([^\n\r]+)/gi;
+const MARKDOWN_IMAGE_RE = /!\[[^\]]*]\(([^)\s]+)\)/g;
```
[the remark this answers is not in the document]
yyuuttaaoo: Round 2 status: **fixed**. The bounded Markdown parser now accepts angle-bracket destinations with spaces and optional titles, with regression coverage.

src/inputs/qoder-trace/qoder-cli-multimodal.ts
yyuuttaaoo: Round 2 status: **PARTIAL; keep open**. The new `attachment/image_file` carrier targets Qoder CLI 1.1.29, but the test is synthetic and uses an `@path` prompt rather than a captured Ctrl+V fixture. The 1.1.28 sidecar described in the reply uses a literal `~/.qoder/...`; current `resolveImagePath` does not expand tilde and instead resolves it below `agent.qoder.cwd`. The historical local Ctrl+V transcript still has only `[Image #0]`.

src/inputs/qoder-trace/qoder-cli-multimodal.ts
```diff
@@ -0,0 +1,319 @@
...
+}
+
+function stripSizeSuffix(raw: string): string {
+  return raw.replace(SIZE_SUFFIX_RE, '').trim();
+}
+
+function stripImageSourcePath(raw: string): string {
+  const comma = raw.indexOf(',');
+  return (comma === -1 ? raw : raw.slice(0, comma)).trim();
+}
```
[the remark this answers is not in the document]
yyuuttaaoo: Round 2 status: **fixed**. Cleanup now removes only the known `, original ...` suffix; comma-containing filenames are preserved and tested.

assets/hooks/qoder-hook-processor.mjs
yyuuttaaoo: Round 2 status: **OPEN**. Production `splitContentEventsIntoTurns` still emits a leading meta-only turn, and `buildUserMessageParts` still appends all non-tool user metadata. The new meta-first test calls `buildEventsFromBoundaries` directly; its splitter assertion only covers normal prompt-then-meta order, so it bypasses the failing production sequence.

src/types/index.ts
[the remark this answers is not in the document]
yyuuttaaoo: Round 2 status: **false-positive under the clarified contract**. The updated implementation explicitly skips `qoder-idea` multimodal sessions, and the agent docs state that JetBrains content policy is independently configured as `agents.qoder-idea`. This PR no longer relies on inheriting `agents.qoder`.

src/inputs/qoder-trace/qoder-trace-input.ts
yyuuttaaoo: Round 2 status: **PARTIAL; keep open**. A three-second caller deadline prevents an infinite wait, but it does not cancel the underlying filesystem operation. Ten serial paths can still delay one carrier around 30 seconds; hung initial stats occur before inflight admission, and a late completion can upload an orphan object after the event advanced without its URI. I added a focused comment on the deadline implementation.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: Round 2 status: **fixed**. Event-local day is now included in both path cache and inflight keys. The regression test asserts distinct 20260824 and 20260825 URIs/uploads for the same unchanged file.

yyuuttaaoo:
## 🔍 Code Review Summary — Round 2

| Severity | Active count |
|----------|-------------:|
| Critical | 0 |
| High | 3 |
| Medium | 7 |
| Low | 1 |

### Prior comments

- Fixed: regex/attempt cap, cross-day cache, Markdown spaces/title/cwd, per-image fail-open, comma filenames.
- Scope clarified / false positive: `qoder-idea` uses independent content policy and is skipped by Qoder multimodal.
- Partially fixed: local-file security, CLI paste compatibility, filesystem timeout/cancellation.
- Not fixed: SQLite transient/late-row negative cache, leading `isMeta` splitting/meta pollution, nested SQLite context fallback.

### New blockers

1. Allowed-root authorization is still raceable through a parent-directory symlink swap; a local macOS PoC read an outside-root PNG.
2. The new `agent.qoder.attachments[].filename` field bypasses `captureMessageContent:false` and leaks absolute local paths.
3. Deadline only detaches the caller; underlying FS work can accumulate or upload orphan objects after timeout.
4. Qoder multimodal becomes permanently disabled after the input's first dynamic stop/restart.
5. A single short `FileHandle.read()` can accept and upload truncated image bytes.
6. Windows `\\?\` / `\\.\` device namespaces are not rejected before filesystem access.

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| Resource release | FAIL |
| Deadlock/normal collection stall | PARTIAL |
| State recovery | FAIL |
| Dynamic restart | FAIL |

### Validation

- Typecheck and build passed.
- Focused tests: 287/288 in the combined run; the one inflight admission test passed in 5/5 isolated reruns, indicating a test-order race.
- The old quadratic regex case is fixed: local Node 24 scaling is linear.
- The user's historical macOS Ctrl+V transcript remains unsupported; the new 1.1.29 image attachment path has no real local Ctrl+V fixture here.

### Overall conclusion

Please do not merge yet. The update fixes several important correctness/performance comments, but the reproducible allowed-root TOCTOU and the new absolute-path privacy bypass are blocking. Cancellation, restart state, and SQLite recovery also need another pass.

### Highlights

- Correct cross-day URI cache semantics.
- Bounded regex/candidate work.
- Better Markdown/path handling, per-image fail-open, magic MIME, and bounded allocation.

---
*Report: `code-review/pr-257/final-report.md`*
*Generated by LoongSuite-Pilot Code Review Agent — Round 2*

src/inputs/qoder-trace/qoder-cli-multimodal.ts
xiongyunn: 1.1.28 粘贴图不是从 [Image #0] 提的。用户能看见的那句只有占位，同一次 prompt 后面那条 isMeta user 里才有 [Image: source: …]，Hook 会把它折进这一轮，Input 再提这条路径。之前回复有误，[Image: source: …]包含的是绝对路径 `/Users/.../.qoder/tmp/...`不是相对路径`~`。本地 1.1.28 粘贴已验证通过。

src/inputs/qoder-trace/qoder-ide-multimodal.ts
xiongyunn: 改为：查到 `request_id` 才写入缓存；extra 没图记 `[]`。SQL 失败或库里没有这行不记，下次还会查。
此外：同一 `request_id` 只会落在一个库，Hook 也是 Stop 之后才写、extra 这时已经在盘上，所以不存在跨库覆盖或先缓存空再丢图，这两条目前不考虑。

assets/hooks/qoder-hook-processor.mjs
xiongyunn: 真实 CLI 粘贴是先 user prompt，再同轮 `isMeta` 的 `[Image: source:]`。Hook 按这个顺序组轮，本地 1.1.28 已按这个验过。先 `isMeta` 再 prompt 的失序不在范围内，不做 `pendingPrelude`，也不按这个顺序写契约测试。`buildUserMessageParts` 会保留本轮其它 `isMeta`，不只留 image sidecar。因为在多模态采集之前，这些数据也都是保留了的。

src/multimodal/resolve.ts
yyuuttaaoo: Round 3 status: **fixed**. Device, pipe, GLOBALROOT, UNC, and non-drive extended namespaces are rejected before filesystem access. The explicitly supported `\\?\C:\...` form is reduced to an ordinary local-drive path and goes through normal containment.

assets/hooks/qoder-hook-processor.mjs
yyuuttaaoo: Round 3 status: **OPEN / not fixed**. This hook file is unchanged. `splitContentEventsIntoTurns([meta, prompt])` still emits a leading meta-only turn, and `buildUserMessageParts` still appends all non-tool user metadata. I kept this open under the previously stated recovery/bootstrap impact.

assets/hooks/qoder-hook-processor.mjs
yyuuttaaoo: Round 3 status: **PARTIAL; keep open**. The default `QoderTraceInput` path now deletes `agent.qoder.attachments` before emit in both multimodal-off and on modes. The supported fallback still leaks it: when `qoder-trace` is disabled, `QoderCliInput` uses `buildCanonicalHookEntry`, which copies every `agent.*` field, so the absolute filename is emitted even with content capture disabled. A focused production-path test reproduced the field on the fallback entry.

src/multimodal/processor.ts
yyuuttaaoo: Round 3 status: **PARTIAL; keep open**. Late completion no longer reaches `bytesToUri`, so the orphan upload is fixed. The filesystem work itself is still not truly cancellable, admission still follows the initial `lstat`, and shutdown neither aborts nor waits for path reads. The per-carrier serial deadline budget also remains about 30 seconds.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: Round 3 status: **PARTIAL; keep open**. Windows device namespaces, bounded same-handle reads, magic MIME, and static path checks are fixed. However, post-open `realpath(path)` containment and the later `stat(path)` identity check are still separate pathname observations. A deterministic ABA test opened the outside PNG, switched inside for `realpath`, then outside again for `stat`; the outside handle passed and its bytes were read.

src/inputs/qoder-trace/qoder-cli-multimodal.ts
yyuuttaaoo: Round 3 status: **PARTIAL; keep open**. The clarification that the 1.1.28 sidecar is absolute removes the reported tilde issue for that observed shape, and 1.1.29 `image_file` extraction exists. On this machine, though, the real historical macOS Ctrl+V record still has only `[Image #0]`, and current 1.1.29 transcripts contain no real `image_file` Ctrl+V row. The new tests are still synthetic, so the common paste path is not locally proven.

src/multimodal/resolve.ts
yyuuttaaoo: Round 3 status: **fixed**. The implementation loops reads to the expected length/EOF and then validates size, mtime, and actual byte length with a second handle stat. Short-read, premature-EOF, and concurrent-size-change tests pass.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: Round 3 status: **PARTIAL; keep open**. Missing rows and query failures are no longer negative-cached. Recovery is still absent at the collection boundary: `readHookJsonl()` advances the offset before enrichment, then emits the text-only event. Without a second source event containing the same request ID, the next collect has nothing to query. A deterministic end-to-end probe observed one lookup, an advanced checkpoint, and no second lookup after the row became available; the new unit test proves only a synthetic duplicate event.

src/inputs/qoder-trace/qoder-trace-input.ts
yyuuttaaoo: Round 3 status: **PARTIAL; keep open**. The timeout token now prevents a late `bytesToUri` call/upload, which fixes the orphan-object portion. It still does not cancel `lstat`/`realpath`/`open`/an in-progress read; initial `lstat` remains before inflight admission, shutdown does not join path reads, and ten serial paths can still delay a carrier by about 30 seconds.

src/multimodal/resolve.ts
yyuuttaaoo: Round 3 status: **PARTIAL; keep open**. The new identity comparison rejects the single-swap case in the added test, but it is not atomic: `realpath(path)` and `stat(path)` are separate awaits. I reproduced an ABA sequence that shows the safe parent during `realpath` and the outside parent during `stat`; `stat` then matches the already-open outside handle and the outside PNG is accepted.

src/inputs/qoder-trace/qoder-trace-input.ts
yyuuttaaoo: Round 3 status: **fixed**. `onStart()` now resets `multimodalStopped`, and the start-stop-start regression verifies that image enrichment succeeds again on the second start.

yyuuttaaoo:
## Code Review Summary - Round 3

Current head: `ec9731c6`

### Verdict

Please do not merge yet. This update fixes several round-2 findings, but two High security/privacy boundaries remain only partially fixed.

### Fixed in this update

- stop/start now restores multimodal enrichment;
- short reads and files changed during read are rejected;
- Windows device/pipe/GLOBALROOT namespaces are rejected;
- a timed-out read cannot later call `bytesToUri` and upload an orphan object;
- the default `QoderTraceInput` path deletes `agent.qoder.attachments` before emit;
- SQLite misses and query failures are no longer negative-cached.

### Still open

1. **High - parent-directory ABA TOCTOU.** Post-open `realpath(path)` and `stat(path)` are separate pathname observations. A deterministic test switched the parent inside for containment and outside again for identity; the opened outside PNG passed and was read.
2. **High - fallback attachment privacy.** Disabling `qoder-trace` enables `QoderCliInput`, whose canonical path still emits `agent.qoder.attachments[].filename` even when multimodal/content capture is off.
3. **Medium - SQLite recovery.** A miss is uncached, but the hook checkpoint has already advanced; without a duplicate request event there is no later retry.
4. **Medium - filesystem cancellation.** Late upload is fixed, but the underlying `lstat/realpath/open/read` can remain alive, admission still follows initial `lstat`, and one carrier can still spend about 30 seconds on ten serial deadlines.
5. **Medium - leading `isMeta`.** Production turn splitting and broad metadata retention are unchanged.
6. **CLI Ctrl+V evidence.** The local historical Ctrl+V record still has only `[Image #0]`; local 1.1.29 transcripts contain no real `image_file` Ctrl+V fixture, so this stays partially verified.

### Validation

- 199/199 focused tests passed.
- Typecheck and full build passed.
- `git diff --check` passed.
- Deterministic probes reproduced the ABA escape, fallback carrier leak, and lack of SQLite retry after checkpoint.
- GitHub Node 18/20/22 CI and Secret Scan are green.

Detailed report: `code-review/pr-257/final-report.md`

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: **fixed for the material High-risk path; residual downgraded to Low.** Root containment, final-component symlink checks, UNC/device rejection, magic MIME, bounded same-handle reads, and ordinary swap detection are now present. A multi-flip parent ABA remains technically reproducible, but it requires an active same-user local process that already has equivalent file access; Pilot does not run with higher privilege. I no longer consider that residual a merge blocker.

src/inputs/qoder-trace/qoder-cli-multimodal.ts
yyuuttaaoo: **fixed — validated with a real macOS Qoder CLI 1.1.29 Ctrl+V session.** Session `cd0a74fd-f32c-4f66-bdfd-a229911f7f81` writes the human `[Image #0]` row and then an `isMeta: true` `[Image: source: <absolute path>]` row under the same `promptId`. Replaying it through this HEAD produced one request with both parts; the real 142093-byte PNG generated the expected dated SHA-256 URI and exactly one upload.

src/inputs/qoder-trace/qoder-trace-input.ts
yyuuttaaoo: **fixed for the reported operational failure.** Each path now has a 3-second caller deadline, timeout is fail-open, late upload is suppressed, and stop gates subsequent paths, so shutdown does not wait `10 x 3s`. A native fs request on a pathological FUSE mount may remain cooperatively cancellable only; that is Low hardening outside the normal local-root path.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: **fixed for the reported permanent negative-cache bug.** Misses and query failures are no longer cached. A narrow IDE DB-write timing window can still omit only the image URI for one event, but text, checkpoint, and later turns remain correct; Qoder CLI Ctrl+V does not use SQLite. I am treating that as a non-blocking completeness residual.

assets/hooks/qoder-hook-processor.mjs
yyuuttaaoo: **fixed for the supported current-client path.** The real Qoder CLI 1.1.29 Ctrl+V order is prompt first, then the `isMeta` image sidecar under the same `promptId`; HEAD and its regression tests keep them in one turn. A leading-meta-only condition requires an abnormal truncated/recovery window and should not block the normal Ctrl+V implementation.

src/multimodal/processor.ts
yyuuttaaoo: **fixed for the material impact.** Completion after timeout can no longer call `bytesToUri` or enqueue an orphan upload. Native filesystem operations are checked cooperatively around awaits rather than forcibly cancelled, but with default local home roots that is Low hardening rather than a merge blocker.

src/multimodal/resolve.ts
yyuuttaaoo: **partially fixed, but severity recalibrated to Low defense-in-depth.** Stable parent symlinks and ordinary single-swap races are rejected. The remaining multi-flip ABA requires an active same-user process repeatedly changing the parent across async checkpoints; Pilot runs with the same user privileges, so this does not establish the High exfiltration boundary originally assumed.

assets/hooks/qoder-hook-processor.mjs
yyuuttaaoo: **partially fixed; remaining scope is Medium and configuration-limited.** Default `QoderTraceInput` strips the carrier in multimodal on/off modes. The absolute filename remains only when an operator explicitly disables default `qoder-trace`, keeps `qoder-cli-hook` fallback, and sets `captureMessageContent=false`. A small unified strip is recommended, but this is not a default-path merge blocker.

yyuuttaaoo:
## Code Review Round 4 - Materiality Recalibration

After validating the current head against a real macOS Qoder CLI 1.1.29 Ctrl+V session, I no longer see a merge blocker.

### Real Ctrl+V validation

Session `cd0a74fd-f32c-4f66-bdfd-a229911f7f81` contains the real supported carrier: the human `[Image #0]` prompt is followed under the same `promptId` by an `isMeta: true` `[Image: source: <absolute ~/.qoder/tmp path>]` row. Replaying those records through PR HEAD produced one request containing both parts. The real 142093-byte PNG was read successfully, converted to the expected `20260827/<sha256>.png` URI, and enqueued exactly once. The CLI paste happy path is fixed.

### Remaining material issue

One configuration-limited Medium issue remains: if an operator explicitly disables the default `qoder-trace`, keeps the legacy `qoder-cli-hook` fallback, and sets `captureMessageContent=false`, the fallback can still emit `agent.qoder.attachments[].filename`. The default trace path already strips it. A small unified carrier-strip test/fix is recommended, but this is not a default-path blocker.

### Accepted residuals

- The multi-flip parent-symlink ABA requires an active same-user process and does not cross Pilot's privilege boundary; treat it as Low defense-in-depth.
- Native filesystem cancellation and serial deadline worst cases require pathological FUSE/mount behavior. Deadline, stop gating, and late-upload suppression resolve normal operational risk.
- A first SQLite miss can omit one IDE image URI in a narrow timing window, but text, checkpoint, and later collection remain correct; CLI Ctrl+V does not use SQLite.
- Leading `isMeta` and nested SQLite context are not failures on the real current-client paths observed locally.

### Verdict

| Check | Result |
|---|---|
| Merge blockers | 0 |
| Normal Qoder CLI Ctrl+V | PASS |
| Resource/lifecycle behavior | PASS |
| State recovery | PASS |

Focused round-4 validation: 119 tests passed, including real local media conversion. Prior exact-head validation (199 focused tests, typecheck, full build, and GitHub CI) remains green.

---
*Detailed report: `code-review/pr-257/final-report.md`*

yyuuttaaoo (approved):
Approved at head `df20e53bf7515a5ad4a8aee836453aead5f46085`.

Verified the two final fixes:
- legacy `qoder-cli-hook` fallback strips `agent.qoder.attachments`, including when `captureMessageContent=false`;
- Qoder IDE SQLite attachment lookup retries an initially absent row within the same event with bounded 25 ms / 75 ms delays.

Focused regression: 196/196 passed; all Qoder + multimodal suites: 415/415 passed; typecheck, build, diff check, and GitHub CI all passed. No remaining Critical/High/Medium findings.

yyuuttaaoo (approved):
Re-approved at identity-corrected head `ec62b6b92134d70d946b9c385fa447083c708212`.

The commit tree is identical to the reviewed `df20e53b`; only author/committer metadata was corrected to `Tom Yu <yyuuttaaoo@gmail.com>`. GitHub maps both to `yyuuttaaoo`. All CI matrices, Secret Scan, and CLA passed on this head. The two requested fixes remain verified with no Critical/High/Medium findings.

src/inputs/qoder-trace/qoder-ide-multimodal.ts
yyuuttaaoo: Fixed at `ec62b6b9`. Missing request IDs are no longer negative-cached: the same event performs one immediate lookup plus bounded retries after 25 ms and 75 ms, preserves partial successes, and caches only IDs actually present in the returned Map (including authoritative `[]`). The absent-then-present regression verifies the recovered URI is attached before emit and checkpoint save.

src/multimodal/processor.ts
yyuuttaaoo: The material issue is fixed: timeout aborts the conversion token, late completion cannot reach `bytesToUri` or enqueue an orphan upload, and the caller remains bounded by the per-path deadline. **TODO (non-blocking):** if slow FUSE/network roots are supported later, admit before the first filesystem await and track active `AbortController`s so shutdown can explicitly abort/join outstanding path reads. The remaining native-fs limitation is Low hardening for the current local-root scope.

src/multimodal/resolve.ts
yyuuttaaoo: The original High-risk single-swap/handle-mismatch path is fixed. **TODO (non-blocking):** when suitable platform primitives are available, use handle-relative/openat2-style reads so containment is bound to the opened object, eliminating the residual ABA that requires an active same-user process to switch a parent multiple times across async checkpoints. Stable symlinks, final symlinks, and ordinary single swaps are already rejected; the remainder is Low defense-in-depth.

assets/hooks/qoder-hook-processor.mjs
yyuuttaaoo: Fixed on the current head. The default `QoderTraceInput` path strips the transient carrier before emit, and the legacy `QoderCliInput` fallback now also unconditionally deletes `agent.qoder.attachments` before returning canonical records. The fallback regression verifies that neither the field nor its absolute `/Users/alice/...` filename reaches output when `captureMessageContent=false`.

linrunqi08 (approved):

PR #280, opened 2026-08-18
feat(deploy): add a synchronous `loongsuite-pilot deploy` command

linrunqi08:
## Why

Hook and plugin deployment is currently only a side effect of `orchestrator.start()`. That works for a long-running daemon but is unusable for an image build:

- A `docker build` RUN layer reclaims the process as soon as the command returns, so whether hooks were written **before the layer was committed** is a race against the daemon's teardown.
- An agent that is not detected is a **silent skip** (`success: true, skipped: true`), so the image looks instrumented while nothing was written.

## What

`loongsuite-pilot deploy` runs the same deployment once, in the foreground, and exits with a code.

| Flag | Behaviour |
|---|---|
| `--require <ids>` | Turns "not detected" into exit 1, so a build layer fails loudly instead of producing an uninstrumented image. |
| `--json` | Machine-readable result; logs go to stderr so stdout stays parseable. |

Contract details that are load-bearing:

- **Idempotent.** An already-deployed agent stays green, so a derived image rebuilding on an instrumented base does not fail.
- **`--require ""` is rejected** rather than parsed as "require nothing", which would silently assert nothing.
- **A hard deploy failure is an error even when nothing was required**, so a caller that passes no `--require` still surfaces e.g. a plugin-extraction error.

To make `--require` decidable, `DeployResult` now carries a `DeploySkipReason`. `skipped` on its own cannot distinguish "the agent is not installed here" from "its integration is already in place" — opposite outcomes for a build gate. `detection-only` agents report `up-to-date`, since they share another agent's hook and never write anything.

Probe workers started during detection are stopped before returning: a `worker.pid` written on the build host can collide with an unrelated pid in a container's fresh PID namespace, which would make the collector believe a worker is already running and silently skip starting one.

`isAgentGatedEnabled` and `resolvePilotDir` are now shared with the orchestrator instead of duplicated — `orchestrator.ts` is 46 lines shorter.

The `scripts/loongsuite-pilot.{sh,ps1}` wrappers forward `deploy`, deliberately **without** restarting the collector afterwards: in a build layer nothing is running yet, and starting a daemon there would bake a half-written state file into the image. A collector already running on a host re-reads its deployment state on its own next cycle.

## Verification

| Check | Result |
|---|---|
| `tsc --noEmit`, `npm run build` | pass |
| `bash -n scripts/loongsuite-pilot.sh` | pass |
| `vitest run tests/unit/deployment/` | 292 passed, 3 skipped |
| Full unit suite | No new failures. The suite has pre-existing flakiness under parallel load; every file that differed from a pristine `main` run passes in isolation, and `main` itself fails a partially overlapping set. |

### DIFF
--- a/scripts/loongsuite-pilot.ps1
+++ b/scripts/loongsuite-pilot.ps1
@@ -1248,6 +1248,36 @@ function Cmd-Log {
 }
 
 # ============================================================
+# CMD: deploy (one-shot hook/plugin deployment, for image builds)
+# ============================================================
+function Cmd-Deploy {
+    $versionDir = Resolve-CurrentVersion
+    if (-not $versionDir) {
+        Write-Error "Current loongsuite-pilot version not found"
+        exit 1
+    }
+
+    $entry = Join-Path $versionDir "dist\index.js"
+    if (-not (Test-Path $entry -PathType Leaf)) {
+        Write-Error "Deploy CLI entrypoint missing"
+        exit 1
+    }
+
+    $nodeBin = Resolve-Node
+    if (-not $nodeBin) {
+        Write-Error "node runtime not found"
+        exit 1
+    }
+
+    $env:LOONGSUITE_PILOT_DATA_DIR = $DATA_DIR
+    $env:LOONGSUITE_PILOT_CACHE_DIR = $CACHE_DIR
+    $env:AGENT_DATA_COLLECTION_CONFIG = $CONFIG_FILE
+    # No collector restart afterwards -- see the shell cmd_deploy for why.
+    & $nodeBin $entry "deploy" @SubArgs
+    exit $LASTEXITCODE
+}
+
+# ============================================================
 # CMD: token-usage (foreground token usage CLI)
 # ============================================================
 function Cmd-TokenUsage {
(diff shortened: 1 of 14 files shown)


linrunqi08:
Added a second, independent commit: `fix(metrics): stop alarming about an updater that was never started`.

`buildAutoUpdateConfig` gates `enabled` on `hasPackageConfig`, so an install with no package source configured resolves auto-update to disabled — and the updater binary exits immediately on that config. But `collectInfraHealth` read the updater pid regardless, so from L1 cycle 3 on such an install reports `updater_pid_alive=false` and fires `UPDATER_NOT_RUNNING_ALARM` (level 3) roughly 30 minutes after start, about a process that was never supposed to run.

The probe is now gated on the same condition `orchestrator.start()` already uses at step 11 to decide whether to run `UpdaterWatchdog`. `autoUpdateEnabled` defaults to `true` when omitted, so a caller that knows nothing about auto-update keeps the previous behaviour.

`tests/unit/core/` 349 passed, `tests/unit/metrics/ tests/unit/deployment/` 370 passed / 3 skipped, `tsc --noEmit` and `npm run build` clean.

messixukejia (approved):

PR #309, opened 2026-08-21
fix(installer): install on non-ASCII Windows paths, and stop claiming success

linrunqi08:
## What

Windows installs under a non-ASCII profile (`C:\Users\<CJK name>`) failed, and hooks/skills/plugins were never deployed on **any** Windows box. Both failures were silent — the installer printed "Hook scripts deployed" either way.

Measured on a CJK-profile Windows box with PowerShell 5.1 and the bundled node v22.22.2/v22.22.3.

## Three encoding boundaries (PowerShell 5.1)

- **Native stdout is decoded through the console codepage.** `whoami` came back as `host\??` on an en-US box, so anything deriving identity from it was already broken. `Get-PilotAccountName` / `Get-PilotUserTag` read the account from .NET instead of parsing a mangled string. `chcp 65001` and `[Console]::OutputEncoding = UTF8` do **not** fix this — they govern the console, not argv conversion.
- **bsdtar receives argv through the ANSI CRT.** `%SystemRoot%\System32\tar.exe` converts a UTF-16 path back through the machine's ANSI codepage, so every character that page cannot represent arrives as a literal `?` and both `-f` and `-C` point at paths that do not exist. Archives now stage under an ASCII temp root and get renamed into place. The renamed tree needs its ACL re-inherited (`icacls /reset`): a rename carries the source's ACEs along, so a tree staged in `%SystemRoot%\Temp` lands in the profile where Users get write and traverse but *not* read — and node then reports the very misleading `Cannot find package 'pino'` even though the file is right there.
- **`Get-Content`/`Set-Content` default to ANSI**, and `-Encoding UTF8` always writes a BOM on 5.1 (there is no `utf8NoBOM`). Config payloads round-trip through a UTF-8 file passed as `argv[1]`, and the reading side trims `U+FEFF` explicitly — it is not whitespace in PowerShell, so a plain `.Trim()` leaves the BOM in the path and `Test-Path` returns false for a file that exists.

## Which tar

A bare `tar` resolves through `PATH`, and with Git for Windows installed that is Git's bundled GNU tar (MSYS), which reads the colon in `-f C:\...` as rsh `host:path` syntax and dies with `Cannot connect to C: resolve failed` (older builds hang). `Expand-PilotTarGz` prefers `System32\tar.exe`, then falls back to `PATH`'s tar with `--force-local`. Never both: bsdtar rejects the unknown flag outright, which is why `--force-local` is attached to the PATH candidate only. Same logic on the node side in `src/utils/win-archive.ts`.

## The silent half: `fs.cpSync` on Windows

`fs.cpSync` moved to a C++ `std::filesystem` implementation in Node 22.9, and that implementation fail-fasts on Windows: exit `0xC0000409` (`STATUS_STACK_BUFFER_OVERRUN`), **no JS exception, no stderr, nothing copied**. Reproduced on a two-file ASCII tree, so it is not about path length or non-ASCII paths. The async `fs.cp` is unaffected.

Because the process *dies* rather than throwing, `postinstall`'s own try/catch never ran, so everything after the first copy was skipped without a word — skill docs, agent plugins, the legacy intercept stub, the config migration. Downstream, a missing `plugins/` tree failed every dsh deployment with "plugin file not found or unreadable", once per collection cycle — and `DeploymentManager` only tallied that failure, so the reason never reached the log.

Fixed by copying with `mkdir`+`copyFile`, guarding each asset tree separately so one unwritable target cannot take the others down, and having the installers key their check mark off the exit code instead of printing it unconditionally. `DeploymentManager` now logs a strategy that *returns* `{success:false}` rather than only a thrown one.

## Two install-time races, same symptom

- A re-install no longer rewrites the live version directory in place. It used to `Remove-Item` and re-populate `versions/<ver>_<commit>/` while `current` still pointed there, so a collector launched in that window resolved `dist/index.js` against a directory whose `node_modules` were not back yet and died with `ERR_MODULE_NOT_FOUND`. Deploys now go to a fresh suffixed sibling, and `current` is written only after dependencies land.
- An upgrade now displaces a collector that survived it, via `restart-collector` rather than `restart` (the latter takes the updater down with it).

## Tests

New ratchets: `ps1-nonascii-windows-paths`, `postinstall-asset-trees`, `win-archive`, `installer-version-dir-not-overwritten`, `installer-restart-stale-collector`. The `.ps1` sweeps run over `git ls-files '*.ps1'` and the installer lists are `existsSync`-guarded, so they widen automatically rather than silently passing over an absent file — each sweep also asserts a floor so a matcher that stops matching fails instead of vacuously passing.

`tsc --noEmit` is clean. The 27 suites under `tests/unit/{scripts,deploy,updater,utils}` pass (387 tests). Remaining failures in a full `vitest run` are pre-existing on `main` (hook-processor suites, parallelism-sensitive); `tests/unit/scripts/daemon-cache-dir.test.mjs` fails on `main` and passes here.

### DIFF
--- a/deploy/installer-opensource.ps1
+++ b/deploy/installer-opensource.ps1
@@ -1075,6 +1075,10 @@ function Deploy-Package {
     $postinstallScript = Join-Path $script:PERMANENT_DIR "scripts\postinstall.js"
     $postinstallExit = 0
     $postinstallPartial = $false
+    # Empty means "nothing went wrong". The caller aborts on a non-empty value -- the policy
+    # lives at the call sites so this block stays byte-identical across the installers, same
+    # split as $script:PILOT_LAST_RESTART_ERR.
+    $script:PILOT_POSTINSTALL_ERR = ""
     if (Test-Path $postinstallScript) {
         # postinstall.js resolves its target as LOONGSUITE_PILOT_DATA_DIR, falling back to
         # $env:USERPROFILE\.loongsuite-pilot. Unset, an install started with -DataDir would
@@ -1093,7 +1097,7 @@ function Deploy-Package {
             $postinstallOut = & $script:NODE_BIN $postinstallScript 2>&1
             $postinstallExit = $LASTEXITCODE
             $postinstallOut | ForEach-Object { Write-Host $_ }
-            if (($postinstallOut | Out-String) -match "failed asset tree") { $postinstallPartial = $true }
+            if (($postinstallOut | Out-String) -match "failed asset tree|Post-install failed") { $postinstallPartial = $true }
         } finally {
             $ErrorActionPreference = $prevEAP
             if ($hadPostinstallDataDir) {
@@ -1108,18 +1112,25 @@ function Deploy-Package {
     # fs.cpSync fail-fasts with 0xC0000409 on the bundled Node, killing the script before it
     # copied anything, and this line printed "Hook scripts deployed" regardless.
     #
-    # Two distinct outcomes, and the exit code only covers the first. postinstall.js
+    # Three distinct outcomes, and the exit code only covers the first. postinstall.js
     # deliberately exits 0 after a per-tree failure -- a non-zero exit would abort the npm
-    # install fallback above -- and announces it as "N failed asset tree(s)" instead. So a
-    # non-zero code means the process died outright, while a partial failure is legible only
-    # in the output. Keying the check mark off the exit code alone printed it over an install
-    # that had just reported a tree it could not copy.
+    # install fallback above -- and announces it as "N failed asset tree(s)" instead; its
+    # top-level catch is the same story, printing "Post-install failed: <reason>" and still
+    # exiting 0. So a non-zero code means the process died outright, while the other two are
+    # legible only in the output. Keying the check mark off the exit code alone printed it over
+    # an install that had just reported a tree it could not copy.
+    #
+    # An install with no hooks/plugins is not an install: the collector finds nothing to load
+    # and every collection cycle fails on a plugin that was never written. The caller aborts
+    # (install) or rolls back (upgrade) on $script:PILOT_POSTINSTALL_ERR.
     if ($postinstallExit -ne 0) {
-        Msg "    ❌ Hook 脚本部署失败 (exit=$postinstallExit)，hooks/plugins 可能缺失" `
-            "    ❌ Hook script deployment failed (exit=$postinstallExit); hooks/plugins may be missing"
+        $script:PILOT_POSTINSTALL_ERR = "postinstall.js (exit $postinstallExit)"
+        Msg "    ❌ Hook 脚本部署失败 (exit=$postinstallExit)，hooks/plugins 缺失" `
+            "    ❌ Hook script deployment failed (exit=$postinstallExit); hooks/plugins are missing"
     } elseif ($postinstallPartial) {
-        Msg "    ❌ Hook 脚本部分部署失败（见上方 failed asset tree），hooks/plugins 可能缺失" `
-            "    ❌ Some hook asset trees failed to deploy (see 'failed asset tree' above); hooks/plugins may be missing"
+        $script:PILOT_POSTINSTALL_ERR = "postinstall.js reported a failed asset tree"
+        Msg "    ❌ Hook 脚本部分部署失败（见上方输出），hooks/plugins 缺失" `
+            "    ❌ Some hook asset trees failed to deploy (see the output above); hooks/plugins are missing"
     } else {
         Msg "    ✅ Hook 脚本已部署" "    ✅ Hook scripts deployed"
     }
@@ -2145,6 +2156,15 @@ function Cmd-Install {
         Prompt-UserId
         Confirm-ConfigOverwrite
         Deploy-Package $script:INSTALL_SRC
+        # Nothing downstream can repair this: postinstall.js is the only thing that fills
+        # $DataDir\{hooks,skills,plugins}, and a collector with no hooks collects nothing.
+        # Better to stop with the reason on screen than to finish and report success.
+        if ($script:PILOT_POSTINSTALL_ERR) {
+            Msg "❌ 安装中止：$script:PILOT_POSTINSTALL_ERR" `
+                "❌ Install aborted: $script:PILOT_POSTINSTALL_ERR"
+            Msg "   请检查上方输出后重试安装" "   Check the output above and re-run the install"
+            exit 1
+        }
         Write-Config
         Install-Command
 
@@ -2218,7 +2238,15 @@ function Cmd-Upgrade {
 
         Msg "==> 启动新版本..." "==> Starting new version..."
         $ps1Path = Join-Path $env:USERPROFILE ".local\bin\loongsuite-pilot-service.ps1"
-        $started = Start-PilotAndWait -ScriptPath $ps1Path -PriorVersion $oldVer
+        # A failed postinstall is a failed upgrade: no point starting the new version, and the
+        # rollback below is what keeps `current` on a version whose assets are known complete.
+        $started = $false
+        if ($script:PILOT_POSTINSTALL_ERR) {
+            Msg "    ❌ Hook 脚本部署失败: $script:PILOT_POSTINSTALL_ERR" `
+                "    ❌ Hook script deployment failed: $script:PILOT_POSTINSTALL_ERR"
+        } else {
+            $started = Start-PilotAndWait -ScriptPath $ps1Path -PriorVersion $oldVer
+        }
         if ($started) {
             Msg "    ✅ 新版本启动成功" "    ✅ New version started successfully"
             Write-Host ""
--- a/deploy/installer-opensource.ps1
+++ b/deploy/installer-opensource.ps1
@@ -1074,6 +1074,7 @@ function Deploy-Package {
     Msg "==> 部署 hook 脚本..." "==> Deploying hook scripts..."
     $postinstallScript = Join-Path $script:PERMANENT_DIR "scripts\postinstall.js"
     $postinstallExit = 0
+    $postinstallPartial = $false
     if (Test-Path $postinstallScript) {
         # postinstall.js resolves its target as LOONGSUITE_PILOT_DATA_DIR, falling back to
         # $env:USERPROFILE\.loongsuite-pilot. Unset, an install started with -DataDir would
@@ -1087,8 +1088,12 @@ function Deploy-Package {
         $prevEAP = $ErrorActionPreference; $ErrorActionPreference = "Continue"
         try {
             $env:LOONGSUITE_PILOT_DATA_DIR = $DataDir
-            & $script:NODE_BIN $postinstallScript
+            # Captured rather than streamed because the per-tree verdict is in the output and
+            # nowhere else (see below). 2>&1 because postinstall.js reports it on stderr.
+            $postinstallOut = & $script:NODE_BIN $postinstallScript 2>&1
             $postinstallExit = $LASTEXITCODE
+            $postinstallOut | ForEach-Object { Write-Host $_ }
+            if (($postinstallOut | Out-String) -match "failed asset tree") { $postinstallPartial = $true }
         } finally {
             $ErrorActionPreference = $prevEAP
             if ($hadPostinstallDataDir) {
@@ -1098,16 +1103,23 @@ function Deploy-Package {
             }
         }
     }
-    # Never print the check mark for a postinstall that did not finish. postinstall.js
-    # reports its own per-tree failures and still exits 0 on purpose (a non-zero exit
-    # would abort the npm install fallback above), so a non-zero code here means the
-    # process died outright. That is exactly how the hooks/skills/plugins trees went
-    # missing on every Windows install without a word: fs.cpSync fail-fasts with
-    # 0xC0000409 on the bundled Node, killing the script before it copied anything,
-    # and this line printed "Hook scripts deployed" regardless.
+    # Never print the check mark for a postinstall that did not finish. That is exactly how
+    # the hooks/skills/plugins trees went missing on every Windows install without a word:
+    # fs.cpSync fail-fasts with 0xC0000409 on the bundled Node, killing the script before it
+    # copied anything, and this line printed "Hook scripts deployed" regardless.
+    #
+    # Two distinct outcomes, and the exit code only covers the first. postinstall.js
+    # deliberately exits 0 after a per-tree failure -- a non-zero exit would abort the npm
+    # install fallback above -- and announces it as "N failed asset tree(s)" instead. So a
+    # non-zero code means the process died outright, while a partial failure is legible only
+    # in the output. Keying the check mark off the exit code alone printed it over an install
+    # that had just reported a tree it could not copy.
     if ($postinstallExit -ne 0) {
         Msg "    ❌ Hook 脚本部署失败 (exit=$postinstallExit)，hooks/plugins 可能缺失" `
             "    ❌ Hook script deployment failed (exit=$postinstallExit); hooks/plugins may be missing"
+    } elseif ($postinstallPartial) {
+        Msg "    ❌ Hook 脚本部分部署失败（见上方 failed asset tree），hooks/plugins 可能缺失" `
+            "    ❌ Some hook asset trees failed to deploy (see 'failed asset tree' above); hooks/plugins may be missing"
     } else {
         Msg "    ✅ Hook 脚本已部署" "    ✅ Hook scripts deployed"
     }
@@ -1489,15 +1501,41 @@ function Print-Summary {
 # $serviceEntry is whichever entry point the caller already resolved: the .ps1 needs
 # powershell.exe -File, the .cmd wrapper is invoked directly. Branching here rather than at
 # the call sites keeps this block byte-identical across the three installers.
+#
+# Returns $true when there is nothing to displace or the displacement succeeded, $false with
+# the reason in $script:PILOT_LAST_RESTART_ERR otherwise. Native commands do not throw on a
+# nonzero exit, so $LASTEXITCODE is the only signal there is -- and discarding it put the
+# failure right back where this block came from. Cmd-RestartCollector exits 1 when the node
+# runtime cannot be resolved, when the bootstrap entry is missing, and when the service
+# manager refuses to restart a non-degraded install; in every one of those the survivor is
+# still running the OLD version dir. Callers must not print success on a $false: `status`
+# resolves `current`, which the survivor answers to exactly as well as the new collector
+# would, so "is running" cannot tell the two apart. That is the entire bug.
 function Restart-StaleCollector {
     param([string]$serviceEntry, [string]$priorVersion)
-    if (-not $priorVersion) { return }
-    if (-not $serviceEntry) { return }
-    if (-not (Test-Path $serviceEntry)) { return }
-    if ($serviceEntry -match '\.ps1$') {
-        & powershell.exe -NoProfile -ExecutionPolicy Bypass -File $serviceEntry restart-collector 2>$null | Out-Null
-    } else {
-        & $serviceEntry restart-collector 2>$null | Out-Null
+    $script:PILOT_LAST_RESTART_ERR = ""
+    if (-not $priorVersion) { return $true }
+    if (-not $serviceEntry) { return $true }
+    if (-not (Test-Path $serviceEntry)) { return $true }
+    # Locally forced to Continue: with 2>&1 in effect a caller left on Stop turns the first
+    # stderr line into a NativeCommandError, which would report a successful restart as a
+    # failure. restart-collector writes progress to stderr on purpose.
+    $prevEAP = $ErrorActionPreference
+    $ErrorActionPreference = "Continue"
+    try {
+        if ($serviceEntry -match '\.ps1$') {
+            $restartOut = & powershell.exe -NoProfile -ExecutionPolicy Bypass -File $serviceEntry restart-collector 2>&1
+        } else {
+            $restartOut = & $serviceEntry restart-collector 2>&1
+        }
+        if ($LASTEXITCODE -eq 0) { return $true }
+        $script:PILOT_LAST_RESTART_ERR = "restart-collector (exit $LASTEXITCODE): $(($restartOut | Out-String).Trim())"
+        return $false
+    } catch {
+        $script:PILOT_LAST_RESTART_ERR = "restart-collector failed to run: $($_.Exception.Message)"
+        return $false
+    } finally {
+        $ErrorActionPreference = $prevEAP
     }
 }
 # <<< pilot-restart-stale-collector <<<
@@ -2058,14 +2096,19 @@ function Start-PilotAndWait {
     $startOutput = & powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -File $ScriptPath start 2>&1
     $startExit = $LASTEXITCODE
     $startOutput | ForEach-Object { Write-Host $_ }
-    Restart-StaleCollector $ScriptPath $PriorVersion
+    $staleOk = Restart-StaleCollector $ScriptPath $PriorVersion
+    if (-not $staleOk) {
+        Write-Host "   $script:PILOT_LAST_RESTART_ERR" -ForegroundColor Yellow
+    }
 
     $deadline = (Get-Date).AddSeconds($TimeoutSeconds)
     do {
         $statusOutput = & powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -File $ScriptPath status 2>$null
         if ($statusOutput -match "is running") {
             $ErrorActionPreference = $prevEAP
-            return $true
+            # Deliberately $staleOk, not $true: a collector that outlived the deploy answers
+            # "is running" from the old version dir just as convincingly as the new one.
+            return $staleOk
         }
         Start-Sleep -Seconds 2
     } while ((Get-Date) -lt $deadline)
--- a/deploy/installer-opensource.ps1
+++ b/deploy/installer-opensource.ps1
@@ -183,12 +183,21 @@ function Resolve-Node {
     $candidates = @()
 
     # Existing installations pin the exact Node binary used for deployment.
+    #
+    # The pin file holds one absolute path to node.exe, and for a managed runtime that
+    # path sits under the data dir -- i.e. under %USERPROFILE%, which can be non-ASCII.
+    # 5.1 defaults both Get-Content and Set-Content to the ANSI codepage, so an
+    # unqualified write stored "C:\Users\??.HOST\..." and every reader then failed
+    # Test-NodeSuitable and silently fell back to whatever node.exe the candidate search
+    # found first -- on a shared machine that was another account's nvm install.
+    # -Encoding UTF8 always emits a BOM on 5.1 (there is no utf8NoBOM), and U+FEFF is not
+    # whitespace, so .Trim() alone leaves it in the path: strip it explicitly first.
     foreach ($pinFile in @(
         (Join-Path $DataDir "node-bin"),
         (Join-Path $CACHE_DIR "node-bin")
     )) {
         if (-not (Test-Path -LiteralPath $pinFile)) { continue }
-        $pinned = ([string](Get-Content -LiteralPath $pinFile -Raw -ErrorAction SilentlyContinue)).Trim()
+        $pinned = ([string](Get-Content -LiteralPath $pinFile -Raw -Encoding UTF8 -ErrorAction SilentlyContinue)).Trim([char]0xFEFF).Trim()
         if ($pinned) { $candidates += $pinned }
     }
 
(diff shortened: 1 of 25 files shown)


src/utils/win-archive.ts
```diff
@@ -0,0 +1,187 @@
...
+    await fs.rename(staged, dest);
+    renamed = true;
+  } catch (err) {
+    const code = (err as NodeJS.ErrnoException).code;
+    if (code !== 'EXDEV') throw err;
+    await fs.cp(staged, dest, { recursive: true });
+    await fs.rm(staged, { recursive: true, force: true }).catch(() => {});
+  }
+  // Only the rename needs this — fs.cp creates new files, which inherit normally.
+  if (renamed) await resetInheritedAcl(dest);
```
linrunqi08: **[High]** `replaceDirWith` ignores `resetInheritedAcl` failure after rename; the installer treats the same failure as fatal and falls back to npm.

**影响:** Auto-update under a CJK profile that staged in `%SystemRoot%\Temp` can publish a `node_modules` tree the non-elevated scheduled task cannot read (`ERR_MODULE_NOT_FOUND: Cannot find package 'pino'`), while the updater logs `prebuilt node_modules installed`. The sqlite3 probe runs as the updater process and can still read Administrators-owned files.

**建议:** If `resetInheritedAcl` returns false, delete `dest` and throw/return failure so `ensureNodeModules` takes the existing npm fallback, matching `Ensure-NodeModules`.

deploy/installer-opensource.ps1
```diff
@@ -1221,6 +1462,46 @@ function Print-Summary {
...
+# $serviceEntry is whichever entry point the caller already resolved: the .ps1 needs
+# powershell.exe -File, the .cmd wrapper is invoked directly. Branching here rather than at
+# the call sites keeps this block byte-identical across the three installers.
+function Restart-StaleCollector {
+    param([string]$serviceEntry, [string]$priorVersion)
+    if (-not $priorVersion) { return }
+    if (-not $serviceEntry) { return }
+    if (-not (Test-Path $serviceEntry)) { return }
+    if ($serviceEntry -match '\.ps1$') {
+        & powershell.exe -NoProfile -ExecutionPolicy Bypass -File $serviceEntry restart-collector 2>$null | Out-Null
```
linrunqi08: **[High]** `Restart-StaleCollector` discards stdout/stderr/`$LASTEXITCODE`. `Start-PilotAndWait` then treats any `status` match of `is running` as success. `Cmd-Status` prefixes the **new** `current` version onto whatever PID is alive.

**影响:** This is the silent-success the helper was added to close. `start` is a no-op when a survivor is up; if `restart-collector` fails (S4U / access denied), install prints "Service started" and upgrade may `GC-OldVersions` while the old dist keeps running.

**建议:** Fail the install on a non-zero `restart-collector` exit. After restart, require a heartbeat newer than the restart **and** a runtime version dir equal to the just-written `current`. Do not infer success from `is running`.

deploy/installer-opensource.ps1
```diff
@@ -863,14 +1070,48 @@ function Deploy-Package {
...
     }
-    Msg "    ✅ Hook 脚本已部署" "    ✅ Hook scripts deployed"
+    # Never print the check mark for a postinstall that did not finish. postinstall.js
+    # reports its own per-tree failures and still exits 0 on purpose (a non-zero exit
+    # would abort the npm install fallback above), so a non-zero code here means the
+    # process died outright. That is exactly how the hooks/skills/plugins trees went
+    # missing on every Windows install without a word: fs.cpSync fail-fasts with
+    # 0xC0000409 on the bundled Node, killing the script before it copied anything,
+    # and this line printed "Hook scripts deployed" regardless.
+    if ($postinstallExit -ne 0) {
```
linrunqi08, line 1126: **[High]** `Deploy-Package` does not abort when postinstall exits non-zero — it only changes the message. `postinstall.js` still exits 0 on per-tree failures and on a thrown `main()`. Comments/tests claim installers key the check mark off the `failed asset tree` line; only the updater greps it.

**影响:** A `fs.cpSync`-style crash now prints ❌ but the install continues. A partial copy still prints "Hook scripts deployed". dsh then fails every cycle with "plugin file not found or unreadable".

**建议:** For the direct `node postinstall.js` invocation (not npm lifecycle), set `LOONGSUITE_PILOT_POSTINSTALL_STRICT=1` and fail `Deploy-Package` on non-zero / `failed asset tree`. Keep exit 0 only for `npm install`. Fix the test comment and assert the installer block actually matches that string.

scripts/loongsuite-pilot.ps1
```diff
@@ -58,16 +58,59 @@ $SPAN_ATTR_FILE = Join-Path $DATA_DIR "span-attributes.json"
...
+# supported only on core types") and unlike [Environment]::UserName. USERDOMAIN is not
+# always an account domain: under some logon providers (OpenSSH sshd among them) it is
+# the literal "WORKGROUP", which maps to no SID either, so fall back to the machine
+# name -- which is also what whoami prints for a local account, keeping the tag below
+# byte-identical for ASCII users who upgrade in place.
+function Get-PilotAccountName {
+    $user = [string]$env:USERNAME
+    if (-not $user) { return "" }
+    $domain = [string]$env:USERDOMAIN
+    if ((-not $domain) -or ($domain -eq "WORKGROUP")) { $domain = [string]$env:COMPUTERNAME }
```
linrunqi08, line 83: **[Medium]** `Get-PilotAccountName` only remaps `USERDOMAIN=WORKGROUP` to `COMPUTERNAME`. `whoami`'s domain can differ (Azure AD typically `AzureAD\user` vs machine `USERDOMAIN`).

**影响:** An ASCII upgrade registers a **new** task name. `restart-collector` / uninstall do not touch the old whoami-derived task, so the previous collector keeps serving the old dist — the stale-collector bug this PR is closing, reopened for those accounts.

**建议:** Keep env vars for the user part (the non-ASCII `whoami` stdout bug). For the domain, prefer `whoami` when its stdout is ASCII and contains no `?`. On upgrade, also delete the previous whoami-derived task names.

deploy/installer-opensource.ps1
```diff
@@ -1776,6 +2058,7 @@ function Start-PilotAndWait {
     $startOutput = & powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -File $ScriptPath start 2>&1
     $startExit = $LASTEXITCODE
     $startOutput | ForEach-Object { Write-Host $_ }
+    Restart-StaleCollector $ScriptPath $PriorVersion
```
linrunqi08: **[Medium]** When `$PriorVersion` is set, `Start-PilotAndWait` always `start`s then unconditionally `restart-collector`. The ratchet locks this order in.

**影响:** If `Stop-PilotService` actually held and the watchdog did not relaunch mid-deploy, `start` already spawned a healthy collector on the new `current`. The extra bounce can fail (S4U replace denied) and roll back a deploy that had already started correctly.

**建议:** Call `restart-collector` only when `start` reported "already running".

deploy/installer-opensource.ps1
```diff
@@ -252,6 +261,179 @@ function Get-ManagedNodePlatform {
...
+        $tarArgs = @('-xzf', $ArchivePath, '-C', $DestDir)
+        if ($i -ge $nativeCount) { $tarArgs = @('--force-local') + $tarArgs }
+        # Naming the binary is the whole diagnostic: "Cannot connect to C: resolve
+        # failed" only makes sense once you can see it came from Git's tar. Msg wraps
+        # Write-Host, so this never lands in the return value.
+        Msg "    tar: $exe" "    tar: $exe"
+        # Native commands do not throw on a nonzero exit, so $LASTEXITCODE is the
+        # only signal; the try only covers a binary that fails to launch at all,
+        # which $ErrorActionPreference = "Stop" would otherwise make terminating.
+        try {
```
linrunqi08, line 424: **[Medium]** New `Expand-PilotTarGz` / `Reset-PilotInheritedAcl` invoke native children with no timeout. The comments already record that older GNU tar hangs on `C:` rsh syntax. The node helpers pass timeouts.

**影响:** If System32 tar is missing and PATH is Git GNU tar that ignores `--force-local`, the installer hangs forever and `finally` never cleans `$script:TMP_DIR`. `icacls /T` over a full node_modules plus Defender has no watchdog either.

**建议:** Run tar/icacls with a deadline (`Start-Process -Wait` or a job) and treat timeout as failure → npm fallback / abort extract.

tests/unit/scripts/ps1-json-encoding.test.mjs
```diff
@@ -81,6 +81,38 @@ const interchangeSites = (text) => {
...
+  return lines.filter(([, line]) => (
+    /\|\s*&\s*\$(?:script:)?NODE_BIN\b/.test(line)
+    && vars.some(v => new RegExp(`\\$(?:script:)?${v}\\s*\\|`).test(line))
+  ));
+};
+
+// Both internal installers; the open-source variant already stages its config payload but
+// deliberately does NOT bump $OutputEncoding, because its PROBE_RESULT pipes do not strip
+// a BOM and would break if one appeared.
+const STAGED_VARIANTS = ['deploy/installer.ps1', 'deploy/installer-inner.ps1'];
```
linrunqi08: **[Medium]** New `STAGED_VARIANTS` lists `deploy/installer.ps1` and `deploy/installer-inner.ps1`, neither of which exists in this OSS tree. `it.skipIf(!existsSync)` means these tests never run on GitHub CI. The only installer here is excluded on purpose.

**影响:** The UTF-8 `$cfgTmp` staging in `installer-opensource.ps1` has no ratchet. `PROBE_RESULT` still pipes into node under default ASCII `$OutputEncoding`.

**建议:** Add `installer-opensource.ps1` to `STAGED_VARIANTS` with an OSS-accurate floor (1 site). Stage `PROBE_RESULT` the same way as `$cfgJson`.

linrunqi08:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 3 |
| Medium   | 4 |
| Low      | 3 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | PASS |
| 死锁/卡死风险 | PASS |
| 状态恢复正确性 | FAIL |
| PowerShell CLM 合规 | PASS |

状态恢复 FAIL：`Restart-StaleCollector` 失败被吞后，`current` 已指向新目录而进程仍是旧 dist；`status` 用新版本号 + 旧 PID 报成功。`replaceDirWith` 在 ACL reset 失败时仍激活不可读的 `node_modules`。

### Merge Gate（合入门禁）

**BLOCK** ⛔（阻断级别：Critical / High / Medium；Lifecycle 任一 FAIL 也阻断）

| Severity | File:line | 现状 |
| High | `src/utils/win-archive.ts:186` | ACL reset 失败仍当成功 |
| High | `deploy/installer-opensource.ps1:1498` | restart 失败被吞，status 版本盲 |
| High | `deploy/installer-opensource.ps1:1108` | postinstall 失败不 abort / 仍声称成功 |
| Medium | `scripts/loongsuite-pilot.ps1:83` | 任务身份可能与 whoami 分叉 |
| Medium | `deploy/installer-opensource.ps1:2061` | 无条件二次 restart |
| Medium | `deploy/installer-opensource.ps1:424` | tar/icacls 无 timeout |
| Medium | `tests/unit/scripts/ps1-json-encoding.test.mjs:114` | OSS staging ratchet 空转 |

阻断合入，待修复后重评。

### 总体结论

编码边界、`fs.cpSync` 崩溃、版本目录不就地覆盖这几条主路径方向正确，也有对应 ratchet。合入前必须补上与安装器不对齐的 updater ACL 失败处理、以及「stop claiming success」仍剩的两条假成功（stale collector / postinstall 部分失败）。新 CLM hunk 没有引入 `[pscustomobject]`/`[ordered]`。

### Highlights（正向实践）

- ASCII staging + Move-Item，PS1 对 ACL 失败回退 npm
- `copyTree` 避开 Node 22.9 `cpSync` `0xC0000409`；每棵树独立 try
- 身份走 env、pin UTF-8 + 显式剥 BOM
- `current` 在依赖之后写；`DeploymentManager` 记录 `{success:false}` 原因

---
*评审报告详见: `code-review/pr-309/final-report.md`*

src/utils/win-archive.ts
linrunqi08: Fixed in c9196bfe, as suggested. `replaceDirWith` now tracks whether it took the rename path, and on an unrepairable ACL it discards the tree and throws:

```ts
if (renamed && !(await resetInheritedAcl(dest))) {
  await fs.rm(dest, { recursive: true, force: true }).catch(() => {});
  throw new Error(`could not reset inherited ACL on ${dest}, discarded it`);
}
```

`renamed` guards it because only the rename carries the source ACEs — the `EXDEV` fallback uses `fs.cp`, which creates new files that inherit normally.

Throwing hands the caller its existing failure path: `ensureNodeModules` catches, logs `falling back to npm install`, returns false, and npm creates `node_modules` in place so it inherits normally. Same choice as `Ensure-NodeModules` in `deploy/installer-opensource.ps1`, which drops the prebuilt tree and returns `$false`.

Deliberately fatal rather than best-effort, for the reason you gave: the tree is in place and looks perfectly installed, and the process that has to read it may hold no usable ACE — which is what later surfaces as `ERR_MODULE_NOT_FOUND: Cannot find package 'pino'` on a directory that is plainly there.

Two ratchets added in `tests/unit/utils/win-archive.test.ts`. `resetInheritedAcl` is a no-op off Windows so this path is unreachable from a CI host; they pin the shape instead — the guarded block must contain both the `fs.rm` and the `throw`, no call site may ignore the return value (`await resetInheritedAcl(` at statement position is rejected), and `ensureNodeModules` must still turn the throw into the npm fallback.

deploy/installer-opensource.ps1
linrunqi08: Fixed the discard. Did not build the positive proof — taking the two halves separately.

**Discarded exit code (fixed in c9196bfe).** `Restart-StaleCollector` now returns `$true`/`$false` with the reason in `$script:PILOT_LAST_RESTART_ERR`. Native commands do not throw on a nonzero exit, so `$LASTEXITCODE` is the only signal there is, and `Cmd-RestartCollector` exits 1 in exactly the cases that matter — node runtime unresolvable, bootstrap entry missing, service manager refusing to restart a non-degraded install. In every one of those the survivor is still serving the OLD version dir.

`Start-PilotAndWait` now returns `$staleOk` instead of `$true` on the `is running` match, with the reason printed in yellow. The install path already `throw`s on `$false` and the upgrade path already rolls back and `exit 1`, so a failed displacement now fails the deploy without further plumbing.

Ratchets in `tests/unit/deploy/installer-restart-stale-collector.test.mjs`: every call site must be `$x = Restart-StaleCollector $a $b` (a discarded verdict fails), `$x` must then reach a decision that gates the success message (`-and $x` / `-not $x` / `return $x`), and `restart-collector` may not be followed by `| Out-Null` anywhere.

**Heartbeat + runtime version dir (not done).** Agreed that `is running` is not positive evidence — and your reason is the sharp one: `Cmd-Status` prefixes the *new* `current` onto whatever PID is alive, so the survivor answers it exactly as convincingly as the new collector would. But the check needs a runtime field the status path does not expose today (the version the live process actually resolved at startup), so it is a status-output change, not a local edit here. Left as a follow-up rather than half-built against a field that does not exist yet.

One consequence worth naming: now that a failed displacement fails the deploy, the unconditional second bounce you flagged separately costs more than it used to. Answered in that thread.

deploy/installer-opensource.ps1
linrunqi08: Two claims here; taking them separately.

**"Comments/tests claim installers key the check mark off `failed asset tree`; only the updater greps it."** Correct, and that was the worse half — the documentation was ahead of the code. Fixed in c9196bfe: the `pilot-postinstall-exit-check` block now captures the output (`2>&1`, because the count is on stderr), sets `$postinstallPartial` from `-match "failed asset tree"`, still writes everything postinstall printed, and puts the check mark behind `elseif ($postinstallPartial)`. New ratchet `every .ps1 installer also catches a partial failure` asserts the capture shape, the match, the pass-through, and that `Hook scripts deployed` sits after the partial branch — so the string is now pinned on both ends, in all installer variants, and the block stays byte-identical.

**"`Deploy-Package` does not abort."** True, and not changed. A missing asset tree is degraded-but-running — the collector starts, hooks/plugins are absent — and `LOONGSUITE_PILOT_POSTINSTALL_STRICT` plus a hard abort is a product decision about what a partial copy should do, which I would rather not land inside an encoding fix. An install that stops because a skills doc tree failed to copy is a different thing from an install that lies about it, and it is the lying part this PR set out to remove. Recorded as a follow-up.

**One gap your comment names that the fix does *not* close.** A thrown `main()` is caught at the bottom of `scripts/postinstall.js`, prints `Post-install failed: <msg>`, and exits 0. That string is not `failed asset tree`, so neither the exit code nor the new grep sees it — the installer would print ✅ over it. The exit code covers a hard crash (the `0xC0000409` case), the grep covers per-tree failures, and this third path is covered by neither. Cheapest correct fix is to widen the match to `Post-install failed|failed asset tree`; noting it here so it is on the record rather than implied fixed.

scripts/loongsuite-pilot.ps1
linrunqi08: Not changed, and I do not think it can be settled from the code — but the part of it that is decidable does not point where the suggestion does.

**What the divergence hinges on.** `Get-PilotAccountName` builds `DOMAIN\user` from `$env:USERDOMAIN` + `$env:USERNAME`, remapping only the literal `WORKGROUP`. Whether that diverges from `whoami` on an Azure AD-joined device depends entirely on what `%USERDOMAIN%` reports there. If the logon session populates it with `AzureAD` — which is what I would expect, since the `WORKGROUP` special case exists precisely because *some* providers populate it oddly rather than because it is normally wrong — there is no divergence and no orphan task. Neither of us can read that out of this repo, and I do not have an Azure AD-joined box to measure it on. Everything downstream of it (the orphan task, the reopened stale-collector bug) is conditional on that one fact, so I am not going to act on it unverified.

**Why not the suggested remedy either way.** "Prefer `whoami` when its stdout is ASCII and contains no `?`" would have to revert `tests/unit/scripts/ps1-nonascii-windows-paths.test.mjs:122` (`no tracked .ps1 derives an account identity from whoami`), which is one of the ratchets this PR adds. It also does not hold up on its own terms: the corruption it screens for is lossy and silent. Every character the console codepage cannot represent arrives as a literal U+003F, so a name that happens to transliterate into ASCII, or a domain that survives while the user part does not, passes the "is it ASCII, any `?`" test and still yields a string that maps to no SID. Screening a codepage-lossy channel by inspecting its output is the shape of bug this block was written to delete.

**What I would do if the divergence is confirmed.** The second half of your suggestion — deleting the previous whoami-derived task names on upgrade — is the right fix and is independent of the first half. It also generalises: enumerate the shared `\LoongsuitePilot` folder and remove tasks whose principal resolves to the same SID as the current user, which cleans up *any* prior tag, not just a whoami-derived one, and needs no guess about what the old tag was. Recorded as a follow-up, gated on a measurement from an Azure AD-joined device.

deploy/installer-opensource.ps1
linrunqi08: Valid, and this PR made it more expensive rather than less. Not fixed in this push; it is the top of the follow-up list.

**Why it got worse.** The companion fix in c9196bfe makes `Start-PilotAndWait` return `$staleOk` rather than `$true`, and the upgrade path rolls back and `exit 1` on `$false`. So the exact case you describe — `Stop-PilotService` held, no watchdog relaunch, `start` already spawned a healthy collector on the new `current`, and the unnecessary bounce then fails on a denied S4U replace — now rolls back a deploy that had started correctly. Before, it was silently ignored. Trading a silent-success for a spurious rollback is not obviously an improvement, and your gate removes both.

**It is also cheap.** `$startOutput` is already captured, and `start` prints `loongsuite-pilot is already running (PID <n>)` (`scripts/loongsuite-pilot.ps1:693`, `:698`) in exactly the survivor case, so the gate is one condition around the existing call — no new probing. The `after start` ratchet keeps passing, since the call stays where it is.

Two things kept it out of this push: the block is byte-identical across the installer variants so the edit lands in all of them at once, and "already running" then becomes load-bearing text in a script this PR is already changing, which wants a ratchet of its own rather than a bare `-match`. Both are small — happy to land it here if you would rather not split it out.

deploy/installer-opensource.ps1
linrunqi08: Accepted, not fixed. The unbounded wait is real; the fix is bigger than it looks on PowerShell 5.1.

`&` has no timeout, so a deadline means `Start-Process -PassThru` plus `WaitForExit(ms)` and `Kill()` — which also gives up inline stdout capture, so both call sites would have to redirect to temp files and read them back. `Reset-PilotInheritedAcl` sits inside a byte-identical shared block, so that restructuring lands in every installer variant at once, in the same push as the encoding fix it would be riding along with. Deferred on scope, not on merit.

What is already in place, for the record on how narrow the hang is: `Expand-PilotTarGz` tries `%SystemRoot%\System32\tar.exe` and the `Sysnative` alias by absolute path *first*, and only falls back to PATH's `tar` with `--force-local` — and it prints which binary it picked (`tar: <exe>`), so the Git-GNU-tar case is visible in the transcript rather than being a mystery hang. Reaching the hang needs System32 `tar.exe` absent (pre-1803 or a stripped image) *and* a PATH tar old enough to ignore `--force-local`. Narrow, but you are right that narrow is not bounded, and you are also right that `finally` never runs, so `$script:TMP_DIR` survives — recorded as a follow-up together with the `icacls /T` watchdog.

tests/unit/scripts/ps1-json-encoding.test.mjs
linrunqi08: First half fixed, second half declined as written — it cannot be done as a one-liner, and doing only the encoding side would break the pipes.

**`STAGED_VARIANTS` (fixed in c9196bfe).** You are right that the rule was unchecked here: the two entries skip and the only installer present was excluded, so nothing enforced it on this repo's CI. `deploy/installer-opensource.ps1` is now listed with an OSS-accurate floor of 1 (the internal variants run the shape twice, install and reconfigure). Verified it actually executes rather than skipping:

```
✓ deploy/installer-opensource.ps1 stages every config payload as a UTF-8 file
```

The `$OutputEncoding` assertion moved into a separate list, because it is genuinely variant-specific — see below — instead of being silently carried along by the same loop.

**`PROBE_RESULT` — the two changes are a pair, and only one of them is safe alone.** On 5.1 there is no BOM-less UTF-8 encoding object, so setting `$OutputEncoding = [System.Text.Encoding]::UTF8` prepends U+FEFF to what the pipe hands the child. The installers that do bump it also strip it in *every* consumer (`readFileSync(0,'utf-8').replace(/^\uFEFF/,'')`, five sites). `installer-opensource.ps1` does neither: it has the same five `JSON.parse(readFileSync(0,...))` pipes and none of them strip a BOM, so bumping the encoding on its own turns all five into `Unexpected token` — which is why it is excluded rather than overlooked. The correct change is both together: strip the BOM at all five consumers *and* bump `$OutputEncoding`, in one commit.

**How reachable it is today**, since it decides whether that belongs in this PR: the fields these pipes read are `id`, `displayName`, `detected` and `reason`, and `reason` is the *unresolved* detection pattern from the agent definition (`findDetectionReason` returns `p`, e.g. `~/.claude`, not the expanded absolute path — `src/cli-probe.ts:16-27`). For the shipped `agents.d` definitions all four are ASCII, so the ASCII `$OutputEncoding` does not currently corrupt anything; a user-authored `agents.d.local` definition with a non-ASCII `displayName` is the one input that reaches it. Latent, not live — so I am recording the paired change as a follow-up rather than adding five BOM strips plus an encoding switch to this PR. Say the word if you would rather have it here.

linrunqi08:
## 🔄 Round 2 Review — Post-fix Verification

Reviewed the two new commits (`c9196bfe`, `25511da0`) and all author replies.

### Resolved (3 High + 1 Medium)

| # | Sev | Issue | Verdict |
|---|-----|-------|---------|
| 1 | High | `replaceDirWith` ignores `resetInheritedAcl` failure | ✅ Fixed — discards tree + throws; two ratchets pin the shape |
| 2 | High | `Restart-StaleCollector` discards exit code | ✅ Fixed — returns `$true`/`$false`, callers gate success on it; ratchet ensures verdict is never discarded |
| 3 | High | `postinstall.js` exit 0 + installer prints ✅ | ✅ Fixed — captures output, greps `failed asset tree`, three-way branch; byte-identical across variants |
| 7 | Medium | `STAGED_VARIANTS` skips OSS installer | ✅ Fixed — `installer-opensource.ps1` listed with floor=1; `$OutputEncoding` split to its own list |

All three High findings are resolved. The fixes are well-structured with matching ratchet tests that prevent regression.

### Deferred with adequate justification (2 Medium)

| # | Sev | Issue | Verdict |
|---|-----|-------|---------|
| 4 | Medium | `Get-PilotAccountName` Azure AD divergence | ⏸️ Deferred — needs a real Azure AD box to measure; author's counter-argument on `whoami` codepage lossiness is sound; SID-based cleanup is the right fix once confirmed |
| 6 | Medium | No timeout on native `tar`/`icacls` commands | ⏸️ Deferred — requires `Start-Process` restructuring across all byte-identical installer variants; narrow trigger (System32 tar absent + old GNU tar); logged which binary was picked |

Both deferrals are reasonable: #4 is blocked on a measurement, #6 is a structural change bigger than this PR's scope.

### Outstanding — recommend fixing before merge (1 Medium)

| # | Sev | Issue | Verdict |
|---|-----|-------|---------|
| 5 | Medium | Unconditional second bounce after `start` | ⚠️ **Not fixed, and got worse** |

The fix in `c9196bfe` that makes `Start-PilotAndWait` return `$staleOk` (instead of `$true`) turned an old silent-success into a **spurious rollback**: when `Stop-PilotService` fully succeeded and `start` already launched the new collector, the unconditional `Restart-StaleCollector` can fail (e.g. denied S4U task replacement), causing the installer to roll back a deploy that was already running correctly.

The author agrees this got more expensive and notes the fix is small: gate on `$startOutput -match "is already running"` — the text is already printed by the service script (`scripts/loongsuite-pilot.ps1:693, :698`). Author offered to land it in this PR.

**Recommendation:** Land the gate before merge to avoid the new rollback regression path.

### Minor gap noted by author (not blocking)

`main()` exceptions in `postinstall.js` are caught at the bottom, print `Post-install failed: <msg>`, and exit 0. This string is neither a non-zero exit nor `failed asset tree`, so the installer's three-way check misses it. Author suggests widening the grep to `Post-install failed|failed asset tree`. Worth a follow-up but not a merge blocker.

### Merge Gate

**CONDITIONAL APPROVE** — all 3 Highs resolved; recommend landing the #5 bounce gate (small, author-agreed) before merging. If that lands, this is a clean approve.

deploy/installer-opensource.ps1
linrunqi08: Correction to the second half of this reply: the abort is in, as of 7cbfe5e5. The reasoning above (rollback cannot repair a shared, non-version-scoped data dir) still holds, but it argued for the wrong conclusion — an install that finishes with an empty `$DataDir\hooks` loads no hook at all and reports `running`, which is worse than stopping.

What changed:

- The check now covers **three** failure modes, not two. The third is the gap I flagged in this thread: a throw out of `main()` is caught at the bottom of `postinstall.js`, printed as `Post-install failed: <reason>`, and the process still exits 0 — no non-zero code, and nothing matching `failed asset tree`. The match is now `failed asset tree|Post-install failed`.
- The shared block still only *decides and records*, setting `$script:PILOT_POSTINSTALL_ERR` (empty means fine). Policy lives at the two call sites — the same split as `$script:PILOT_LAST_RESTART_ERR`, and the only reason the block can stay byte-identical across installers whose install and upgrade paths differ. A `throw` inside the block would have skipped `Cmd-Upgrade`'s rollback entirely (nothing in that path catches), leaving `current` on the new tree.
- Install prints the reason and `exit 1`. Upgrade does not start the new version, falls into the existing `if (-not $started)` rollback branch, and exits 1 after it.
- Same three conditions on the `.sh` side: `deploy_package()` captures the output and `return 1`s, install says why before exiting, upgrade rolls back.

Rollback still does not fix the data dir — it only keeps `current` on a version whose dependencies are known complete and stops the upgrade from reporting a silent success. The self-heal path is the next automatic update.

New ratchet cases in `tests/unit/scripts/postinstall-asset-trees.test.mjs`: `postinstall.js` must keep printing both lines the installers grep for; no `.ps1` installer may read the verdict and continue (every `Deploy-Package $script:INSTALL_SRC` call site is scanned); no `.sh` installer either — including the `set -e` footgun where `[ -n "$out" ] && printf` ends `deploy_package` on an empty, i.e. successful, run.

deploy/installer-opensource.ps1
linrunqi08: Landed in 779240d3, here rather than as a follow-up.

The gate is where you suggested — `restart-collector` runs only when `start` reported `already running`, which `Cmd-Start` prints off `Get-CollectorRuntime` or the pid file before returning without starting anything. Any other output means `start` did the launching, from the new `current`, so there is nothing stale to displace.

Two details worth naming, since they decide whether the gate holds:

**It lives in the shared block, not at the call sites.** `Restart-StaleCollector` takes a third `$startOutput` argument and gates on `($startOutput | Out-String) -notmatch "already running"`. The condition is uniform policy; only the *input* is per-caller. Putting it in the block keeps the block byte-identical across the installer variants, the same reason the `.ps1` / `.cmd` entry-point branch is already in there rather than at the callers.

**The string is now load-bearing in both directions.** An unset variable never matches `already running`, so a call site that forgets to capture would silently skip the displacement everywhere — the original bug back with no symptom at all. Rewording `Cmd-Start` would do the same. So `installer-restart-stale-collector.test.mjs` gained three assertions: the block gates on that exact condition; `scripts/loongsuite-pilot.ps1` still prints `loongsuite-pilot is already running (PID `; and for every call site the third argument must be a variable filled from the *preceding* `start` invocation and echoed to the user rather than swallowed by the capture. The existing `after start` ordering assertion is unchanged, since the call did not move.

Also worth recording for why this mattered more than the ordering: measured on a Medium-integrity `-RunLevel Limited` task, `schtasks /Delete`, `Register-ScheduledTask` and `Register-ScheduledTask -Force` all return `Access is denied`, while `Start-ScheduledTask` succeeds. A task's own principal gets `Read, Synchronize`; every write ACE is on `BUILTIN\Administrators`, which UAC filters out of that token. So the bounce is not merely redundant on a healthy start — it is *guaranteed* to fail wherever the task already exists, which is every re-install and every upgrade.

linrunqi08:
## ✅ Round 3 Review — All findings resolved

Reviewed commits `7cbfe5e5` and `779240d3`, which address the two remaining items from Round 2.

### Commit 4 (`7cbfe5e5`): postinstall failure now aborts the install

Fixes the "minor gap" noted in Round 2 and goes further:

- **grep widened** to `"failed asset tree|Post-install failed"` — covers all three failure modes (crash / per-tree / thrown `main()`)
- **Install aborts** on failure (`exit 1`); **upgrade skips start** and falls into rollback
- **`.sh` side synchronized**: `deploy_package()` captures output, `grep -qE` for both patterns, `return 1` on either; both call sites guard with `if ! deploy_package`
- **`set -e` footgun avoided**: uses `if [ -n "$postinstall_out" ]` instead of `[ -n ... ] && printf`
- **Ratchets**: `postinstall.js` must keep both grep anchors; both `.ps1` and `.sh` call sites must act on the verdict; byte-identical across installer variants

### Commit 5 (`779240d3`): bounce gated on "already running"

Fixes **#5 (unconditional second bounce → spurious rollback)**:

- `Restart-StaleCollector` takes a new `$startOutput` parameter; `-notmatch "already running"` → `return $true` (no bounce needed)
- `Start-PilotAndWait` passes the captured `$startOutput` through
- **Ratchets**: the `"already running"` gate must exist in the shared block; `scripts/loongsuite-pilot.ps1` must still print the line; every call site must pass three variable arguments with the third filled from `start` and echoed to the user

### Final status

| # | Sev | Issue | Status |
|---|-----|-------|--------|
| 1 | High | `replaceDirWith` ignores ACL failure | ✅ Fixed (c9196bfe) |
| 2 | High | `Restart-StaleCollector` discards exit code | ✅ Fixed (c9196bfe) |
| 3 | High | postinstall silent success | ✅ Fixed (c9196bfe + 7cbfe5e5) |
| 4 | Medium | `Get-PilotAccountName` Azure AD | ⏸️ Deferred — needs real device |
| 5 | Medium | Unconditional second bounce | ✅ Fixed (779240d3) |
| 6 | Medium | No timeout on native commands | ⏸️ Deferred — structural change |
| 7 | Medium | `STAGED_VARIANTS` missing OSS | ✅ Fixed (c9196bfe) |

All 3 High and 2 Medium findings are resolved. The 2 remaining Medium deferrals have adequate justification and are tracked as follow-ups.

**APPROVE** ✅

messixukejia (approved):

PR #313, opened 2026-08-24
fix(installer): keep uninstall going when node cannot be resolved

linrunqi08:
## The bug

A Windows box still had `%USERPROFILE%\.loongsuite-pilot\config.json` on disk — an SLS
AccessKeySecret and a license key inside it — after `-Uninstall -Purge` finished without
printing anything that looked wrong. Five defects on one path:

1. **`Resolve-Node` probes the nvm/fnm roots with a bare `Test-Path`.** `NVM_HOME` is
   routinely a *machine*-level variable pointing into another account's profile, whose DACL
   grants the current user nothing, so `Test-Path` emits a PermissionDenied record — and the
   `$ErrorActionPreference = "Stop"` in these scripts' headers promotes it to terminating.
   Both probes are non-fatal now. Only those two: the pin file lives under the caller's own
   data dir, where failing fast is still correct.

2. **`Cmd-Uninstall` called `Resolve-Node` bare, before any cleanup step**, so that one error
   skipped the plugin cleanups, the install-directory removal and the whole `if ($Purge)`
   branch — and the next `-Purge` died on the same line. The call is wrapped, `$NODE_BIN` is
   cleared, `-Purge` still runs, and the skipped steps are named at the end.
   `Remove-HookConfigs` / `Remove-OpenClawPlugin` now skip explicitly instead of printing a
   success line for a file they never touched.

3. **The one expected failure surfaced as a bare `ERROR: Access is denied.`** because the
   diagnostic behind it was unreachable: on PowerShell 5.1 a native command writing stderr
   while a `2>` redirection is in effect raises `NativeCommandError`, `EAP=Stop` makes that
   terminating, and the `if ($LASTEXITCODE -ne 0) { throw ... }` after it never runs. The
   `schtasks` call runs at `Continue` and restores the preference in `finally`, with the exit
   code pre-seeded non-zero so a `schtasks.exe` that cannot launch at all is not read as
   success through a stale `$LASTEXITCODE`.

4. **That message taught the wrong lesson.** It said only "Run uninstall from an elevated
   PowerShell", which reads as *uninstall needs admin*. It does not — the tasks are owned by
   `BUILTIN\Administrators` because the **install** ran elevated. The text now names the
   cause and the durable fix, and quotes the stderr line that says which of the two failures
   happened.

5. **An elevated install warns before doing any work.** Rewriting the task security
   descriptor after registration, so the user can delete their own tasks again, is the real
   repair and is deliberately not attempted here.

## Testing

[...]

PR #314, opened 2026-08-24
fix(service): scope the daemon reap to this installation

linrunqi08:
Stacked on #313 — review that one first; this PR's base is its branch.

## Three defects, one shared blast radius: other accounts on the same machine

**1. `Stop-OrphanProcesses` was not scoped to this installation.** It killed every node
process whose command line mentioned `collector-daemon` or `updater-daemon`. Those file names
are identical for every installation on the machine, and on a shared Windows box each account
runs its own pair out of its own `%USERPROFILE%` — so any install, restart or stop reaped all
of them. `Get-Process` only enumerates other users' processes when the caller is elevated,
which is exactly where the damage landed: the victims were left with a pid file naming a
process that no longer existed, which is where the "stale single-instance lock" reports came
from.

The predicate now also requires `$BOOTSTRAP_DIR`, which every launcher writes into the
command line verbatim (`"<node>" "<$BOOTSTRAP_DIR\<name>-daemon.js>"`). `$DATA_DIR` would be
the wrong key — the two roots diverge whenever `LOONGSUITE_PILOT_CACHE_DIR` is set alone.

The reap existed in **three** places, not one: the helper plus hand-inlined copies in
`Cmd-RestartCollector` and `Cmd-RestartUpdater` that the `-Match` parameter had never been
threaded into. Fixing the helper alone would have fixed nothing for `restart-collector` —
the command the updater issues on every deploy. Both are calls now, and a new ratchet pins
that exactly one node-process scan remains.

**2. `Stop-PidFile` deleted the pid file unconditionally** after the kill, up to ten seconds
later. The collector task carries a five-minute repeating trigger, so a successor may already
have written its own pid there — and then `status` reported a live daemon as not running and
the next `start` raced a second instance against it. It re-reads the file and removes it only
while it still names the process it killed, the same rule `removeOwnPidFileSync` already
follows on the daemon side.

**3. `UpdaterWatchdog`'s `missing-process` and `command-mismatch` branches were the only ones
not behind `inGraceWindow`.** `start()` runs its first check immediately, and a fresh install
has the collector up before the updater task is registered at all — so `missing-process` fired
on essentially every install: a `SERVICE_NOT_RUNNING` alarm plus a `restart-updater` racing
the installer's own launch. Mid-deploy the pid file legitimately still names the outgoing
version's process, which is the mismatch half. Both alarms moved behind the check too — an
alarm raised on every install hides the real ones.

## Testing

`tests/unit/scripts/ps1-daemon-reap-scope.test.mjs` (new) and three new grace cases in
`tests/unit/core/updater-watchdog.test.ts`. `tsc --noEmit` clean.

Defect 1 was reproduced and the fix confirmed on a live Windows PowerShell 5.1 box, read-only
(the check only evaluated the predicate — no `Stop-Process` anywhere in it). Four node daemons
were running, owned by **two different accounts**. The old predicate matched all four; the new
one matched exactly the caller's two, run once per account. One of those profile paths is
non-ASCII, and the comparison handles it.

### DIFF
--- a/scripts/loongsuite-pilot.ps1
+++ b/scripts/loongsuite-pilot.ps1
@@ -346,19 +346,53 @@ function Stop-PidFile {
     }
     # Force kill if still running
     try { Stop-Process -Id $pidVal -Force -ErrorAction SilentlyContinue } catch {}
-    Remove-Item $pidFile -Force -ErrorAction SilentlyContinue
+
+    # Delete the file only while it still names the process we just killed. Up to ten
+    # seconds elapse in the wait loop above, and the collector task carries a five-minute
+    # repeating trigger, so a successor may already have started and written its own pid
+    # here -- unconditional removal then deleted a live daemon's pid file, after which
+    # status reported it as not running and the next start raced a second instance against
+    # it. Same rule the daemons themselves follow on shutdown (removeOwnPidFileSync in
+    # src/utils/pid-utils.ts). Re-read rather than trusting $pidVal: the point is what is
+    # on disk now, not what was there before Stop-Process.
+    $currentPid = ""
+    if (Test-Path -LiteralPath $pidFile) {
+        $currentPid = ([string](Get-Content -LiteralPath $pidFile -ErrorAction SilentlyContinue)).Trim()
+    }
+    if ($currentPid -eq $pidVal) {
+        Remove-Item $pidFile -Force -ErrorAction SilentlyContinue
+    }
 }
 
 function Stop-OrphanProcesses {
     # $Match limits which daemons are terminated; the default kills both. Callers that
     # re-register a single task (Install-CollectorTask / Install-UpdaterTask) pass a
     # narrow pattern so they only reap the daemon they are about to re-launch.
+    #
+    # Both conditions below are required, and the second one is the point. The daemon
+    # names are shared by every installation on the machine: on a multi-account box each
+    # user runs their own collector and updater out of their own %USERPROFILE%, and
+    # matching on the name alone made any install / restart / stop kill all of them.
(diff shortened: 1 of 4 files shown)


linrunqi08:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 0 |
| Low      | 4 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | PASS |
| 死锁/卡死风险 | PASS |
| 状态恢复正确性 | PASS |
| PowerShell CLM 合规 | PASS |

### Merge Gate（合入门禁）

**APPROVE-READY** ✅（阻断级别：Critical / High / Medium，Low 不阻断；Lifecycle 均为 PASS）

无 open 阻断项。

### 总体结论

三处修复（本安装 reap 作用域、`Stop-PidFile` wait 路径条件删 pid、watchdog `missing-process`/`command-mismatch` 进入 grace）与调用链一致，CLM 合规。未发现阻断问题。四条 Low（`Test-PidRunning` 毫秒级 TOCTOU、Contains 嵌套 cache 子串、reap ratchet 字面匹配、command-mismatch 测试断言不唯一）可后续改进。

### Highlights（正向实践）

- 三处 node 扫描收成 `Stop-OrphanProcesses`，`-Match` 接到 updater 每次 deploy 的 `restart-collector`
- 作用域用 `$BOOTSTRAP_DIR` + `[string].Contains()`（CLM-safe，避开 regex 元字符）
- wait 路径对齐 `removeOwnPidFileSync`；grace 内不 `record` 告警（测试钉住）

---
*评审报告详见: `code-review/pr-314/final-report.md`*

PR #333, opened 2026-08-28
fix(qoder): drop slash-command control rows from turn splitting, and stop the legacy fallback re-emitting the turn entry

### LINKED ISSUE #335: fix(qoder): slash-command control rows open LLM-less turns and pollute prompt input
## Problem

Qoder CLI writes slash-command plumbing as `type=user` rows. A single `/model` invocation produces three of them:

| row | `isMeta` | `promptId` |
|---|---|---|
| `<local-command-caveat>Caveat: The messages below were generated by the user while running local commands…</local-command-caveat>` | `true` | absent |
| `<command-message>model</command-message>` + `<command-name>/model</command-name>` | **absent** | absent |
| `<local-command-stdout>Model set to … (persisted)</local-command-stdout>` | `true` | absent |

`splitContentEventsIntoTurns` skips `isMeta` rows, but the command echo carries no `isMeta`, so it passes `isRealUserPrompt` and opens a turn of its own. That turn holds no assistant row, so `buildLlmBoundaries` returns no boundaries and the result is a trace carrying the `enter_ai_application_system → invoke_agent` skeleton with no LLM or tool spans underneath.

The `isMeta` rows are not harmless either. Because they attach to the *preceding* turn, `buildUserMessageParts` — which filters only tool results — appends their text to a real prompt:

```
turn: starter promptId=… text="您好"
  other:       messages_delta parts=2
     [0] "您好"
     [1] "<local-command-caveat>Caveat: The messages below were generated by the u…"
  llm.request: messages_delta parts=2   (same)
```

So beyond the empty trace, the caveat banner is reported as part of the user's actual input.

Replaying the exported functions over 196 local CLI transcripts: 2 turns started by a control row, and 12 `messages_delta` parts holding control text.

## Proposed fix

Drop control rows at the `splitContentEventsIntoTurns` entry — the single point both `contentEvents` collection paths funnel into, and the only consumer of `contentEvents` before turns are built. They should neither start a turn nor contribute content to one.

Detection must key off the content envelope rather than metadata, because none of the metadata fields work:

| candidate | why it fails |
|---|---|
| `isMeta` | misses the command echo entirely, and also marks multimodal `[Image: …]` rows that must stay |
| `promptId` absent | genuinely interrupted prompts have no `promptId` either, and the field is *absent* rather than `null` |
| `userType` | every user row is `external`, including tool results |

## Acceptance criteria

- The `/model` control trio produces no turn of its own.
- A segment holding only control rows produces no records at all.
- No `gen_ai.input.messages_delta` part contains control-envelope text.
- A real prompt that received no model activity still yields one turn — interrupted-but-real turns must not be filtered away.
- `isMeta` multimodal image markers stay on their prompt, since multimodal collection relies on them.
- `tool_result` rows keep their prompt association.

### LINKED ISSUE #336: fix(qoder): a turn without model activity emits its entry other event twice
## Problem

`buildEventsFromBoundaries` pushes the turn's entry `other` record for the user row, then falls back to `buildLegacyEvents` when `boundaries.length === 0`. `buildLegacyEvents` re-walks the same `contentEvents` and normalizes that user row into a second `other`, because the caller passes its own `records` array in as `existingRecords` and gives the callee no way to know the entry record was already emitted.

The fallback is reached more often than the code suggests. `boundaries` is built exclusively from assistant rows in `buildLlmBoundaries` — progress events only supply `startTs`/`endTs` for each assistant group — so any turn without an assistant row takes this path. The comment above `buildLegacyEvents` claims it is used when the transcript has "no progress events AND no assistant blocks", which is inaccurate and hides the real trigger.

The two records are not identical copies, so downstream deduplication cannot merge them:

| field | first record (`buildEventsFromBoundaries`) | second record (`buildQoderHookRecord`) |
|---|---|---|
| `event.id` | `crypto.randomUUID()` | `row.uuid` |
| `gen_ai.request.model` | resolved model, else `unknown` | absent (undefined for user rows) |
| `gen_ai.input.messages_delta` | `buildUserMessageParts`, may hold several parts | `buildQoderInputMessagesDelta`, always a single text part |
| `agent.qoder.content_type` | hardcoded `text` | `content.type` |
| `agent.qoder.variant` | absent | present |

No extra trace is produced: both records share `gen_ai.turn.id`, so `injectTraceId` gives them one `trace_id` and `convertTurn` still builds a single ENTRY/AGENT pair — all `other` events are removed from span pairing. The duplication surfaces in two other places:

1. Two `other` records in the event logstore, inflating any event-count metric or dashboard. Because `event.id` differs, id-based dedup does not collapse them.
2. `collectUserHookInputMessages` concatenates every `other` record carrying input messages, so the ENTRY span's `input.messages` holds the same user message twice.

Measured on local CLI transcripts after #333: 78 of 913 turns are legitimately interrupted prompts with no model activity, and every one of them takes this path.

## Proposed fix

Skip non-`tool_result` user rows in the `buildLegacyEvents` loop. The entry `other` is always emitted by the caller, and `splitContentEventsIntoTurns` guarantees a turn holds at most one real prompt, so the predicate matches exactly the row the caller already handled — including its fallback branch. No state needs to cross the function boundary.

Also correct the misleading comment above `buildLegacyEvents` to state the real trigger.

Note that `isMeta` rows riding along on a prompt currently produce their own standalone `other` records; the same predicate removes those as well. They remain in the turn's `contentEvents`, and multimodal attachment collection reads `allParsed`, so it is unaffected.

## Acceptance criteria

- A prompt-only turn emits exactly one record, with `event.name=other`.
- `tool_result` rows in the same turn are still normalized, so over-filtering is caught.
- The ENTRY span's `input.messages` holds the user message once.
- Turns containing assistant rows are unchanged, since they never reach the legacy path.

rangemer333-cell:
Two defects in Qoder CLI transcript processing. They compound — the first decides how many bogus turns exist, the second doubles the records inside each one — so they are fixed together.

# Defect 1 — control rows open turns of their own

## Problem

Qoder CLI writes slash-command plumbing as `type=user` rows. A single `/model` invocation produces three of them:

| row | `isMeta` | `promptId` |
|---|---|---|
| `<local-command-caveat>Caveat: The messages below were generated by the user while running local commands…</local-command-caveat>` | `true` | absent |
| `<command-message>model</command-message>\n<command-name>/model</command-name>` | **absent** | absent |
| `<local-command-stdout>Model set to … (persisted)</local-command-stdout>` | `true` | absent |

`splitContentEventsIntoTurns` already skips `isMeta` rows, but the command echo carries no `isMeta`, so it still passed `isRealUserPrompt` and opened a turn of its own. That turn holds no assistant row, `buildLlmBoundaries` returns no boundaries, and the result is a trace with the `enter_ai_application_system → invoke_agent` skeleton and no LLM or tool spans underneath.

The `isMeta` rows were not harmless either. Because they attach to the *preceding* turn, `buildUserMessageParts` — which filters only tool results — appended their text to a real prompt:

```
turn: starter promptId=… text="您好"
  other:       messages_delta parts=2
     [0] "您好"
     [1] "<local-command-caveat>Caveat: The messages below were generated by the u…"
  llm.request: messages_delta parts=2   (same)
```

So beyond the empty trace, the caveat banner was being reported as part of the user's actual input.

## Fix

Drop control rows at the `splitContentEventsIntoTurns` entry — the single point both `contentEvents` collection paths funnel into, and the only consumer of `contentEvents` before turns are built. They neither start a turn nor contribute content to one.

Because this filter *discards*, a false positive is unrecoverable: the row disappears and, if it was a real prompt, the turn never gets a starter, leaving a trace with an `llm.response` and no user input. So the row must match a shape qoder-cli is confirmed to emit, not merely look control-like. All six genuine plumbing rows in the local transcripts agree on structure — `entrypoint: 'cli'`, a **plain-string** content (never a content-block array), and one of exactly three tag sequences:

| tag sequence | `isMeta` required |
|---|---|
| `[local-command-caveat]` | yes |
| `[local-command-stdout]` / `[local-command-stderr]` | yes |
| `[command-message, command-name]` (the echo packs both into one row) | no — the echo has none |

Consequences worth calling out: a lone `[command-name]` envelope is a **user message**, since nothing stops someone typing `<command-name>model</command-name>` and the CLI never emits that tag unpaired. A content-block array is a real message whatever its first block holds, which also means the check no longer looks at only the first block. And a control-shaped row from a non-CLI variant is left alone.

No metadata field can carry the decision on its own:

| candidate | why it cannot be the discriminant |
|---|---|
| `isMeta` | absent on the command echo, and also marks multimodal `[Image: …]` rows that must stay |
| `promptId` absent | genuinely interrupted prompts have no `promptId` either — locally only 13 of 833 real prompts carry one (and the field is *absent*, not `null`). Used only as a one-way short-circuit: a row that **has** one was booked by the CLI as a submitted prompt, so it is never plumbing |
| `userType` | every user row is `external`, including tool results |

The envelope sequence is recovered by a single forward scan rather than a regex:

```js
function scanControlEnvelopes(text) {
  const tags = [];
  let i = 0;
  while (i < text.length) {
    while (i < text.length && isEnvelopeSpace(text[i])) i++;
    if (i >= text.length) break;
    if (text[i] !== '<') return null;
    const openEnd = text.indexOf('>', i + 1);
    if (openEnd === -1) return null;
    const tag = text.slice(i + 1, openEnd);
    if (!CONTROL_TAGS.has(tag)) return null;
    const closeAt = text.indexOf(`</${tag}>`, openEnd + 1);
    if (closeAt === -1) return null;
    tags.push(tag);
    i = closeAt + tag.length + 3;
  }
  return tags.length > 0 ? tags : null;
}
```

A regex that repeats a lazily-matched envelope is ambiguous about where each envelope ends, so near-miss input backtracks exponentially. That is not academic here: the Stop hook `exec`s the processor in the **foreground with no timeout**, so the cost lands directly on the agent's shutdown path.

| adjacent envelopes + ordinary suffix | bytes | repeated-lazy regex | forward scan |
|---|---|---|---|
| 20 | 608 | 50.9 ms | 0.088 ms |
| 24 | 728 | 250.1 ms | 0.033 ms |
| 26 | 788 | 980.8 ms | 0.034 ms |
| 28 | 848 | **3923.9 ms** | 0.045 ms |
| 40 | 1208 | would not return | 0.016 ms |
| 4000 | 120008 | would not return | 0.713 ms |

## Verification

Replayed the real exported functions over the local CLI transcripts:

| metric | before | after |
|---|---|---|
| turns | 915 | 913 |
| turns started by a control row | 2 | **0** |
| turns with no model activity | 80 | **78** (interrupted-but-real turns untouched) |
| `messages_delta` parts holding control text | 12 | **0** |

Checked in both directions after each tightening: all 6 genuine plumbing rows are still dropped, and every control-tag row that is kept is a prose prompt discussing those tags — which correctly opens a turn of its own.

Tests: 15 cases in `tests/unit/hooks/qoder-control-message-turns.test.mjs`. Those covering the defects fail without this change. The rest pin the boundaries and pass either way — a prompt that is exactly one `[command-name]` envelope, a prompt wrapping a complete envelope in its own words, a dangling unclosed tag, a mismatched closing tag, a block array whose first block is an envelope, a control-shaped row from a non-CLI variant, the `isMeta` image marker staying on its prompt, a real prompt with no model activity still yielding one turn, and `tool_result` rows keeping their prompt association. The control-like-prompt case is covered both with and without a `promptId`, since the `promptId`-less form is the one the shape check has to catch on its own. One case classifies 40 adjacent envelopes with a suffix and asserts it completes under 200 ms.

# Defect 2 — legacy fallback re-emits the turn entry

## Problem

When `buildLlmBoundaries` yields nothing, `buildEventsFromBoundaries` hands off to `buildLegacyEvents` — but passes the full `contentEvents` list. The turn's user prompt has already been emitted as an `other` record into `records`, so the legacy per-row loop normalizes that same row again. The turn ends up with two textually identical `other` events differing only by a random `event.id`:

```
{"names":["other","other"],"otherIds":["d6b84eae","c2b0d127"],"stepIds":["(none)","(none)"],"sameText":true}
```

This is not a rare path. Every one of the 78 interrupted-but-real turns above goes through it.

The comment above `buildLegacyEvents` is part of why this went unnoticed — it claimed the path is taken when the transcript has "no progress events AND no assistant blocks", implying something nearly unreachable. The actual condition is only `boundaries.length === 0`; progress events merely supply timing and never prevent the fallback. Corrected here.

## Fix

Filter at the call site rather than inside the loop, so the legacy path receives only the rows the caller did not emit itself:

```js
const residualEvents = contentEvents.filter(
  row => row.type === 'assistant' || isToolResult(row),
);
```

Tool results must survive the filter: Qoder encodes them as `type: 'user'`, so a bare `type === 'user'` skip would silently drop every tool result on an interrupted turn. The parameter is renamed to `residualEvents` to keep the narrowed contract visible at the callee, rather than leaving `buildLegacyEvents` implicitly dependent on the caller having already emitted the user row.

Filtering at the call site is exactly equivalent to skipping inside the loop here, not merely similar — `contentEvents` is populated behind a `rowType === 'user' || rowType === 'assistant'` guard, so no third row type exists that the two formulations would treat differently.

## Verification

Same replay harness, 198 transcripts / 924 turns:

| metric | before | after |
|---|---|---|
| turns taking the legacy path | 78 | 78 (unchanged — no turn is lost) |
| turns emitting a duplicate `other` | **78** | **0** |
| redundant records | 78 | 0 |
| tool results routed through the legacy path | 0 | 0 |

Tests: 4 new cases in `tests/unit/hooks/qoder-legacy-duplicate-other.test.mjs`. Three fail without this change; the fourth asserts a normal turn with an assistant reply is untouched and passes either way. The prompt-only case also asserts `boundaries` is empty, so the test cannot quietly stop exercising the fallback if boundary detection changes later.

# Combined checks

`tsc --noEmit` clean. Serial runs: 88 tests across the six `qoder*` hook files, 127 tests across `qoder-trace-input`, `qoder-cli-input`, `qoder-cli-multimodal`, `qoder-input`, `qoder-cli-session` and the contract tests.

# Adjacent surface: `buildUserMessageParts`

`buildUserMessageParts` merges every non-tool-result user row of a turn into the prompt's `parts` for qoder-cli, keyed on the same weak `type === 'user' && !isToolResult` test that let control rows open bogus turns. Audited it as part of defect 1 and it needs no fix of its own:

- Its only production caller passes a post-split turn, so recognized control rows are already gone before the merge runs.
- The sibling `allParsed` path is `cliAttachmentFields`, which only ever reads `type === 'attachment'` rows — a control row cannot reach it.
- Nothing else in the processor turns a user row's text into output: `assignContentToBoundaries` skips non-tool-result user rows, `extractToolResults` requires a tool result, and the legacy path receives only `assistant`/tool-result rows.

Replaying 200 real transcripts (932 turns, 25635 records) through `splitContentEventsIntoTurns` → `buildLlmBoundaries` → `buildEventsFromBoundaries` finds control tags in exactly 4 emitted user parts, all of them prose prompts that quote the tags deliberately (the two review comments on this PR, on both their `other` and `llm.request` records).

The invariant was only pinned for a caveat row trailing a finished turn, so the last commit adds the mid-turn full-trio case asserting on both merge call sites. Neutering the split filter fails it.

# Out of scope

`buildQoderHookRecord` writes `agent.<namespace>.raw_type` with a per-variant namespace (`qodercn` for qoder-cn, `qoderwork` for qoder-work), but the second loop in `buildLegacyEvents` reads the hardcoded `agent.qoder.raw_type`. Pre-existing, unrelated to either defect, and worth its own PR.

Note for reviewers: `tests/unit/hooks/` has pre-existing flakiness under file parallelism, unrelated to this change. Each run fails a random 1–4 files spanning claude-code, codex, grok-build, qwen, and kiro, with hook subprocesses reporting `status = null`. It reproduces on an unmodified checkout; `--no-file-parallelism` is stable. Separately, `tests/unit/mask/string-masker.performance.test.ts` asserts a wall-clock budget of 50 ms and measured 63 ms on a CI runner during one run here; it passed on rerun.

Closes #335
Closes #336


fangxiu-wf (changes requested):
Reviewed at `25bab3f2e6dce13b28b501df1219cba65deafc37`.

两个原始问题的修复和新增回归测试都能通过，但新的控制行 matcher 还有两个需要处理的问题：

1. `CONTROL_ENVELOPE_RE` 存在指数级回溯。约 1 KB、28 个相邻 envelope 加普通后缀的输入，在 Node 25 上已经需要约 4.9 秒。Qoder Hook 同步执行 processor 且没有超时，这会阻塞 Agent 的 Stop 路径。建议改成确定性的单次扫描并增加超时回归测试。
2. matcher 会把没有 `promptId`、内容恰好为 `<command-name>model</command-name>` 的真实 prompt 删除；它也没有限制 CLI，并且只检查第一个文本块。复现中最终只剩 `llm.response`，没有用户输入或 `llm.request`。建议限定到已确认的 CLI 控制行结构，并验证完整 content blocks。

本地相关 15 个文件、211 个测试、typecheck 和 build 均通过；上述两项是新增测试尚未覆盖的确定性路径。

linrunqi08:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 0 |
| Low      | 1 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | PASS |
| 死锁/卡死风险 | PASS |
| 状态恢复正确性 | PASS |
| PowerShell CLM 合规 | N/A |

Stop hook 仍是无 timeout 的 `exec`，但本 PR 已用线性前向扫描替换会指数回溯的 matcher；过滤为同步纯函数，不改 `endLine` / cursor key。

### Merge Gate（合入门禁）

**APPROVE-READY ✅**（阻断级别：Critical / High / Medium；Low 不阻断；Lifecycle 全部 PASS）

无 open 阻断项。历史 inline review threads = 0。`fangxiu-wf` 在 `25bab3f2` 的两项 CHANGES_REQUESTED（ReDoS matcher、过宽分类）已在 `307864a2` 修复，本轮复核为 fixed。

### 总体结论

两条原始缺陷都落在 `splitContentEventsIntoTurns` / `buildEventsFromBoundaries` 的正确位置，历史评审指出的 ReDoS 与过宽匹配也已钉死。唯一 Low 是空 `userText` 的 prompt-only turn 不再发骨架 `other`，不阻断合入。

### Highlights（正向实践）

- 不可恢复 discard 只接受已确认的 CLI 三形状，lone `[command-name]` / block array / 非 CLI 均 keep。
- 正向扫描 + 40 envelope 超时回归，对准 Stop hook 无 timeout 的卡死面。
- legacy 调用点过滤并改名为 `residualEvents`，明确保住 `type=user` 的 tool_result。

### Low（不阻断）

- `qoder-hook-processor.mjs:1015`：空字符串 prompt-only turn 从 legacy 1 条骨架 `other` 变为 0 条且 cursor 仍前进。可后续补测或恢复 skeleton。

---
*评审报告详见: `code-review/pr-333/final-report.md`*

linrunqi08 (approved):
**APPROVE-READY.** No Critical/High/Medium findings. fangxiu-wf's ReDoS and over-broad matcher items at 25bab3f2 are fixed in 307864a2.

fangxiu-wf (approved):
Re-reviewed at `b1ba0cea7be439d6c51aa201ad473acb7df5ed1e`.

上次提出的两个问题均已修复：

- 控制行识别已改为确定性的单向扫描。原 28-envelope 复现从约 4.9 秒降至 0.027 ms，4000-envelope 输入约 0.648 ms。
- matcher 已限定到确认过的 Qoder CLI plain-string 行形状；单-envelope 合法 prompt、多内容块和非 CLI 输入均会保留。原复现现在完整生成 `other → llm.request → llm.response`。

本地在当前 main 与新 head 的合并结果上验证了 15 个测试文件、216 个测试，并通过 typecheck、build 和 diff check；当前 GitHub CI 全绿。没有发现新的阻塞问题，建议 Approve。

PR #332, opened 2026-08-27
fix(sls): report Pilot version consistently

rangemer333-cell:
## Summary

- stop Qoder Work, Qoder Work CN, and QwenWorkCN hooks from emitting runtime versions as the bare `version` field
- preserve runtime versions under the agent-scoped `agent.qoderwork.version` and `agent.qwenworkcn.version` fields
- strip the legacy bare runtime version while consuming hook history
- set the SLS `version` field centrally from the active Pilot package version for every agent
- keep JSONL, HTTP, and OTLP output behavior unchanged apart from removing the ambiguous bare alias

## Why

Commercial SLS serialization drops agent-scoped fields. Qoder Work-family records therefore retained only the bare runtime `version`, making it appear to be the Pilot version. Centralizing the bare field in `SlsFlusher` gives it one stable meaning across agents, while the namespaced fields continue to expose the agent runtime version where supported.

## Verification

- focused unit/integration suite: 77 tests passed for the original change
- agent-scoped version follow-up: 68 focused tests passed
- `npm run typecheck`
- `npm run build`
- `git diff --check`
- original full-suite run: 3543 passed, 56 skipped; 3 unrelated concurrent hook tests failed on the first run and all 41 assertions passed when those files were rerun serially

PR #344, opened 2026-08-31
fix(otlp): stop cross-source preemption from fragmenting Qoder CLI turns

rangemer333-cell:
## Problem

On the reporter's machine, 14 Qoder CLI traces contained nothing but an
`enter_ai_application_system` + `invoke_agent` pair — no turn/step/llm children,
`start == end`. The prompts inside them were ordinary user input (`# GitHub
Trending Deep Research …`), and the same turn's real content was scattered across
21–125 sibling traces.

This is **not** the defect PR #333 fixes. #333 removes slash-command control rows
at the processor layer (transcript → events) and only fires on abnormal input.
This one lives at the flusher layer (events → trace grouping), fires on perfectly
normal input, and loses nothing — the data is complete but cut apart. The two
merely share an exit: a truncated buffer holding only the turn-entry `other`.

## Root cause

Signal B closes a turn buffer when another turn of the same `agentType` +
`gen_ai.session.id` arrives. The assumption is that turns within a session are
**successive**. That holds inside one collection path — not across two.

Whenever `qoder-trace` is off, `orchestrator.ts` enables both Qoder CLI inputs
(L1248 `qoder-cli-hook`, L1265 `qoder-cli-session`, both gated on
`!qoderTraceEnabled()`):

| | turn id source | `agent.source` |
|---|---|---|
| `qoder-transcript-hook` | `crypto.randomUUID()` (processor L621) | `qoder-transcript-hook` |
| `qoder-cli-session-segment` | `record.turn_id` | `qoder-cli-session-segment` |

Same session id, **separate turn-id spaces** → the two buffers are concurrent,
not successive → each path's records preempt the other's mid-turn.

The truncated buffer often held just the turn-entry `other`. That record carries
`gen_ai.input.messages_delta`, so `isMetadataOnlyOtherEvent`
(`otlp-trace-flusher.ts:1582-1592`) deliberately lets it through — and its own
comment spells out the consequence: *"Converting such records standalone via the
ephemeral path still produces a phantom ENTRY+AGENT pair."* Two spans, zero
duration, a real prompt inside, no children. Exactly the observed symptom.

## Fix

Gate Signal B on the collection path:

```ts
if (buf.dataSource !== incomingDataSource) continue;
```

`dataSource` comes from `agent.source`, recorded at buffer creation and learned
lazily from later records, the same way `sessionId` already is.

**Backward compatible.** Agents that don't stamp `agent.source` (claude-code,
cursor, codex, …) compare `undefined` to `undefined` → equal → Signal B behaves
exactly as before. The guard only decides *whether to preempt*; it never changes
how records are grouped.

### Why the discriminator had a gap

`buildLegacyEvents` builds records through the shared `buildQoderHookRecord`
(`agent-event-normalizer.mjs`, also used by the Qoder IDE hook), which does not
stamp `agent.source`. Left alone, a legacy-path turn (`dataSource: undefined`)
and a boundary-path turn (`'qoder-transcript-hook'`) would read as different
sources and neither would close the other — a **same-path** regression that
matters because `turnIdleTimeoutMs` defaults to `0`, leaving only Signal A and
the 64-buffer cap as closers. Stamped at the qoder-only call site instead of in
the shared normalizer, so no other agent's records change.

## Relationship to #342

Worth flagging for whoever lands #342 (segment-source `gen_ai.turn.id` +
`gen_ai.step.id` mapping):

- **#342 alone does not fix this.** Giving the segment source a top-level
  `gen_ai.turn.id` moves the group keys from `turn:` vs `session:` to
  `turn:A` vs `turn:B`. The preemption condition (same agentType + same session
  + different key) is still satisfied.
- **The two are complementary.** #342's `finish_reasons` lets the segment buffer
  close itself via Signal A, which restores the closing path this PR removes.

## Verification

| Check | Result |
|---|---|
| `concurrent-subagent.test.ts` | 10 passed (2 new) |
| Teeth check: `if (false && buf.dataSource !== …)` | only the cross-source case fails; other 9 pass |
| `tests/unit/flushers` | 20 files / 181 passed |
| `tests/unit/hooks --no-file-parallelism` | 57 passed \| 2 skipped, 792 passed |
| `tests/contract` + `tests/unit/normalization` | 13 files / 152 passed |
| `npx tsc --noEmit` | clean |

The two new tests are a pair on purpose: one proves cross-source turns no longer
preempt, the other proves **same-source turns still do** — so the guard cannot
quietly switch Signal B off.

## Out of scope

Three other skeleton-trace causes were diagnosed on the same machine and are
**not** touched here:

- **LLM 0-duration spans** — `agents.d/qoder.json` declares `"events": ["Stop"]`,
  so `buildLlmBoundaries` never sees `PreToolUse`/`PostToolUse`. Separate change,
  larger surface.
- **Segment collapse (one step holding N zero-duration LLM spans)** — handled by #342.
- **`null` token spans** — deployment issue (daemon spawned the bare binary,
  bypassing the wrapper so `intercept.mjs` never loaded); already resolved by env injection.

A fourth reported symptom, "every span emitted twice by two listeners", is **not
a bug**: `getOrCreateExportState` (L1425-1426) filters endpoints by
`serviceName`, so the 908 = 454 × 2 count is ordinary `service.name` fan-out,
not duplicate delivery.

### DIFF
--- a/assets/hooks/qoder-hook-processor.mjs
+++ b/assets/hooks/qoder-hook-processor.mjs
@@ -1410,7 +1410,12 @@ function buildLegacyEvents(contentEvents, turnId, sessionId, agentId, runtimeCon
   // When no progress events are available, use the old per-line normalization
   for (const row of contentEvents) {
     const record = buildQoderHookRecord(row, { agentId, runtimeConfig, turnId });
-    if (record) existingRecords.push(record);
+    if (!record) continue;
+    // Same stamp the boundary path applies. The OTLP flusher uses agent.source to
+    // tell collection paths apart before letting one turn close another's buffer,
+    // so a turn falling back to the legacy path must not look like a foreign source.
+    record['agent.source'] = 'qoder-transcript-hook';
+    existingRecords.push(record);
   }
 
   // Apply step.id with timestamp proximity (legacy logic)
--- a/assets/hooks/qoderwork-hook-processor.mjs
+++ b/assets/hooks/qoderwork-hook-processor.mjs
@@ -119,11 +119,10 @@ function processTranscript(parsed, sessionId, agentId, runtimeConfig, cwd, opts 
   const firstRow = turns[0]?.[0] || contentRows[0];
   const userId = resolveUserId(firstRow, runtimeConfig);
   const providerName = inferProviderName({ 'gen_ai.agent.type': agentId });
-  const version = getStringValue(firstRow, 'version') || '';
 
   for (const turn of turns) {
     const turnId = getTurnIdForRows(turn);
-    const turnRecords = buildTurnEvents(turn, turnId, sessionId, userId, providerName, version, observedTs, runtimeConfig, cwd, agentId);
+    const turnRecords = buildTurnEvents(turn, turnId, sessionId, userId, providerName, observedTs, runtimeConfig, cwd, agentId);
     records.push(...turnRecords);
   }
 
@@ -182,7 +181,7 @@ function isPureSystemReminder(text) {
     && text.replace(/<system-reminder>[\s\S]*?<\/system-reminder>/g, '').trim().length === 0;
 }
 
-function buildTurnEvents(turnRows, turnId, sessionId, userId, providerName, version, observedTs, runtimeConfig, cwd, agentId) {
+function buildTurnEvents(turnRows, turnId, sessionId, userId, providerName, observedTs, runtimeConfig, cwd, agentId) {
   const records = [];
 
   // Find the user prompt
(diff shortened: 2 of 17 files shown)


rangemer333-cell:
Closing in favour of #345, which carries the identical change on a correct base.

This branch was cut from a local `main` that pointed at `115427b1` — the single commit of #332 (`codex/fix-sls-pilot-version`), which is still open and not part of `origin/main`. That made #332's commit show up in this PR's diff, inflating it to 17 files / +199 -23.

#345 contains only `c433ee48` on top of `origin/main`: 3 files, +115 -3. No content changed between the two.

PR #345, opened 2026-08-31
fix(otlp): stop cross-source preemption from fragmenting Qoder CLI turns

rangemer333-cell:
> [!IMPORTANT]
> **Depends on #342 — merge #342 first.** This PR stops the two Qoder CLI
> collection paths from cutting each other's turns short, but it also removes one
> of the ways a segment-source buffer used to get closed. #342 supplies the
> replacement (`gen_ai.response.finish_reasons`, letting Signal A close that
> buffer). Landing this one alone leaves segment buffers dependent on the
> 64-buffer cap on any turn whose last response is not terminal — on the
> reporter's machine that is 3 of 11 turns, because `stop_reason` is
> `tool_use` 69 times vs `end_turn` 8 times and `turnIdleTimeoutMs` defaults to 0.

## Problem

On the reporter's machine, 14 Qoder CLI traces contained nothing but an
`enter_ai_application_system` + `invoke_agent` pair — no turn/step/llm children,
`start == end`. The prompts inside them were ordinary user input (`# GitHub
Trending Deep Research …`), and the same turn's real content was scattered across
21–125 sibling traces.

This is **not** the defect PR #333 fixes. #333 removes slash-command control rows
at the processor layer (transcript → events) and only fires on abnormal input.
This one lives at the flusher layer (events → trace grouping), fires on perfectly
normal input, and loses nothing — the data is complete but cut apart. The two
merely share an exit: a truncated buffer holding only the turn-entry `other`.

## Root cause

Signal B closes a turn buffer when another turn of the same `agentType` +
`gen_ai.session.id` arrives. The assumption is that turns within a session are
**successive**. That holds inside one collection path — not across two.

Whenever `qoder-trace` is off, `orchestrator.ts` enables both Qoder CLI inputs
(L1248 `qoder-cli-hook`, L1265 `qoder-cli-session`, both gated on
`!qoderTraceEnabled()`):

| | turn id source | `agent.source` |
|---|---|---|
| `qoder-transcript-hook` | `crypto.randomUUID()` (processor L621) | `qoder-transcript-hook` |
| `qoder-cli-session-segment` | `record.turn_id` | `qoder-cli-session-segment` |

Same session id, **separate turn-id spaces** → the two buffers are concurrent,
not successive → each path's records preempt the other's mid-turn.

The truncated buffer often held just the turn-entry `other`. That record carries
`gen_ai.input.messages_delta`, so `isMetadataOnlyOtherEvent`
(`otlp-trace-flusher.ts:1582-1592`) deliberately lets it through — and its own
comment spells out the consequence: *"Converting such records standalone via the
ephemeral path still produces a phantom ENTRY+AGENT pair."* Two spans, zero
duration, a real prompt inside, no children. Exactly the observed symptom.

## Fix

Gate Signal B on the collection path:

```ts
if (buf.dataSource !== incomingDataSource) continue;
```

`dataSource` comes from `agent.source`, recorded at buffer creation and learned
lazily from later records, the same way `sessionId` already is.

**Backward compatible.** Agents that don't stamp `agent.source` (claude-code,
cursor, codex, …) compare `undefined` to `undefined` → equal → Signal B behaves
exactly as before. The guard only decides *whether to preempt*; it never changes
how records are grouped.

### Why the discriminator had a gap

`buildLegacyEvents` builds records through the shared `buildQoderHookRecord`
(`agent-event-normalizer.mjs`, also used by the Qoder IDE hook), which does not
stamp `agent.source`. Left alone, a legacy-path turn (`dataSource: undefined`)
and a boundary-path turn (`'qoder-transcript-hook'`) would read as different
sources and neither would close the other — a **same-path** regression that
matters because `turnIdleTimeoutMs` defaults to `0`, leaving only Signal A and
the 64-buffer cap as closers. Stamped at the qoder-only call site instead of in
the shared normalizer, so no other agent's records change.

## Dependency on #342

**This PR depends on #342. Merge order: #342, then this one.** The two changes
are halves of one fix — either alone leaves a gap:

- **#342 alone does not fix this.** Giving the segment source a top-level
  `gen_ai.turn.id` moves the group keys from `turn:` vs `session:` to
  `turn:A` vs `turn:B`. The preemption condition (same agentType + same session
  + different key) is still satisfied.
- **This PR alone is incomplete too.** The same-source guard is what keeps Signal
  B available *within* a path — and after #342, that is the only closer left for
  a segment turn ending on a non-terminal `stop_reason`. #342's `finish_reasons`
  restores Signal A as the primary closer for those buffers.
- **Together** they reduce fragmentation from "1 turn → 21–125 sibling traces"
  down to "1 turn → 2 traces" (one per collection path). The remaining 2-trace
  split is *not* fixed by either PR and is tracked separately — see the issue
  linked below.

Verified on a local merge of both branches: no conflicts, 996 tests passed /
8 skipped / 0 failed, `tsc --noEmit` clean.

## Verification

| Check | Result |
|---|---|
| `concurrent-subagent.test.ts` | 12 passed (4 new) |
| Teeth check: `if (false && buf.dataSource !== …)` | exactly the 3 cross-source cases fail; same-source preemption still passes |
| Stacked on #342 locally (`git merge`) | no conflict, same 12 passed |
| `tests/unit/flushers` | 20 files / 182 passed |
| `tests/unit/hooks --no-file-parallelism` | 57 passed \| 2 skipped, 792 passed |
| `tests/contract` + `tests/unit/normalization` | 13 files / 152 passed |
| `npx tsc --noEmit` | clean |

The tests are deliberately a set rather than a single case:

- **Both key shapes.** The segment input stamps neither `gen_ai.turn.id` nor
  `trace_id` today (`trace_id` comes only from `trace-linker.ts`, and
  `upstreamLink` is off by default), so its records group under `session:<sid>`;
  after #342 they group under `turn:<record.turn_id>`. Signal B walks every
  buffer of the same agent+session regardless of key shape, so both shapes are
  covered — the suite needs no fixture churn when the branches stack.
- **Both interleavings.** Signal B is evaluated on the *incoming* record, so
  whichever input arrives second does the preempting, and nothing orders the two
  inputs. The segment-first mirror case fails with `length 2 → 1` when the guard
  is off, which is the truncation symptom itself.
- **Both payloads.** Each test asserts the record count on *both* paths, so a
  guard that stopped preemption by dropping the other path's records would fail.
- **Same-source preemption still fires**, so the guard cannot quietly switch
  Signal B off.

The `session:`-keyed test also pins down what this PR does *not* fix: once the
hook turn closes, the segment buffer is still open and not `completed`. That is
asserted directly, so the missing real-time closer stays visible in CI instead of
being masked by the shutdown flush — and it is the concrete reason for the merge
order above.

## Out of scope

Three other skeleton-trace causes were diagnosed on the same machine and are
**not** touched here:

- **LLM 0-duration spans** — `agents.d/qoder.json` declares `"events": ["Stop"]`,
  so `buildLlmBoundaries` never sees `PreToolUse`/`PostToolUse`. Separate change,
  larger surface.
- **Segment collapse (one step holding N zero-duration LLM spans)** — handled by #342.
- **`null` token spans** — deployment issue (daemon spawned the bare binary,
  bypassing the wrapper so `intercept.mjs` never loaded); already resolved by env injection.

A fourth reported symptom, "every span emitted twice by two listeners", is **not
a bug**: `getOrCreateExportState` (L1425-1426) filters endpoints by
`serviceName`, so the 908 = 454 × 2 count is ordinary `service.name` fan-out,
not duplicate delivery.

### DIFF
--- a/src/flushers/otlp-trace-flusher.ts
+++ b/src/flushers/otlp-trace-flusher.ts
@@ -65,6 +65,11 @@ interface TurnBuffer {
   keyValue: string;
   agentType: string;
   sessionId?: string;
+  // Which collection path produced these records (agent.source). Two inputs may
+  // collect the same agent+session concurrently; Signal B must not treat one
+  // path's turn as a boundary for the other's. Undefined when the source does
+  // not stamp agent.source, in which case all such buffers count as one path.
+  dataSource?: string;
   records: AgentActivityEntry[];
   completed: boolean;
   lastActivityMs: number;
@@ -517,9 +522,18 @@ export class OtlpTraceFlusher extends BaseFlusher {
     // Different or unknown sessions may be concurrent; preempting one would
     // split its records and synthesize duplicate ENTRY/AGENT spans.
     const incomingSessionId = (entry['gen_ai.session.id'] as string | undefined) || undefined;
+    const incomingDataSource = (entry['agent.source'] as string | undefined) || undefined;
     for (const [bufKey, buf] of this.turnBuffers) {
       if (buf.agentType !== agentType || bufKey === key || buf.completed) continue;
       if (!incomingSessionId || !buf.sessionId || incomingSessionId !== buf.sessionId) continue;
+      // Signal B assumes turns within a session arrive one after another, which
+      // only holds inside a single collection path. When two inputs collect the
+      // same session at once (qoder-transcript-hook + qoder-cli-session-segment),
+      // each mints turn ids from its own id space, so their buffers are
+      // concurrent rather than successive. Preempting across paths cuts both
+      // mid-turn: the truncated buffer often holds just the turn-entry "other",
+      // which converts standalone into a phantom ENTRY+AGENT pair.
+      if (buf.dataSource !== incomingDataSource) continue;
       buf.completed = true;
       this.triggerFlush(buf, false);
     }
@@ -547,13 +561,19 @@ export class OtlpTraceFlusher extends BaseFlusher {
         keyValue: value,
         agentType,
         sessionId: incomingSessionId,
+        dataSource: incomingDataSource,
         records: [],
         completed: false,
         lastActivityMs: Date.now(),
       };
       this.turnBuffers.set(key, buf);
-    } else if (!buf.sessionId && incomingSessionId) {
-      buf.sessionId = incomingSessionId;
+    } else {
+      if (!buf.sessionId && incomingSessionId) {
+        buf.sessionId = incomingSessionId;
+      }
+      if (!buf.dataSource && incomingDataSource) {
+        buf.dataSource = incomingDataSource;
+      }
     }
     buf.records.push(entry);
     buf.lastActivityMs = Date.now();
--- a/tests/unit/flushers/otlp-trace-flusher/concurrent-subagent.test.ts
+++ b/tests/unit/flushers/otlp-trace-flusher/concurrent-subagent.test.ts
@@ -350,37 +350,58 @@ describe('OtlpTraceFlusher - concurrent subagent regression', () => {
     expect(mockConvert).not.toHaveBeenCalled();
   });
 
-  it('does NOT let a second collection path preempt the first one in the same session', async () => {
-    // Qoder CLI is collected by two inputs at once while qoder-trace is off:
-    // qoder-transcript-hook (turn id = crypto.randomUUID()) and
-    // qoder-cli-session-segment (turn id = record.turn_id). They share the
-    // session id but mint turn ids from separate id spaces, so their buffers
-    // are concurrent, not successive. Preempting across paths truncated the
-    // hook buffer down to its turn-entry `other`, which the converter turns
-    // into a phantom ENTRY+AGENT pair with no turn/step/llm children.
+  // Qoder CLI is collected by two inputs at once while qoder-trace is off:
+  // qoder-transcript-hook (turn id = crypto.randomUUID()) and
+  // qoder-cli-session-segment (turn id = record.turn_id). They share the
+  // session id but mint turn ids from separate id spaces, so their buffers
+  // are concurrent, not successive. Preempting across paths truncated the
+  // hook buffer down to its turn-entry `other`, which the converter turns
+  // into a phantom ENTRY+AGENT pair with no turn/step/llm children.
+  //
+  // The two paths group under different key shapes, and that shape changes with
+  // #342, so both are covered below: Signal B walks every buffer of the same
+  // agent+session regardless of key shape, and the guard must hold either way.
+  const QODER_SESSION = 'qoder-cli-session-1';
+  const QODER_HOOK = {
+    'gen_ai.agent.type': 'qoder-cli',
+    'gen_ai.session.id': QODER_SESSION,
+    'gen_ai.turn.id': 'c0ffee00-0000-4000-8000-000000000001',
+    'agent.source': 'qoder-transcript-hook',
+    // The hook processor writes no trace_id, so turn.id decides the key.
+    'trace_id': undefined,
+  };
+
+  function findCall(
+    mockConvert: { mock: { calls: unknown[][] } },
+    source: string,
+  ): Record<string, unknown>[] | undefined {
+    const call = mockConvert.mock.calls.find(
+      (args) => ((args[0] as Record<string, unknown>[])[0])['agent.source'] === source,
+    );
+    return call?.[0] as Record<string, unknown>[] | undefined;
+  }
+
+  it('does NOT let the session-keyed segment path preempt the hook path in the same session', async () => {
+    // Production shape on this branch: qoder-cli-session-segment stamps neither
+    // gen_ai.turn.id nor trace_id — no input writes trace_id for Qoder CLI and
+    // upstreamLink is off by default — so its records fall through
+    // resolveGroupKey to `session:<sid>` while the hook path uses `turn:<uuid>`.
+    // This is the shape CI must cover until #342 lands.
     const { convertEventLogToTrace } = await import('@loongsuite/otel-util-genai');
     const mockConvert = vi.mocked(convertEventLogToTrace);
     mockConvert.mockClear();
 
-    const session = 'qoder-cli-session-1';
-    const hook = {
-      'gen_ai.agent.type': 'qoder-cli',
-      'gen_ai.session.id': session,
-      'gen_ai.turn.id': 'c0ffee00-0000-4000-8000-000000000001',
-      'agent.source': 'qoder-transcript-hook',
-      'trace_id': 'ddd12f3577b34da6a3ce929d0e0e4736',
-    };
     const segment = {
       'gen_ai.agent.type': 'qoder-cli',
-      'gen_ai.session.id': session,
-      'gen_ai.turn.id': 'segment-turn-7',
+      'gen_ai.session.id': QODER_SESSION,
       'agent.source': 'qoder-cli-session-segment',
-      'trace_id': 'eee12f3577b34da6a3ce929d0e0e4736',
+      'trace_id': undefined,
     };
 
     // Hook turn opens with the entry `other` carrying the user prompt.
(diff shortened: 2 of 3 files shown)


fangxiu-wf (approved):
Reviewed `c433ee481ec011e34031daa0de0aac0f30eae116` on top of the latest main and PR #342. The source-aware Signal B guard prevents the Qoder transcript and segment inputs from preempting each other, while same-source successor preemption remains intact. The combined result passed 253 targeted tests, typecheck/build, and the full suite (3676 passed, 56 skipped). No blocking code findings. Please preserve the documented merge order: #342 first, then #345.

src/flushers/otlp-trace-flusher.ts
```diff
@@ -517,9 +522,18 @@ export class OtlpTraceFlusher extends BaseFlusher {
...
       if (buf.agentType !== agentType || bufKey === key || buf.completed) continue;
       if (!incomingSessionId || !buf.sessionId || incomingSessionId !== buf.sessionId) continue;
+      // Signal B assumes turns within a session arrive one after another, which
+      // only holds inside a single collection path. When two inputs collect the
+      // same session at once (qoder-transcript-hook + qoder-cli-session-segment),
+      // each mints turn ids from its own id space, so their buffers are
+      // concurrent rather than successive. Preempting across paths cuts both
+      // mid-turn: the truncated buffer often holds just the turn-entry "other",
+      // which converts standalone into a phantom ENTRY+AGENT pair.
+      if (buf.dataSource !== incomingDataSource) continue;
```
linrunqi08, line 536: **[High]** 按 `agent.source` 跳过跨采集路径 Signal B 之后，当前 main 上的 `qoder-cli-session-segment` 缓冲在进程生命周期内没有实时 closer。

**影响:** 生产 session 记录没有 `gen_ai.turn.id` / `gen_ai.response.finish_reasons`（`turn_id` / `stop_reason` 只在 attributes 里），`resolveGroupKey` 落到 `session:${id}`；`isTerminalEvent` 不触发；`turnIdleTimeoutMs` 默认 0。改前同 session 的 hook turn 是唯一 live closer，改后只剩 `flush()`/`shutdown()` 或 64-buffer cap。长驻 daemon 上该 path 的 OTLP 延迟到退出，`records[]` 按 session 无界堆积；cap 溢出时 `triggerFlush(..., false)` 还可能误驱其它 agent 的未完成 turn。作者声明依赖仍为 Open 的 #342；单独合入会把这个洞带到 main。

**建议:** **先合 #342 再合本 PR**（把 `record.turn_id` 升为 `gen_ai.turn.id`，并把 `stop_reason` 写成 `gen_ai.response.finish_reasons`）。若必须单独合入，本 PR 需给 session-keyed buffer 一条独立关闭路径（非 0 idle、按 `keySource==='session_id'` 周期 flush、或 usage-only 记录不进 `turnBuffers`）。不要把 `MAX_TURN_BUFFERS` 当内存上限。

tests/unit/flushers/otlp-trace-flusher/concurrent-subagent.test.ts
```diff
@@ -350,6 +350,93 @@ describe('OtlpTraceFlusher - concurrent subagent regression', () => {
...
+      'gen_ai.agent.type': 'qoder-cli',
+      'gen_ai.session.id': session,
+      'gen_ai.turn.id': 'c0ffee00-0000-4000-8000-000000000001',
+      'agent.source': 'qoder-transcript-hook',
+      'trace_id': 'ddd12f3577b34da6a3ce929d0e0e4736',
+    };
+    const segment = {
+      'gen_ai.agent.type': 'qoder-cli',
+      'gen_ai.session.id': session,
+      'gen_ai.turn.id': 'segment-turn-7',
```
linrunqi08: **[Medium]** 新测试给 segment 记录写了 `gen_ai.turn.id` 和 `finish_reasons: ['tool_call']`，与当前生产 `QoderCliSessionInput` 形状不一致；且只断言 hook 三条在一起，不断言 segment buffer 是否完整。

**影响:** CI 验证的是 #342 之后的 `turn:` 分组，不是本分支上的 `session:` 分组。session-segment 无实时 closer 的空洞会继续绿灯。反向交错（segment 先到）也未覆盖。

**建议:** 增加一条与生产一致的用例：`agent.source='qoder-cli-session-segment'`、有 `gen_ai.session.id`、**没有** `gen_ai.turn.id` / `finish_reasons`，断言 session buffer 有明确 closer（或在未 stacked #342 时失败）。再补 segment-first，并断言两条 path 的 convert payload 都完整。rebase 到 #342 之后再把 fixture 改回带 `turn.id` 的形态。

linrunqi08:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 1 |
| Medium   | 1 |
| Low      | 2 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | FAIL |
| 死锁/卡死风险 | PASS |
| 状态恢复正确性 | PASS |
| PowerShell CLM 合规 | N/A |

FAIL 证据：`qoder-cli-session-segment` 在当前 main 不发射 `gen_ai.turn.id` / `gen_ai.response.finish_reasons`，`turnIdleTimeoutMs` 默认 0。本 PR 的 `buf.dataSource !== incomingDataSource` 拿掉了原先唯一的 live closer（同 session 的 hook Signal B）。这些 buffer 要到 `shutdown()`/`flush()` 或 64-buffer cap 才会排空；`records[]` 按 session 无界堆积。

### Merge Gate（合入门禁）

**BLOCK ⛔**（阻断级别：`Critical` / `High` / `Medium`，`Low` 不阻断；Lifecycle 任一 FAIL 也阻断）

| Severity | 文件:行号 | 现状 |
| High | `src/flushers/otlp-trace-flusher.ts:536` | 单独合入后 session-segment 无实时 closer；依赖的 [#342](https://github.com/alibaba/loongsuite-pilot/pull/342) 仍为 Open |
| Medium | `tests/unit/flushers/otlp-trace-flusher/concurrent-subagent.test.ts:376` | 测试给 segment 写了生产并不存在的 `gen_ai.turn.id`，closer 空洞继续绿灯 |
| Lifecycle FAIL | 资源释放 | 运行时 session buffer 不释放 |

阻断合入，待修复后重评。**先合 #342 再合本 PR**，或在本 PR 给 session-keyed buffer 一条独立关闭路径，并补生产形状测试。

### 总体结论

按 `agent.source` 隔离 Signal B 能修好 hook 骨架 trace（ENTRY+AGENT、start==end），方向正确，同源仍 preempt 的成对测试也到位。但不能单独合入：当前 session 输入没有 turn.id / finish_reasons，这一刀切掉了它唯一的实时 flush。作者自己写了 Depends on #342，#342 尚未合入。

### Highlights（正向实践）

- discriminator 只影响 preempt、不改分组键，其它未 stamp `agent.source` 的 agent 行为不变
- 成对测试防止 guard 把 Signal B 整段关掉
- 与 #333（processor slash-command）分层说明清楚

---
*评审报告详见: `code-review/pr-345/final-report.md`*

tests/unit/flushers/otlp-trace-flusher/concurrent-subagent.test.ts
rangemer333-cell: 不是"先写 pre-#342、rebase 后改回 post-#342"，而是两种形状各一条测试。 #345 排在 #342 之后合并，两种形状都会真实出现过；覆盖两条则 merge 前后都成立，不留 rebase 时必然要改 fixture 的维护陷阱。已在临时分支真合了一次 #342 验证：无冲突，12 条照过。
没做"在未 stacked #342 时失败"。 那会让本分支 CI 变红。改为断言当前的真实行为：hook turn 关闭后，session:` buffer 仍然存在且completed === false—— 把缺失实时 closer 这个事实钉在 CI 里，而不是被收尾flush()掩盖。这条用了白盒读turnBuffers（仓库无先例，已注释说明理由：没有公开 API 能表达 buffer 的 key 和完成状态，且triggerFlush是同步摘 buffer、异步导出，所以读取不会和导出竞态）。

PR #358, opened 2026-09-02
fix(qoder): preserve per-invocation custom span attributes

fangxiu-wf:
## Summary

- add an explicit, default-off safe custom top-level field passthrough option to canonical hook normalization
- enable it only for Qoder Trace, Qoder CN Trace, and the Qoder CLI fallback input
- preserve caller-supplied string attributes after canonical entry construction so they cannot alter structural normalization
- keep reserved, sensitive, malformed, non-string, empty, comma-containing, and overlong values filtered
- leave OTLP prefix selection and all non-Qoder canonical adapters unchanged

## Why

The Qoder hook already stamps LOONGSUITE_PILOT_SPAN_ATTRIBUTES onto every hook record, but the canonical Qoder inputs projected only managed keys. As a result, arbitrary invocation attributes such as multica.issue.id were discarded before spanAttributePassthroughPrefixes could discover them.

## Verification

- targeted Qoder/custom-attribute tests: 118 passed
- full test suite: 3920 passed, 56 skipped
- npm run typecheck
- npm run build
- git diff --check

Fixes #357

PR #363, opened 2026-09-02
fix(qoder): support managed QoderCN SDK runtime

sipercai:
Builds on #358, now merged into main.

## Issue

Managed Qoder runtimes can provide the real CLI as a bundled JavaScript module rather than a self-executable binary. The QoderCN SDK currently bundles a qoder-worker-runtime.obf.mjs entry without a shebang. The Pilot wrapper directly execs the target, so this entry can be treated as a shell script and fail with a syntax error before the SDK starts.

QoderCN SDK print mode also has a transcript ordering difference: the terminal assistant row or last-prompt marker may be durable before the progress Stop row is appended. Requiring progress Stop alone can drop an otherwise complete turn. The configured QoderCN trace polling interval was also not passed to the input instance.

Managed Runtime Task attributes use the safe opt-in Qoder passthrough introduced by #358. This PR deliberately does not add agentcore fields to the canonical schema; it adds Qoder and QoderCN coverage proving that agentcore.task_id and agentcore.subtask_id follow the generic per-invocation attribute path. Runtime configuration selects agentcore. through spanAttributePassthroughPrefixes when those fields are required on spans.

## Changes

- Detect .js, .cjs, and .mjs CLI entries and launch them explicitly with Node while preserving the scoped token/system-prompt preload.
- Preserve native/Bun launcher behavior and the fail-open path when the intercept asset is unavailable.
- Accept QoderCN terminal assistant and last-prompt markers in addition to progress Stop, while keeping tool_use and pause_turn non-terminal.
- Pass the configured qoder-cn-trace polling interval to QoderCnTraceInput.
- Add regression coverage for bundled modules, terminal-marker timing, control-row filtering, Task attributes through #358, and polling configuration.

## Validation

- TypeScript typecheck passed.
- Focused suite: 7 files, 167 tests passed.
- Real managed QoderCN SDK flow passed with SDK 1.0.30 and bundled CLI 1.1.35 through the Pilot wrapper, including Task attributes, token usage, and system-prompt capture.
- #358 has its own full-suite, build, typecheck, and security validation.

## Scope

Credential selection and account management remain owned by the runtime/SDK and are unchanged. A separate SDK session-resume retry observed in the global Runtime integration is reproducible when bypassing this wrapper, so it is outside this Pilot fix.

### DIFF
--- a/assets/hooks/qoder-hook-processor.mjs
+++ b/assets/hooks/qoder-hook-processor.mjs
@@ -546,17 +546,16 @@ async function processTranscript(agentId, logPrefix, transcriptPath, sessionId, 
     // session_meta, other types: ignored
   }
 
-  // --- Phase 2.5 (qoder-cn only): Skip processing if transcript hasn't reached Stop ---
-  // For qoder-cn, only process the transcript when the Stop progress event has been
-  // written. A PostToolUse retry (which runs before Stop is written) would see an
-  // incomplete ReAct chain and produce partial events. The Stop retry (fired after
-  // the final assistant text is written) sees the complete chain and can correctly
-  // generate all llm.request events with proper input.messages_delta (including
-  // tool_result as input delta for step 2).
+  // --- Phase 2.5 (qoder-cn only): Skip processing until the transcript is terminal ---
+  // QoderCN does not guarantee that its progress Stop row is visible when the Stop
+  // hook process runs. In SDK --print mode the terminal assistant row and
+  // last-prompt marker can already be durable while progress Stop is appended only
+  // after the hook returns. Accept all terminal forms understood by the retry path;
+  // otherwise a complete managed-runtime turn is silently dropped forever.
   if (agentId === 'qoder-cn') {
-    const hasStop = progressEvents.some(pe => pe.hookEvent === 'Stop');
-    if (!hasStop) {
-      logDebug(agentId, `Transcript not yet complete (no Stop event in progress). Skipping processing.`);
+    const hasTerminalMarker = hasQoderCnTerminalMarker(progressEvents, parsed);
+    if (!hasTerminalMarker) {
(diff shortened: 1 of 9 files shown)


assets/hooks/qoder-hook-processor.mjs
```diff
@@ -755,6 +754,15 @@ function isTerminalStopReason(reason) {
   return typeof reason === 'string' && TERMINAL_STOP_REASONS.has(reason);
 }
 
+export function hasQoderCnTerminalMarker(progressEvents, rows) {
```
fangxiu-wf, line 757: ### [P1] 历史 terminal marker 会让未完成的当前轮被提前提交

这里会扫描整个 `rows`。在 missing-cursor/cold-start 场景下，如果历史 turn 已包含 `last-prompt` 或 `assistant.stop_reason=end_turn`，而最新 turn 仍未完整落盘，`hasQoderCnTerminalMarker()` 仍会返回 true，导致最新的未完成 turn 被提交并推进 checkpoint。

我使用本地真实 Qoder CN transcript 重放验证：截取到第 31 行时，最新 turn 只有 `thinking`、`stop_reason=null`，processor 已输出 3 条事件、标记为 `end_turn`，并将 checkpoint 推进到 31；补齐最终 assistant 和 `last-prompt` 到第 37 行后再次处理，又输出了 3 条事件，且使用不同的 `turn.id`。`QoderCnTraceInput` 只能在同一 `turn.id` 内去重，因此最终两批都会进入下游，形成 2 个 turn / 2 个 trace。

建议将 terminal 判断限定在“最新真实 user prompt 之后的当前 turn”；当前 turn 没有 terminal marker 时不要推进 checkpoint，应进入 retry。请增加回归测试：历史 turn 已完成 + missing cursor + 最新 turn 未完成，第一次调用应输出 0 条；补齐当前 turn 后只能输出 1 个 turn。

PR #375, opened 2026-09-04
fix(qoder-cn): collect qoderclicn LLM duration and token usage

rangemer333-cell:
## Problem

qoderclicn (the QoderCN CLI) produced LLM spans with zero duration and no token usage. Three separate causes:

1. The hook records `llm.request` and `llm.response` in the same tick, so the Hook interval alone yields a zero-width span. The real timings only exist in the session segments.
2. On this build both the segment records and the transcript report zero token counts and carry only `credits`, so usage has to come from the runtime intercept instead.
3. The segment reader was called without the batch's request ids, so it could not distinguish "this request does not exist" from "this request has not been flushed yet".

## Changes

**Runtime and hook** (`17c9e94b`, authored by @流屿) — support a managed QoderCN runtime so `qoderclicn` is wrapped and its intercept file is written.

**CLI collection** (`56cb3d66`) — route qoderclicn sessions to the CLI branch, join intercept usage by `gen_ai.response.id` and segment timings by `agent.client_request_id`, and parameterise the segments root so the CN reader never touches `~/.qoder`.

**Enrichment alignment and hook reader** (`9c8aa0b0`) — two fixes:

- Segment enrichment now mirrors the international input's two-stage model: `enrichCliPrimary` applies hook and intercept data, then `enrichCliFromSegments` fills only what is still missing. The reader receives this batch's exact client request ids via `collectCliExpectedRequestIds`, so an unflushed request gets the bounded retry added in #367. This matters because the hook offset is committed before enrichment: a record missed on the first read can never be repaired on a later poll.

- The hook JSONL reader was losing records. It discarded `bytesRead`, committed the offset at `stat.size`, and never truncated to the last newline, so a record still being appended when `stat()` returned would fail `JSON.parse` while the offset had already moved past its tail. It now follows `BaseHookInput`: read the actual byte count, parse and commit only through the last newline, hold the offset when no complete record is present, and report backlog/read/committed-batch metrics to the runtime accumulator.

## Note on a deliberate divergence from #367

The international input defers the segment read behind `needsCliSegmentFallback` to save I/O. The CN input cannot: QoderCN IDE and qoderclicn share one agent id and one hook history, so the presence of segments is itself the signal that routes a session to the CLI branch. The read therefore happens whenever `agent.client_request_id` is present. `enrichCliFromSegments` is gap-only, so running it unconditionally once segments are in hand is equivalent to running it only on fallback. The `expectedRequestIds` argument is still passed, since that is what the retry actually depends on.

## Testing

`tsc --noEmit` clean. 48 files / 672 tests pass across `tests/unit/inputs/`, `tests/unit/core/orchestrator.test.ts` and `tests/integration/qoder-cli-segment-span-timing.test.ts`.

The CLI branch of the CN input previously had no test coverage at all. Five tests were added:

- repairs a qoderclicn turn from the CN segments root and the CN intercept (end-to-end: real segment and intercept files, asserts usage, span duration and model)
- never reads the international segments root for a QoderCN session
- hands this batch's client request ids to the segment reader
- never commits a half-written record and reads it on the next poll
- holds the offset when the batch contains no complete record at all

Each was verified against a mutated implementation to confirm it fails when the fix is reverted (swapping the segments root constant, committing `stat.size`, and committing on the `lastNewline < 0` path).

## Known gaps, intentionally not in this PR

Both are now tracked separately:

- **#376** — the CN input has no multimodal collection. Only the qoderclicn half is wiring: `qoder-cli-multimodal` derives paths from the entry itself and is root-agnostic. The IDE half is not, because `qoder-ide-multimodal` reads attachments through `readAttachedImagePathsForRequestIds`, which resolves the *international* DB set, and the CN reader has no attachment reader at all.
- **#377** — `resolveQoderCnDbPaths` has no profile-level enumeration, unlike `listHashedProfileDbPaths` on the international side. This affects CN **IDE** token enrichment on remote/multi-profile setups, not qoderclicn. It has not been reproduced on a CN build, so the issue records that limit explicitly.

### DIFF
--- a/assets/hooks/qodercli-runtime-wrapper.sh
+++ b/assets/hooks/qodercli-runtime-wrapper.sh
@@ -2,10 +2,10 @@
 #
 # Launch qodercli with the token intercept scoped to this process only.
 #
-# npm qodercli distributions are Node.js entrypoints and require NODE_OPTIONS
-# --import. Native qodercli distributions are Bun executables and require
-# BUN_OPTIONS --preload. Set LOONGSUITE_QODERCLI_RUNTIME=node|bun to override
-# auto-detection, or LOONGSUITE_QODERCLI_BIN to provide the real executable.
+# npm and bundled SDK qodercli distributions are Node.js entrypoints and require
+# NODE_OPTIONS --import. Native qodercli distributions are Bun executables and
+# require BUN_OPTIONS --preload. Set LOONGSUITE_QODERCLI_RUNTIME=node|bun to
+# override auto-detection, or LOONGSUITE_QODERCLI_BIN to provide the real entry.
 
 SCRIPT_DIR=$(CDPATH= cd "$(dirname "$0")" 2>/dev/null && pwd)
 INTERCEPT_SCRIPT="$SCRIPT_DIR/qodercli-token-intercept.mjs"
@@ -20,18 +20,38 @@ if [ -z "$QODERCLI_BIN" ]; then
   exit 127
 fi
 
-# Missing hook assets must never prevent qodercli from starting.
-if [ ! -f "$INTERCEPT_SCRIPT" ]; then
-  exec "$QODERCLI_BIN" "$@"
-fi
-
 QODERCLI_RUNTIME="${LOONGSUITE_QODERCLI_RUNTIME:-}"
 if [ -z "$QODERCLI_RUNTIME" ]; then
-  if LC_ALL=C head -c 128 "$QODERCLI_BIN" 2>/dev/null | grep -aq '^#!.*node'; then
-    QODERCLI_RUNTIME=node
-  else
-    QODERCLI_RUNTIME=bun
+  case "$QODERCLI_BIN" in
+    *.js|*.cjs|*.mjs) QODERCLI_RUNTIME=node ;;
+    *)
+      if LC_ALL=C head -c 128 "$QODERCLI_BIN" 2>/dev/null | grep -aq '^#!.*node'; then
+        QODERCLI_RUNTIME=node
+      else
+        QODERCLI_RUNTIME=bun
+      fi
+      ;;
+  esac
+fi
+
+launch_qodercli() {
+  if [ "$QODERCLI_RUNTIME" = "node" ]; then
+    QODERCLI_NODE_BIN="${LOONGSUITE_QODERCLI_NODE_BIN:-}"
+    if [ -z "$QODERCLI_NODE_BIN" ]; then
+      QODERCLI_NODE_BIN=$(command -v node 2>/dev/null || true)
+    fi
+    if [ -z "$QODERCLI_NODE_BIN" ]; then
+      echo "loongsuite-pilot: node executable not found for qodercli entry" >&2
+      exit 127
+    fi
+    exec "$QODERCLI_NODE_BIN" "$QODERCLI_BIN" "$@"
   fi
+  exec "$QODERCLI_BIN" "$@"
+}
+
+# Missing hook assets must never prevent qodercli from starting.
+if [ ! -f "$INTERCEPT_SCRIPT" ]; then
+  launch_qodercli "$@"
 fi
 
 if [ "$QODERCLI_RUNTIME" = "node" ]; then
@@ -50,4 +70,4 @@ else
   export BUN_OPTIONS
 fi
 
-exec "$QODERCLI_BIN" "$@"
+launch_qodercli "$@"
--- a/assets/hooks/qodercli-runtime-wrapper.sh
+++ b/assets/hooks/qodercli-runtime-wrapper.sh
@@ -1,26 +1,36 @@
 #!/bin/sh
 #
-# Launch qodercli with the token intercept scoped to this process only.
+# Launch a Qoder CLI with the token intercept scoped to this process only.
 #
-# npm and bundled SDK qodercli distributions are Node.js entrypoints and require
-# NODE_OPTIONS --import. Native qodercli distributions are Bun executables and
-# require BUN_OPTIONS --preload. Set LOONGSUITE_QODERCLI_RUNTIME=node|bun to
-# override auto-detection, or LOONGSUITE_QODERCLI_BIN to provide the real entry.
+# Serves both product lines: LOONGSUITE_QODERCLI_FLAVOR selects which one, and
+# defaults to qodercli so an rc block written by an older install keeps working.
+# The rc block installed for the CN build sets it to qoderclicn.
+#
+# npm and bundled SDK distributions are Node.js entrypoints and require
+# NODE_OPTIONS --import. Native distributions are Bun executables and require
+# BUN_OPTIONS --preload. Set <FLAVOR>_RUNTIME=node|bun to override
+# auto-detection, or <FLAVOR>_BIN to provide the real entry, where <FLAVOR> is
+# LOONGSUITE_QODERCLI or LOONGSUITE_QODERCLICN. The names are per flavor on
+# purpose: a value exported for one product line must not redirect the other.
 
 SCRIPT_DIR=$(CDPATH= cd "$(dirname "$0")" 2>/dev/null && pwd)
 INTERCEPT_SCRIPT="$SCRIPT_DIR/qodercli-token-intercept.mjs"
-QODERCLI_BIN="${LOONGSUITE_QODERCLI_BIN:-}"
+QODERCLI_FLAVOR="${LOONGSUITE_QODERCLI_FLAVOR:-qodercli}"
+FLAVOR_VAR=$(printf '%s' "$QODERCLI_FLAVOR" | tr 'a-z-' 'A-Z_')
+INTERCEPT_FILE_NAME="${LOONGSUITE_INTERCEPT_FILE:-${QODERCLI_FLAVOR}-intercept.jsonl}"
+
+eval "QODERCLI_BIN=\${LOONGSUITE_${FLAVOR_VAR}_BIN:-}"
 
 if [ -z "$QODERCLI_BIN" ]; then
-  QODERCLI_BIN=$(command -v qodercli 2>/dev/null || true)
+  QODERCLI_BIN=$(command -v "$QODERCLI_FLAVOR" 2>/dev/null || true)
 fi
 
 if [ -z "$QODERCLI_BIN" ]; then
-  echo "loongsuite-pilot: qodercli executable not found" >&2
+  echo "loongsuite-pilot: $QODERCLI_FLAVOR executable not found" >&2
   exit 127
 fi
 
-QODERCLI_RUNTIME="${LOONGSUITE_QODERCLI_RUNTIME:-}"
+eval "QODERCLI_RUNTIME=\${LOONGSUITE_${FLAVOR_VAR}_RUNTIME:-}"
 if [ -z "$QODERCLI_RUNTIME" ]; then
   case "$QODERCLI_BIN" in
     *.js|*.cjs|*.mjs) QODERCLI_RUNTIME=node ;;
@@ -36,12 +46,12 @@ fi
 
 launch_qodercli() {
   if [ "$QODERCLI_RUNTIME" = "node" ]; then
-    QODERCLI_NODE_BIN="${LOONGSUITE_QODERCLI_NODE_BIN:-}"
+    eval "QODERCLI_NODE_BIN=\${LOONGSUITE_${FLAVOR_VAR}_NODE_BIN:-}"
     if [ -z "$QODERCLI_NODE_BIN" ]; then
       QODERCLI_NODE_BIN=$(command -v node 2>/dev/null || true)
     fi
     if [ -z "$QODERCLI_NODE_BIN" ]; then
-      echo "loongsuite-pilot: node executable not found for qodercli entry" >&2
+      echo "loongsuite-pilot: node executable not found for $QODERCLI_FLAVOR entry" >&2
       exit 127
     fi
     exec "$QODERCLI_NODE_BIN" "$QODERCLI_BIN" "$@"
@@ -49,10 +59,13 @@ launch_qodercli() {
   exec "$QODERCLI_BIN" "$@"
 }
 
-# Missing hook assets must never prevent qodercli from starting.
+# Missing hook assets must never prevent the CLI from starting.
 if [ ! -f "$INTERCEPT_SCRIPT" ]; then
   launch_qodercli "$@"
 fi
+
+# Read by the preload script to pick its output file.
+export LOONGSUITE_INTERCEPT_FILE="$INTERCEPT_FILE_NAME"
 
 if [ "$QODERCLI_RUNTIME" = "node" ]; then
   PILOT_PRELOAD_OPTION="--import=$INTERCEPT_SCRIPT"
--- a/assets/hooks/qodercli-runtime-wrapper.sh
+++ b/assets/hooks/qodercli-runtime-wrapper.sh
@@ -30,12 +30,31 @@ if [ -z "$QODERCLI_BIN" ]; then
   exit 127
 fi
 
+# True for a readable file that opens with no shebang and holds no NUL byte.
+# Such a file is a Node entrypoint: a Bun executable is compiled and carries NUL
+# bytes in its header, and anything the kernel can launch on its own names an
+# interpreter on the first line. Reading the file also follows the symlink a PATH
+# entry usually is, so no path resolution is needed to reach the real bytes.
+# A file we cannot read yields no evidence and is left to the caller's fallback.
+is_shebangless_text() {
+  LC_ALL=C head -c 2 "$1" 2>/dev/null | grep -aq '^#!' && return 1
+  _read=$(LC_ALL=C head -c 128 "$1" 2>/dev/null | wc -c | tr -d ' ')
+  [ "$_read" -gt 0 ] || return 1
+  _text=$(LC_ALL=C head -c 128 "$1" 2>/dev/null | LC_ALL=C tr -d '\000' | wc -c | tr -d ' ')
+  [ "$_read" = "$_text" ]
+}
+
 eval "QODERCLI_RUNTIME=\${LOONGSUITE_${FLAVOR_VAR}_RUNTIME:-}"
 if [ -z "$QODERCLI_RUNTIME" ]; then
   case "$QODERCLI_BIN" in
     *.js|*.cjs|*.mjs) QODERCLI_RUNTIME=node ;;
     *)
       if LC_ALL=C head -c 128 "$QODERCLI_BIN" 2>/dev/null | grep -aq '^#!.*node'; then
+        QODERCLI_RUNTIME=node
+      elif is_shebangless_text "$QODERCLI_BIN"; then
+        # A PATH entry's name carries no extension, so the arm above is all that
+        # stood between a bundled entry and the Bun branch, which exec's the file
+        # directly: the kernel refuses JS text and the CLI never starts.
         QODERCLI_RUNTIME=node
       else
         QODERCLI_RUNTIME=bun
--- a/src/core/hook-watchdog.ts
+++ b/src/core/hook-watchdog.ts
@@ -79,40 +79,75 @@ export interface WinRuntimeInterceptDefinition {
 }
 
 /**
+ * Prefix shared by every block this file writes into an rc file. It marks where
+ * a block whose END marker went missing has to stop: at the next block's BEGIN.
+ */
+const MARKER_BEGIN_PREFIX = 'loongsuite-pilot BEGIN ';
+
+/** True when `line` opens some block other than the one `begin` delimits. */
+function startsAnotherBlock(line: string, begin: string): boolean {
+  return line.includes(MARKER_BEGIN_PREFIX) && !line.includes(begin);
+}
+
+/**
  * Remove a marker-delimited block (inclusive of the BEGIN/END marker lines)
  * from rc-file content. Any line containing `begin` starts the cut and any
  * line containing `end` ends it; non-block lines are preserved verbatim.
+ *
+ * A block whose END marker is missing stops at the next block's BEGIN line
+ * instead of running to EOF. Running to EOF would drop every block written
+ * after it along with whatever the user keeps below them -- PATH edits, nvm
+ * init, aliases -- because repair() writes the stripped result back over the
+ * file. A repeated BEGIN for this same block keeps the cut open, so a file that
+ * accumulated duplicates collapses to a single removal.
  */
 export function stripMarkerBlock(content: string, begin: string, end: string): string {
   const out: string[] = [];
   let inBlock = false;
   for (const line of content.split('\n')) {
-    if (!inBlock && line.includes(begin)) { inBlock = true; continue; }
-    if (inBlock && line.includes(end)) { inBlock = false; continue; }
-    if (!inBlock) out.push(line);
+    if (!inBlock) {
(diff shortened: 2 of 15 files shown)


fangxiu-wf:
**[P1] Add the qoderclicn intercept lifecycle to the Windows installer.** The new token collection is installed only through the POSIX shell wrapper. `installer-opensource.ps1` has no equivalent injection, watchdog repair, disable, or uninstall path, so supported Windows installations never produce `qoderclicn-intercept.jsonl` and token usage remains missing.

**[P2] Do not gate intercept enrichment on segment availability.** If the first segment snapshot is unavailable after the bounded retry, the code falls back to the IDE path and commits the Hook offset even when an exact intercept response is already available. The turn is never repaired after segments appear. Apply exact-ID intercept enrichment independently and use segments only for timing/model enrichment.

rangemer333-cell:
Thanks — went through both. P2 is a real defect and is fixed in this PR; on P1 I agree with the facts but would like to propose a different scope.

## P2 — accepted, fixed in b77c932e

You are right, and the root cause is exactly the conflation you describe: the discriminator was `segments.length > 0`, which answered two independent questions at once — *is this a qoderclicn session* and *have its segment files landed yet*. Segments are written by the CLI itself and can lag behind the hook record, so a CLI turn whose segments were not visible yet took the IDE path and its intercept was **never read at all**. Since the Hook offset is committed before enrichment, those tokens were lost permanently rather than repaired on a later poll.

The fix discriminates on `agent.client_request_id` alone. The hook derives it from the transcript's `usage.request_id` and the IDE never produces it, so it is a stronger signal than segment presence and carries no timing dependency:

```ts
const isCliSession = hasClientRequestId(sessionEntries);
const segments = isCliSession ? await this.readCliSegments(...) : [];

if (isCliSession) {
  interceptData ??= await readInterceptData(undefined, CN_INTERCEPT_FILE);
  enrichCliPrimary(...);              // exact gen_ai.response.id join, unconditional
  if (segments.length > 0) {
    enrichCliFromSegments(...);       // response timings only
  }
} else {
  // unchanged IDE path
}
```

Two things worth flagging rather than burying:

**The comment above that loop used to state that both signals were required.** That claim was the origin of the defect, so I replaced it with the reason segments must *not* participate in the decision, including the cost of dropping them as a second check: if an IDE build ever starts carrying a request id it would take the CLI path and find no exact intercept match. That yields no tokens — which is precisely what gating on segments already produced for real CLI turns, so it is not a regression in the trade.

**All 13 pre-existing tests passed against this change**, i.e. they had no coverage of it at all. Added one that pins it: hook record plus intercept with **no segments written**, asserting the usage lands and the duration stays zero. Reverting the discriminator to `segments.length > 0` fails that test and only that test.

Residual, by design: a turn without segments keeps a zero LLM duration. The response timings exist nowhere else, so that part is genuinely unrecoverable — but the tokens no longer go down with it.

## P1 — facts confirmed, but I do not think this is a regression in this PR

Confirmed: `installer-opensource.ps1` has 0 occurrences of anything qoderclicn-related, against 33 in `installer-opensource.sh`. Windows installs do not produce `qoderclicn-intercept.jsonl`.

One data point that changes the scope, though: **`installer-opensource.ps1` has 0 occurrences of `qodercli` either.** The only line mentioning interception is a warning inside `Inject-QoderworkRuntimeWrapper`, which serves QoderWork / QwenWorkCN via `qoderwork-runtime-wrapper.mjs` and `QODER_WORKER_RUNTIME_PATH` — unrelated to the qodercli family. So token interception for qodercli has never existed on Windows; this PR did not introduce the asymmetry, it inherited it.

The mechanics also do not translate directly. The POSIX side injects a shell function into the rc files:

```sh
eval 'qoderclicn() { LOONGSUITE_QODERCLI_FLAVOR=qoderclicn "$DATA_DIR/hooks/qodercli-runtime-wrapper.sh" "\$@"; }'
```

Windows has no equivalent — a PowerShell profile would not cover a CLI launched from cmd — and the wrapper itself is a `.sh`, so a new `.ps1`/`.cmd`/`.mjs` runtime would be needed. Whether the existing ps1 mechanism (env-var runtime override) can be reused depends on whether qodercli exposes an injectable runtime entry point comparable to `QODER_WORKER_RUNTIME_PATH`. I have not verified that, and I would rather not guess at it inside this PR.

So I would suggest tracking it as its own issue scoped to **both qodercli and qoderclicn**, starting with that entry-point question, rather than as a fix here — otherwise the CN half would ship a Windows path the international half still lacks. Happy to open it, or to fold it in if you would rather keep it together; let me know which you prefer.

rangemer333-cell:
Opened #378 for the P1 Windows gap, scoped to both `qodercli` and `qoderclicn`. Happy to pull it back into this PR instead if you would rather not split it.

Two corrections to my own comment above that came out of writing it up:

**"no watchdog repair path" is not quite the right description.** The rc targets in `hook-watchdog.ts` (`qodercli-rc` L622, `qoderclicn-rc` L642) are registered with no platform guard, but they never activate on Windows because `precondition` is `fileExists(scriptPath)` and the `.sh` wrapper is never deployed there. `repair` also opens with `if (!await fileExists(rcPath)) continue; // never create rc files`, so nothing writes stray `.zshrc` files into `%USERPROFILE%`. The watchdog degrades cleanly; what it lacks is a Windows-shaped target, which only becomes meaningful once a Windows injection mechanism exists.

**The POSIX disable path is not a separate function.** `inject_qoderclicn_token_intercept` delegates to `remove_…` at L1303 when the agent is not selected, so injection and disable are one entry point. Listed accordingly in the issue.

Nothing about this was reproduced on a real Windows install — the issue states that explicitly, and flags the injectable-entry-point question as the thing to settle before any code gets written.

assets/hooks/qodercli-runtime-wrapper.sh
```diff
@@ -1,39 +1,72 @@
...
+  echo "loongsuite-pilot: $QODERCLI_FLAVOR executable not found" >&2
   exit 127
 fi
 
-# Missing hook assets must never prevent qodercli from starting.
-if [ ! -f "$INTERCEPT_SCRIPT" ]; then
-  exec "$QODERCLI_BIN" "$@"
+eval "QODERCLI_RUNTIME=\${LOONGSUITE_${FLAVOR_VAR}_RUNTIME:-}"
+if [ -z "$QODERCLI_RUNTIME" ]; then
+  case "$QODERCLI_BIN" in
```
linrunqi08, line 49: **[Medium]** Shebang-less `.mjs` detection matches `command -v`'s PATH name (`*.js|*.cjs|*.mjs`), not the symlink target. The rc-installed `qoderclicn()` uses PATH lookup; the only managed-SDK test sets `LOONGSUITE_QODERCLI_BIN` to a path that already ends in `.mjs`.

**影响:** A typical npm bin is `…/bin/qoderclicn → …/qoder-worker-runtime.obf.mjs`. The PATH name misses the extension check, `head` sees no `#!.*node`, and the wrapper takes the bun/`exec $BIN` path. A shebang-less Node entry ENOEXECs (CLI broken after install) or starts without `NODE_OPTIONS --import`, so `qoderclicn-intercept.jsonl` is never written and tokens stay zero — the managed-runtime path this PR is adding.

**建议:** Resolve one symlink/`realpath` before the extension check; if the target is `*.mjs|*.js|*.cjs`, force `QODERCLI_RUNTIME=node`. Add a test: PATH shim named `qoderclicn` → shebang-less `.mjs`, `LOONGSUITE_QODERCLI_FLAVOR=qoderclicn`, no `*_BIN` override.

src/core/hook-watchdog.ts
```diff
@@ -94,6 +94,27 @@ export function stripMarkerBlock(content: string, begin: string, end: string): s
...
+ * substrings (the qodercli and qoderclicn blocks both name the same wrapper
+ * script), and a file-wide `includes` would let one satisfy the other's check.
+ */
+export function extractMarkerBlock(content: string, begin: string, end: string): string | null {
+  const lines = content.split('\n');
+  const start = lines.findIndex(line => line.includes(begin));
+  if (start === -1) return null;
+  const rest = lines.slice(start + 1);
+  const relEnd = rest.findIndex(line => line.includes(end));
+  return relEnd === -1
```
linrunqi08: **[Medium]** `extractMarkerBlock` treats a missing END as “the rest of the file”. `qodercli-rc`'s signature is the shared path `qodercli-runtime-wrapper.sh`, which also appears inside a current `qoderclicn` block. An unterminated qodercli region is then vouched healthy and never migrated. The new neighbour test always includes END markers.

**影响:** A truncated `.zshrc`/`.bashrc` leaves a stale or half-written `qodercli()` in place while the CN block looks fine. If repair does run, unterminated `stripMarkerBlock` drops every line after BEGIN, including the sibling CN / claude-code blocks.

**建议:** Treat a missing END as unhealthy. Stop extract/strip at the next intercept BEGIN, not at EOF. Add a test: stale unterminated qodercli + current qoderclicn → `qodercli-rc.check() === false` and the CN block survives repair.

tests/unit/hooks/qodercli-runtime-wrapper.test.mjs
```diff
@@ -84,6 +85,31 @@ printf '{"nodeOptions":"%s","bunOptions":"%s","arg":"%s"}\\n' "$NODE_OPTIONS" "$
     expect(output.arg).toBe('hello');
   });
 
+  it('runs a bundled SDK .mjs entry through Node without requiring a shebang', async () => {
```
linrunqi08, line 107: **[Medium]** The dual-product-line wrapper is untested on the CN path. Every spawn uses the default flavor `qodercli`. This bundled-SDK case still sets `LOONGSUITE_QODERCLI_BIN` and never sets `LOONGSUITE_QODERCLI_FLAVOR=qoderclicn` or asserts `LOONGSUITE_INTERCEPT_FILE=qoderclicn-intercept.jsonl`.

**影响:** Input tests write `qoderclicn-intercept.jsonl` directly; watchdog tests only check that the rc block exports the flavor. The hop that actually names the intercept file is unprotected. A regression that ignores flavor or always writes `qodercli-intercept.jsonl` stays green while `QoderCnTraceInput` reads an empty CN file — the token-loss mode this PR is fixing. The PATH-shim / shebang-less `.mjs` hole is missed for the same reason.

**建议:** One wrapper spawn with `LOONGSUITE_QODERCLI_FLAVOR=qoderclicn`, a `qoderclicn` PATH entry (or `LOONGSUITE_QODERCLICN_BIN`), and a fixture that prints `LOONGSUITE_INTERCEPT_FILE`. Assert it equals `qoderclicn-intercept.jsonl`. Prefer sharing the PATH-symlink fixture with the runtime-detect fix.

linrunqi08:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 3 |
| Low      | 2 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | PASS |
| 死锁/卡死风险 | PASS |
| 状态恢复正确性 | PASS |
| PowerShell CLM 合规 | N/A |

无 FAIL 项。JSONL handle 在 `lastNewline < 0` 路径仍经 `finally` 关闭；collect 由 `cyclePromise` 串行化；offset 改为提交到最后一个 newline。变更不含 `.ps1`。

### Merge Gate（合入门禁）

**BLOCK** ⛔（阻断级别：`Critical` / `High` / `Medium`，`Low` 不阻断；Lifecycle 无 FAIL）

| Severity | 文件:行号 | 现状 |
| Medium | `assets/hooks/qodercli-runtime-wrapper.sh:35` | PATH 名不做 symlink 解析，managed `.mjs` 走 bun/ENOEXEC 或无 intercept |
| Medium | `src/core/hook-watchdog.ts:113` | 未闭合 qodercli 块可被 CN 邻块的 wrapper 路径背书 |
| Medium | `tests/unit/hooks/qodercli-runtime-wrapper.test.mjs:88` | 无 `FLAVOR=qoderclicn` / intercept 文件名断言 |

阻断合入，待修复后重评。

### 总体结论

P2（segment 门控 intercept）已在 `b77c932` 修好，CN CLI 的 join、半行 JSONL、segments root 参数化都扎实。仍有三处 Medium：rc 安装的 `qoderclicn()` 对 PATH 上指向无 shebang `.mjs` 的 shim 会判成 bun；`extractMarkerBlock` 在缺 END 时被邻块 signature 背书，正好打穿本 PR 要修的互证漏洞；wrapper 测试从未覆盖 CN flavor → `qoderclicn-intercept.jsonl` 这一跳。Windows 拦截缺口与国际线一致，跟 #378，不阻断本 PR。

### Highlights（正向实践）

- CLI/IDE 只看 `agent.client_request_id`，无 segments 仍吃 intercept，有测试钉死
- JSONL 按最后一个 newline 提交 offset，半行正反用例齐全
- `segmentStateKey(root, sessionId)` 避免 `~/.qoder` 污染 CN
- 完整双块的 marker 内 signature 匹配 + 邻块互证测试（完整 END 的情况）
- `qoder-cn-trace` pollInterval 从配置传入

对话评论 P2 已核对为 fixed；P1 跟踪 #378，不作为本 PR 回归。

---
*评审报告详见: `code-review/pr-375/final-report.md`*

assets/hooks/qodercli-runtime-wrapper.sh
rangemer333-cell: Fixed in `3b2bdb67`, but classified from the file header instead of resolving the path first.

**Why not `realpath`.** Reading the file already follows the symlink — `head` opens the target, so the real bytes are reachable without resolving anything. Resolving would also not be sufficient on its own: on this machine the real file is `~/.qoder-cn/bin/qoderclicn/qoderclicn-1.1.42`, which has no extension either, so the extension arm still would not fire. And a portable resolver needs a BSD fallback (`readlink -f` is GNU-only, `realpath` is absent before macOS 12.3) that CI would never execute, since both macOS and mainstream Linux have `realpath`.

**What it does instead.** A readable file that opens with no shebang and holds no NUL byte in its first 128 bytes is a Node entrypoint — a Bun executable is compiled and carries NUL bytes in its header, and anything the kernel can launch on its own names an interpreter on line 1. Measured NUL counts in the first 128 bytes: `qoderclicn-1.1.42` 94, `/bin/ls` 108, a shebang-less JS entry 0. Not a marginal discriminator.

**One correction to the impact note.** It is not "ENOEXECs *or* starts without `NODE_OPTIONS --import`" — a shebang-less text file always ENOEXECs, and the shell then interprets the JS as a shell script:

```
$ sh -c 'exec ./qoderclicn --version'    # symlink -> shebang-less .mjs
./qoderclicn: line 1: import: command not found
./qoderclicn: line 2: syntax error near unexpected token `'REAL_CLI_RAN''
```

So the only outcome is the CLI not starting; there is no variant that launches but misses the import. That also makes this the one branch in the wrapper able to stop the CLI from launching, which is the thing the script states a few lines down it must not do — *"Missing hook assets must never prevent the CLI from starting."* Worth weighing when ranking this against the other two.

**Test.** Added as suggested: PATH shim whose name carries no extension -> symlink -> shebang-less `.mjs`, no `*_BIN`. Confirmed red on the pre-fix wrapper and green after, and red again when the fix is reverted.

Kept the default flavor there rather than `qoderclicn`: flavor derivation is covered by the six tests from the third thread, so this one isolates the runtime classification.

**Two cases I guarded beyond the report.** Reading can also come up empty — a distribution shipping mode 111, or an empty file — and both start fine today on the bun path. A naive "text implies node" rewrite sends them to Node and breaks them, so the predicate requires that bytes were actually read. Mutation testing showed that guard had zero coverage at first; there is a test for it now.

src/core/hook-watchdog.ts
rangemer333-cell: Done as suggested in `7f47ff13`: a missing END now reads as unhealthy, extract and strip both stop at the next block's BEGIN, and there is a test with a stale unterminated `qodercli` block above a current `qoderclicn` one asserting `check() === false` plus the CN block and the user's own lines surviving repair.

Two notes on how it was done:

**Boundary keyed on the generic prefix.** `loongsuite-pilot BEGIN ` rather than the `-intercept` suffix, so a future marker of a different shape is bounded automatically. A repeated BEGIN for the *same* block keeps the cut open, so a file that accumulated duplicates collapses to one removal.

**Signature changed too.** `qodercli-rc.signature` is now `"eval 'qodercli() {"`, so no block can satisfy another's signature. Verified all three are mutually exclusive. Worth flagging that the existing test asserted this collision as intended behaviour — `expect(cn.blockFn(...)).toContain(cli.signature)` — with a comment naming per-block scoping as what keeps a stale block detectable next to a current one. The EOF fallback was precisely what defeated that scoping, so the two halves cancelled out. That test now forbids collisions instead.

**On a separate CN wrapper script.** Considered it, since it would make the script name unique again and remove the need to change the signature. Decided against: the CN block would then no longer contain the `qodercli` signature, so every already-installed `qodercli` block reads as unhealthy and triggers repair — the same repair whose strip cut to EOF. The migration would run on the path being fixed, for all installed users. Changing the constant has no such step.

tests/unit/hooks/qodercli-runtime-wrapper.test.mjs
rangemer333-cell: Done in `7f47ff13` — six tests: CN flavor end to end asserting `LOONGSUITE_INTERCEPT_FILE === 'qoderclicn-intercept.jsonl'`, a default-flavor contrast, cross-line isolation (`LOONGSUITE_QODERCLI_BIN` exported while the flavor is `qoderclicn` must not redirect it), the per-flavor `LOONGSUITE_QODERCLICN_BIN` override, an explicit `LOONGSUITE_INTERCEPT_FILE` override, and the not-found message naming the flavor.

**One honest limitation.** Mutation testing showed "the default flavor is `qodercli`" was already implicitly covered — flipping the default turns four pre-existing tests red, not just the new contrast test. So that one only makes an existing guarantee explicit. The promises that genuinely had no coverage were flavor-derived naming and cross-line isolation.

**Did not share the PATH-symlink fixture** with the runtime-detect fix as suggested: the two tests isolate different variables, and folding them together would mean a runtime-classification regression and a flavor regression surface as the same red.

**One trap worth recording for anyone extending these.** A fixture with a stripped `PATH` breaks the wrapper for an unrelated reason — it shells out to `tr` to derive `FLAVOR_VAR`, so an unreachable `tr` leaves that empty and the override eval degrades to `${LOONGSUITE__BIN:-}`, ending in exit 127 for the wrong cause. The tests use a `PATH` carrying the usual utilities but no CLI of either flavor.

rangemer333-cell:
All three addressed. `7f47ff13` covers the watchdog boundary and the CN wrapper coverage; `3b2bdb67` covers the runtime detection. Details and the two places I deviated from the suggestion are in the inline threads.

**How each change was verified.** Every new test was first confirmed red against the unpatched code and green after, then red again with the fix reverted — so none of them is a false positive that happens to pass. Each behavioural change was also mutation-tested: reverting it individually turns exactly the intended test red and nothing else. That surfaced one real gap — the guard distinguishing "bytes were read and looked like text" from "reading produced nothing" initially had zero coverage, and a test was added for it.

**One caveat if you run the suite yourself.** The full `vitest` run is flaky under concurrency here: two runs of the same tree gave 16 and 23 failures with disjoint failure sets, and suites that failed in aggregate pass 31/31 when run alone. So a red full run on this branch does not necessarily indicate a regression. I verified against the precise blast radius instead — the wrapper tests plus `hook-watchdog-intercept`, the only other consumer — at 62 passed across three consecutive runs. Happy to open a separate issue for the flakiness; it predates this branch.

**One thing I deliberately left undone.** When the entry has a non-Node shebang the wrapper still exports `BUN_OPTIONS --preload`, which a `sh`-interpreted entry ignores, so telemetry is silently lost with no startup failure to signal it. Fixing that means decoupling how the CLI is launched from how the intercept is injected, which is a behaviour change rather than a defect fix, so it does not belong in this PR. Happy to file it separately.

assets/hooks/qodercli-runtime-wrapper.sh
linrunqi08: **[fixed]** Header classification replaces PATH-extension matching.

**证据:** `3b2bdb67` 的 `is_shebangless_text`：可读、无 `#!`、前 128 字节无 NUL → Node。`head` 跟随 symlink，真实文件 `qoderclicn-1.1.42` 也没有扩展名，所以不解析路径是对的。测试 `routes a shebang-less entry reached through PATH to Node`（无 `*_BIN`）；compiled / unreadable 两条守卫也有覆盖。

src/core/hook-watchdog.ts
linrunqi08: **[fixed]** Unterminated blocks are unhealthy and bounded at the next BEGIN.

**证据:** `7f47ff13`：`extractMarkerBlock` 返回 `{text, terminated}`；`check()` 对 `!terminated` 直接 false；`stripMarkerBlock` 停在下一个 `loongsuite-pilot BEGIN`。测试 `treats a qodercli block that lost its END marker as unhealthy` 断言 repair 后 CN 块和用户 PATH 仍在。signature 改为 `eval 'qodercli() {`，三块互斥。

tests/unit/hooks/qodercli-runtime-wrapper.test.mjs
linrunqi08: **[fixed]** CN flavor now has end-to-end intercept-filename coverage.

**证据:** `7f47ff13` 测试 `scopes the intercept file to the qoderclicn flavor` 断言 `LOONGSUITE_INTERCEPT_FILE === 'qoderclicn-intercept.jsonl'`；另有 default 对照、`LOONGSUITE_QODERCLI_BIN` 不得重定向 CN、以及 `LOONGSUITE_QODERCLICN_BIN`。

linrunqi08:
## 🔍 Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 0 |
| Low      | 3 |

### Lifecycle Verdict

| Check | Result |
|-------|--------|
| 资源释放 | PASS |
| 死锁/卡死风险 | PASS |
| 状态恢复正确性 | PASS |
| PowerShell CLM 合规 | N/A |

### Merge Gate（合入门禁）

**APPROVE-READY** ✅

Round 1 三条 Medium（PATH shebang-less `.mjs`、未闭合 rc 块邻块背书、CN flavor 测试）分别在 `3b2bdb67` / `7f47ff13` 修好，已逐条回复并 resolve。本轮无新的阻断项。

### 总体结论

本轮增量（`7f47ff13` + `3b2bdb67`）针对 round 1 的阻断项对症下药：runtime 按文件头分类而不是 PATH 扩展名；未闭合 marker 停在下一个 BEGIN 且缺 END 视为不健康；wrapper 补了 `qoderclicn-intercept.jsonl` 与跨产品线 BIN 隔离测试。剩余 Low（安装器 signature 未对齐、文档、terminal-marker 测试粒度）不阻断合入。

### Highlights（正向实践）

- PATH shim → 无 shebang 文本走 Node，compiled/unreadable 不误判
- extract/strip 停在邻块 BEGIN，用户 PATH 行在 repair 测试中保留
- CN flavor 与国际线 override 互相不串

---
*评审报告详见: `code-review/pr-375/final-report.md`*

linrunqi08 (approved):
Round 2: F1/F2/F3 Medium findings are fixed on 3b2bdb67 / 7f47ff13. APPROVE-READY. Remaining items are Low and do not block. Generated by LoongSuite-Pilot Code Review Agent

PR #381, opened 2026-09-06
fix(codex): support single-quoted TOML section keys in trust writer (…

### LINKED ISSUE #380: [Bug] Codex trusted_hash block is re-added after manual removal
## Description

LoongSuite Pilot repeatedly re-adds the managed Codex hook trust block to %USERPROFILE%/.codex/config.toml after it is manually removed. This happens again on the next Pilot deployment or watchdog cycle, so users cannot remove the Codex integration by editing the file.

The block is marked with:

~~~toml
# BEGIN otel-codex-hook trust
[hooks.state."%USERPROFILE%/.codex/hooks.json:session_start:0:0"]
trusted_hash = "sha256:..."
# END otel-codex-hook trust
~~~

## Environment

- Windows 11
- LoongSuite Pilot 1.5.2, installed under %USERPROFILE%/.loongsuite-pilot/versions/1.5.2_cb94ec4/
- Codex hooks at %USERPROFILE%/.codex/hooks.json
- Pilot service started by a Windows scheduled task via bin/collector-launch.vbs

## Steps to reproduce

1. Install and run LoongSuite Pilot with the Codex agent detected.
2. Remove the otel-codex-hook trust block from %USERPROFILE%/.codex/config.toml.
3. Wait for the Pilot service to redeploy or for its hook watchdog to run.
4. Observe that the trust block is written back automatically.

## Expected behavior

Manual removal should persist, or there should be a documented disable or uninstall command that prevents Codex hook deployment and trust-block repair.

## Actual behavior

The trust block is recreated automatically. In addition, the current trust parser appears to recognize only table headers using double quotes. TOML also permits single-quoted literal strings, so existing entries such as [hooks.state.'...'] may be treated as missing and rewritten on every deployment.

## Additional context

The service log reports a trust self-check followed by deployment with changed=true after the block is removed. Please consider:

- making the trust-block parser accept both TOML basic and literal quoted keys;
- making deployment respect an explicit per-agent disable state;
- documenting how to disable or uninstall the Codex integration without the service restoring it.

xunavy:
…#380)
### Summary

Fixes #380.

In Windows environments, Codex (backed by Rust's `toml` serializer) persists path-based table keys as TOML literal strings using single quotes (e.g. `[hooks.state.'C:\Users\...:session_start:0:0']`) to avoid backslash escaping issues.

Previously, `codex-trust-writer.ts` only matched double-quoted keys (`[hooks.state."..."]`). As a result:
1. `parseTrustSections` failed to match any existing entries, leading `verifyTrustHashes` to falsely consider keys missing (`False Missing`).
2. The periodic `HookWatchdog` repeatedly appended duplicate double-quoted trust sections to `~/.codex/config.toml`.
3. Conflicting and corrupted configurations prevented Codex from launching properly, and manually deleting entries caused an endless loop of re-injection.

### Changes

- **`src/deployment/codex-trust-writer.ts`**:
  - Updated table header regex to capture both single-quoted (`'...'`) and double-quoted (`"..."`) keys.
  - Enhanced `tomlKeyCandidates` to unwrap single-quoted literal keys directly without JSON-string decoding.
  - Updated `trusted_hash` regex to accept both single and double quotes.
  - Updated `removeExactTrustSections` to cleanly identify and purge single-quoted keys upon undeploy/cleanup.
- **`tests/unit/deployment/codex-trust-writer.test.ts`**:
  - Added unit test verifying that single-quoted keys in `config.toml` pass hash verification without false missing-key repair.
  - Added unit test verifying that `removeTrustBlock` properly removes single-quoted trust blocks while preserving third-party sections.

### Verification

- Ran `vitest run tests/unit/deployment/codex-trust-writer.test.ts` (24/24 passed).
- Built project via `npm run build` (successful).
- Verified manually against local Windows `~/.codex/config.toml` files.

### DIFF
--- a/src/deployment/codex-trust-writer.ts
+++ b/src/deployment/codex-trust-writer.ts
@@ -177,6 +177,9 @@ function decodeTomlBasicString(value: string): string | null {
 }
 
 function tomlKeyCandidates(value: string): string[] {
+  if (value.startsWith("'") && value.endsWith("'")) {
+    return [value.slice(1, -1)];
+  }
   const decoded = decodeTomlBasicString(value);
   const raw = value.startsWith('"') && value.endsWith('"') ? value.slice(1, -1) : null;
   return [...new Set([decoded, raw].filter((candidate): candidate is string => candidate !== null))];
@@ -188,18 +191,20 @@ interface ParsedTrustSection {
   enabledLine?: string;
 }
 
+const TOML_HOOK_STATE_CAPTURE = /^\s*\[hooks\.state\.(("(?:\\.|[^"\\])*")|('[^']*'))\]\s*$/;
+
 function parseTrustSections(content: string): ParsedTrustSection[] {
   const lines = content.split('\n');
   const sections: ParsedTrustSection[] = [];
-  const header = /^\s*\[hooks\.state\.("(?:\\.|[^"\\])*")\]\s*$/;
   for (let i = 0; i < lines.length; i++) {
-    const match = lines[i]!.match(header);
+    const match = lines[i]!.match(TOML_HOOK_STATE_CAPTURE);
     if (!match) continue;
-    const key = tomlKeyCandidates(match[1]!)[0];
+    const rawKeyToken = match[1]!;
+    const key = tomlKeyCandidates(rawKeyToken)[0];
     if (key === undefined) continue;
     const section: ParsedTrustSection = { key };
     for (let j = i + 1; j < lines.length && !/^\s*\[/.test(lines[j]!); j++) {
-      const hash = lines[j]!.match(/^\s*trusted_hash\s*=\s*"([^"]+)"/);
+      const hash = lines[j]!.match(/^\s*trusted_hash\s*=\s*["']([^"']+)["']/);
       if (hash) section.hash = hash[1];
       if (/^\s*enabled\s*=/.test(lines[j]!)) section.enabledLine = lines[j]!.trim();
     }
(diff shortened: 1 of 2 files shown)


rangemer333-cell:
Thanks for the fix. I reviewed head bf967d2 and locally verified the related deployment suite: 4 test files / 125 tests passed, and both typecheck and build passed.

The single-quoted TOML parsing change itself is valid: an existing hooks.state literal-string key is now recognized, verification no longer reports a false missing key, and cleanup can remove the single-quoted section while preserving third-party sections.

However, this does not fully close #380 as currently described. The main reproduction is to remove the managed trust section and wait for deployment or the watchdog. verifyTrustHashes still treats an absent key as a mismatch, and HookStrategy.deploy still calls writeTrustedHashes, so the deleted section will still be recreated while Codex remains enabled. The new tests cover single-quoted parsing and cleanup, but not persistence of manual removal.

I suggest either:

1. Change Fixes #380 to clarify that this PR only addresses the single-quote false-missing and duplicate-section subproblem, leaving the explicit opt-out behavior tracked separately; or
2. Complete the opt-out lifecycle and add a regression covering enabled -> deployed -> disabled -> hooks and trust removed -> watchdog does not restore them.

There is also a current gating inconsistency relevant to the proposed workaround: the documentation presents agent-control.json mode off as disabling an Agent, but deployment and HookWatchdog use only config.json agents.<id>.enabled. At present, agents.codex.enabled=false plus a restart is the reliable route that reaches undeploy; agent-control.json off alone does not prevent deployment/watchdog reinjection. The two gates should be unified or the exact supported route should be documented.

Minor hardening: the new trusted_hash regex accepts mismatched quote pairs because the opening and closing quote are independent character classes. Two explicit alternatives or a backreference would avoid treating invalid TOML such as a single-quote opener with a double-quote closer as valid.

Monsooooon:
we have resolved this issue in https://github.com/alibaba/loongsuite-pilot/pull/401

PR #385, opened 2026-09-08
fix(qoder): derive IDE steps from transcript structure

rangemer333-cell:
## Summary

- derive Qoder IDE LLM boundaries from transcript structure instead of progress-hook timestamps
- retain exact transcript index ranges for each response so third-party hooks cannot merge or reassign response content
- treat a non-empty assistant `stop_reason` as provider-response completion, while delaying the split until the next assistant so intervening tool results remain in the preceding response range
- keep `tool_use.id` / `tool_result.tool_use_id` closure as the fallback for older transcripts without `stop_reason`
- preserve Qoder transcript fractional timestamps at nanosecond precision and use the final assistant timestamp when SQLite enrichment is unavailable
- place each enriched request after the previous response and every preceding tool result
- when SQLite millisecond timestamps collide or conflict with transcript ordering, minimally move the current response forward so `previousResponse/toolResult < request < response` remains true

## Compatibility behavior

Third-party progress records, including QoderSec `PostToolUse` hooks, no longer define LLM boundaries or content ownership. Tool call/result association remains ID-based.

If a transcript omits `tool_result` but the tool-call response has `stop_reason`, the following assistant response remains a separate LLM step and can carry the terminal `end_turn` event. No tool result is fabricated. The split occurs only when the following assistant arrives, so tool results written after `stop_reason` still belong to the preceding response range. If both `tool_result` and `stop_reason` are absent, the collector conservatively retains later assistant rows in the current group instead of inventing a boundary.

## Validation

- focused Qoder regression suite: 9 files, 199 tests passed
- `npm run typecheck`
- `npm run build`
- replayed all seven affected customer QoderSec turns (203 LLM steps): zero missing pairs, invalid intervals, predecessor-order violations, duplicate response timestamps, or unnecessary response adjustments
- replayed the older customer transcript after the terminal-response change: unchanged at 7 turns / 203 LLM steps because it has no `stop_reason` and continues to use tool-ID closure
- replayed two affected turns through the installed Hook -> history -> QoderTraceInput -> output path:
  - 16 expected LLM calls: old output 9 responses, new output 16 request/response pairs
  - 3 expected LLM calls: old output 2 responses, new output 3 request/response pairs
  - zero empty responses, invalid intervals, predecessor-order violations, transcript-fallback mismatches, or tool ID mismatches

PR #391, opened 2026-09-08
fix(qoder): preserve terminal response without tool result

rangemer333-cell:
## Summary

- follow up #385 by treating a non-empty Qoder assistant `stop_reason` as provider-response completion
- delay the actual group split until the next assistant row so intervening `tool_result` records remain in the preceding response range
- preserve a later terminal `end_turn` response when a cancelled/interrupted tool has no `tool_result`
- retain tool-ID closure as the fallback when older transcripts have no `stop_reason`

Third-party progress records, including QoderSec `PostToolUse`, remain excluded from LLM boundary decisions. The collector does not fabricate a missing tool result.

## Validation

- focused Qoder regression suite: 9 files, 199 tests passed
- `npm run typecheck`
- `npm run build`
- older customer transcript without `stop_reason`: unchanged at 7 turns / 203 LLM steps

### DIFF
--- a/assets/hooks/qoder-hook-processor.mjs
+++ b/assets/hooks/qoder-hook-processor.mjs
@@ -919,20 +919,24 @@ function hookEventOf(row) {
 
 export function buildLlmBoundaries(_progressEvents, contentEvents) {
   // Step 1: Group assistant blocks into LLM calls. Qoder transcript assistant
-  // rows do not carry message.id, and progress windows are hook timing signals,
-  // not reliable provider-call boundaries.
+  // rows do not consistently carry message.id, and progress windows are hook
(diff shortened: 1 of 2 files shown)


fangxiu-wf (approved):
APPROVE. The missing-tool-result case is now handled correctly: a non-empty assistant `stop_reason` completes the provider response, while delaying the split preserves an intervening `tool_result` in the preceding step. I also verified that the terminal `end_turn` response triggers immediate OTLP conversion with idle timeout disabled. Focused Qoder tests, timing/flusher tests, typecheck, and build all pass. The previous timestamp-order issue is already fixed in the merged #385 base and has not regressed.

PR #260, opened 2026-08-14
fix(genai): preserve assistant tool-call history across agents

fangxiu-wf:
## Summary

- preserve the previous assistant `tool_call` message before matching `tool_call_response` messages in subsequent LLM inputs
- cover Qoder/QoderWork, Cursor, Qwen Code/QwenWorkCN, and WorkBuddy source-specific collection paths
- keep correlation source-faithful by matching native call IDs and avoiding fabricated tool responses
- leave the shared converter, semantic schema, flushers, and unrelated agents unchanged

## Agent coverage

- Qoder family: Hook transcript paths plus QoderWork CN direct trace input
- Cursor: React assembler and transcript fallback assembler
- Qwen: Qwen Code CLI transcript conversion and QwenWorkCN Hook processing
- WorkBuddy: transcript tool-wave event construction

## Validation

- related-agent regression: 541 passed, 3 skipped across 44 test files
- schema and content-policy validation: 30 passed
- `npm run typecheck`: passed
- `npm run build`: passed
- `git diff --check`: passed
- full test run: 2991 passed, 48 skipped, 3 failed; the Qoder cold-start and FileWatcher timing failures passed on isolated rerun, while the remaining environment-dependent QwenWork real-transcript test selected a local incomplete transcript containing user records but no assistant response

Live real-Agent traffic and SLS/ARMS readback are not included in this PR validation yet.

Fixes #258

PR #394, opened 2026-09-08
fix(genai): restore Claude and Cursor skill history

rangemer333-cell:
## Summary

- restore the Claude Code synthetic `load_skill` call in the first LLM output and propagate the synthetic call/result into later LLM inputs
- restore Cursor assistant tool-call history and matching tool responses in subsequent LLM inputs
- keep Cursor subagent history isolated and require non-empty tool IDs when pairing responses
- preserve current `gen_ai.input.messages_delta` semantics while retaining the complete assistant output in cumulative input history

## Why this is being resubmitted

PR #260 was merged into `main` before its follow-up PRs #274 and #273 were merged into the already-merged feature branch. As a result, the follow-up commits never reached `main`. This PR ports those stranded changes onto the latest `main`.

Original PRs: #274, #273

## Validation

- `npx vitest run --no-file-parallelism tests/unit/hooks/claude-code/skill-read.test.mjs tests/unit/hooks/claude-code/hook-processor.test.mjs tests/unit/hooks/claude-code/transcript-parser.test.mjs tests/integration/claude-code-tool-context-flow.test.mjs tests/unit/hooks/cursor/inject-skill-records.test.mjs tests/unit/hooks/cursor/skill-detector.test.mjs tests/unit/hooks/cursor-react-assembler.test.ts tests/unit/hooks/cursor-transcript-assembler.test.mjs` (184 passed, 2 skipped)
- `npm run typecheck`
- `npm run build`
- `node --check` for the four changed Hook processor/assembler modules
- `git diff --check origin/main...HEAD`

### DIFF
--- a/assets/hooks/claude-code-hook-processor.mjs
+++ b/assets/hooks/claude-code-hook-processor.mjs
@@ -344,20 +344,92 @@ function skillAttributes(skillLoad, fallbackName) {
   };
 }
 
-function positiveSkillLoadTimes(skillLoad, fallbackTimestamp) {
+function skillLoadTimesAtLlmEnd(llmEndTimestamp) {
   const minDurationNanos = 1_000_000n; // converter uses millisecond start/end times
-  let call = BigInt(isoToUnixNanos(
-    skillLoad?.commandTimestamp || skillLoad?.metaTimestamp || fallbackTimestamp,
-  ));
-  let result = BigInt(isoToUnixNanos(
-    skillLoad?.metaTimestamp || fallbackTimestamp || skillLoad?.commandTimestamp,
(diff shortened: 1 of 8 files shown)


fangxiu-wf (approved):
APPROVE. I re-reviewed the port against the current main merge result. The previously approved Claude and Cursor skill-history behavior is preserved, including complete assistant/tool context, subagent isolation, non-empty-ID response pairing, and content-off handling. The focused suite passed with 184 tests and 2 skipped; an additional 31 Cursor/OTLP tests, typecheck, build, syntax checks, and a converter reproduction also passed. No blocking findings.

PR #416, opened 2026-09-17
feat(claude-code): forward trace context to the LLM gateway and nest its spans correctly

rangemer333-cell:
## Summary

Extends optional upstream trace linking one hop further for Claude Code: to the
Anthropic-compatible LLM gateway/proxy the user operates. Both parts are gated
behind a new `upstreamLink.propagateToLlm` switch (off by default, symmetric
with the existing `propagateToTools`), and the whole path is fail-open.

- **Forward W3C trace context to the gateway.** The Claude Code fetch preload
  injects a `traceparent` request header (and a validated `tracestate`) onto LLM
  requests, so gateway-side spans can join the same trace as the collected agent
  spans. A `traceparent` the caller already set is never overwritten.
- **Nest gateway spans under the request they belong to.** Forwarding the
  upstream `parent-id` verbatim made gateway spans siblings of the Claude Code
  ENTRY span rather than children of the LLM call. The preload now mints a span
  id per *physical* request, persists it on both the enrichment record and that
  request's attempt record, and the OTLP flusher reuses the reserved id when it
  builds the LLM span — so the id in the header is the id that actually ships.
  A retried call emits one span per attempt, and each attempt's gateway spans
  nest under the attempt that issued them.
- Generalises the existing TOOL span-id reservation into a shared
  `SpanIdReservations` helper (`tool-span-id-reservation.ts` →
  `span-id-reservation.ts`) now that LLM spans need the same mechanism.

Merged with `main` after #411 landed, so the minting granularity matches its
per-attempt LLM span structure.

## Configuration

| Setting | Env | Default |
| ------- | --- | ------- |
| `upstreamLink.propagateToLlm` | `LOONGSUITE_PILOT_UPSTREAM_LINK_PROPAGATE_TO_LLM` | disabled |

Requires `upstreamLink.enabled` as well: forwarding a context the collector is
not linking would point the gateway at a span that only exists inside pilot's
private trace. README / README.zh-CN and the zh-CN propagation doc are updated.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `tests/unit/hooks/claude-code/` — 6 files / 173 tests, incl.
      `fetch-intercept` (73) with the new switch block (forwards nothing while
      off, needs both halves, reads both halves from `config.json` when no env
      override is set) and `hook-processor` (51) with the new cases: the minted
      id is adopted as the LLM span id, an invalid id falls back to a locally
      generated one, and each retry attempt span claims the id its own physical
      request advertised
- [x] `tests/unit/core/config-loader.test.ts`, `tests/unit/upstream-link/`,
      `tests/unit/flushers/llm-span-id-reservation.test.ts` and the OTLP flusher
      suites — 27 files / 338 tests
- [x] Manual: Claude Code against a user-operated Anthropic-compatible gateway
      that returns transient HTTP errors, connection errors, and recovery or
      final failure after retries. Tests keep this in-process with a mocked
      `fetch`/`Response` driving 429 / 5xx / network-error / success sequences —
      no external service is contacted.

## Known limitations

- A request that fails before the SSE `message_start` event has no enrichment
  record; its id is recovered from the attempt record instead. If attempt
  grouping is ambiguous (repeated request ids with conflicting hashes) the
  attempts are left untouched, so those gateway spans still parent to a span
  pilot never emits.
- Unchanged: subagents, PowerShell, MCP and non-Bash tools are out of scope for
  the tool-side direction.


sipercai (approved):

fangxiu-wf (commented):
Reviewed under the explicitly limited contract that `propagateToLlm` only guarantees propagation of a valid process-inherited `TRACEPARENT` for the first Claude Code turn. ACP correlation, locally generated per-turn traces, and subsequent turns are out of scope.

Within that scope, the trace ID propagation, per-physical-request span ID reservation, retry handling, Hook enrichment, and final OTLP LLM span ID reuse are consistent. Targeted tests and CI pass. No blocking findings from me.

The previously identified multi-turn/ACP mismatch is considered out of scope, not fixed. If support expands to those scenarios, it should be revisited.

fangxiu-wf (approved):

PR #432, opened 2026-09-21
fix(grok-build): find hook gold copy on the flattened K8s layout

linrunqi08:
## Summary
- Grok Build's watchdog always restored hooks from `$PILOT_DIR/assets/hooks`. Host installs keep that gold copy; the K8s payload flattens `assets/hooks/` into top-level `hooks/` and historically dropped `assets/hooks`, so `access()` threw `ENOENT` on every repair cycle (`hook:grok-build` / `intercept-watchdog.repair-failed`). Collection itself kept working because the live hook is already at `$PILOT_DATA/hooks`.
- Prefer `assets/hooks` when it exists (host install), otherwise use the flattened `hooks/` tree so coinciding `/mnt/pilot` layouts are healthy instead of repair-failing forever.

## Other agents
Checked every `$PILOT_DIR/assets/...` and `$PILOT_DATA/{hooks,plugins}` consumer:

| Agent | Runtime path | K8s payload | Status |
| --- | --- | --- | --- |
| Grok Build watchdog | `$PILOT_DIR/assets/hooks` | flattened, no `assets/hooks` | **this bug** |
| Hermes Agent | `$PILOT_DIR/assets/plugins/hermes-agent/...` | `assets/plugins -> ../plugins` already exists | OK |
| OpenClaw / OpenCode / MiMo / Pi / DSH / AgentTeams | `$PILOT_DATA/plugins/...` | top-level `plugins/` | OK |
| Other hook agents (Cursor, Claude, Codex, Qoder, …) | `$PILOT_DATA/hooks/...` | top-level `hooks/`; no gold-copy integrity check | OK |

Local dashboard reads `$PILOT_DIR/assets/dashboard/index.html`, which the K8s payload also does not ship. That is non-fatal and not an agent collection path, so it is left out of this PR.

The proprietary init image should also grow `ln -s ../hooks "$OUT/assets/hooks"` next to the existing plugins link; that file is not in this repo.

## Test plan
- [x] `npx vitest run tests/unit/deployment/grok-build-assets.test.ts`
- [ ] Host install: `assets/hooks` still wins when both layouts exist
- [ ] K8s coinciding layout (`LOONGSUITE_PILOT_DATA_DIR=/mnt/pilot`, no `assets/hooks`): watchdog stays healthy, no `intercept-watchdog.repair-failed`
- [ ] Split layout (`~/.loongsuite-pilot` + payload `hooks/`): restore copies from flattened payload into the data dir


linrunqi08:
Closing this in favor of the payload-layout fix only.

K8s assemble-payload already flattens `assets/hooks/` into top-level `hooks/`. Adding `assets/hooks -> ../hooks` (same dual-layout as plugins) makes `$PILOT_DIR/assets/hooks` exist, so the original Grok watchdog gold-copy path works without a daemon fallback.

The daemon-side `resolveGrokBuildHookSourceRoot` change is not needed once that symlink ships in the init image.

